PR #1, opened 2018-09-11
Implementation/fetch data for dashboardnews

nestordedios:


nestordedios:
I will delete this PR since it's updated in #2

PR #11, opened 2018-10-08
Improvements to Bolt Parameters

xiaohutai:
- Update Symfony 4.2.x-dev, based on https://github.com/symfony/demo/pull/865
- Added to `composer.json`:
```diff
+"symfony/security-bundle": "4.2.x-dev",
+"symfony/serializer": "4.2.x-dev",
```
- If you _use_ the variables like `bolt.secret` it will be compiled! Inject `\Psr\Container\ContainerInterface $container` in a Controller and then dump it, like so:

image

- For now, I injected the values from `config.yaml` in `'bolt.config.general'`. I'm not sure how to officially use/inject these in other parts of the application. See https://symfony.com/doc/current/service_container/synthetic_services.html
  - It might mean that services use these cannot be auto-wired.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,19 +17,21 @@
         "miljar/php-exif": "^0.6.4",
         "nesbot/carbon": "^1.34",
         "sensio/framework-extra-bundle": "^5.1",
-        "sensiolabs/security-checker": "^4.1",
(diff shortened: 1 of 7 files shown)


bobdenotter:
Nice. 👍

Issue #8, opened 2018-10-02
Convert Twig-based "slug" field into Vue component. 


bobdenotter:
Currently, the "slug" field is generated in HTML. Code for the field is here: https://github.com/bobdenotter/symfony-skeleton/blob/master/templates/editcontent/fields/slug.twig

<img width="867" alt="screenshot 2018-10-02 at 19 38 08" src="https://user-images.githubusercontent.com/1833361/46367310-79b64480-c67d-11e8-863a-148b6b2ebf4a.png">

Ideally, this field should be converted to a nice, self-contained Vue component, that's easier to maintain. 

Current JS code entry point is here: https://github.com/bobdenotter/symfony-skeleton/blob/master/assets/js/bolt.js

Build instructions are in README.md. After setting up, you should be able to log on to the Admin Panel, at `/bolt`, and from the Dashboard click on "Edit" on any of the dummy records to go to the Edit page. 

Note: the current code is very prototype-y. Feel free to re-arrange stuff, if it makes more sense. :-)

jackiboy:
@bobdenotter you can close this badboy

bobdenotter:
Fixed in #26

PR #26, opened 2018-10-18
Added Slug Vue Component

jackiboy:
#8 Added Slug Vue Component.


.gitignore
```diff
@@ -39,3 +39,14 @@ yarn-error.log
 /.php_cs
 /.php_cs.cache
 ###< friendsofphp/php-cs-fixer ###
+
+### VisualStudioCode ###
```
bobdenotter: Would it suffice to just put `.vscode/*` in .gitignore? The `admin/` should def not be in here. :-)

templates/editcontent/fields/image.twig
```diff
@@ -43,12 +43,12 @@
 
         </div>
         <div class="col-4" id="{{ id }}-uploader" style="padding: 0; min-height: 180px;">
-            <file-pond name="filepond[]" ref="pond" accepted-file-types="image/jpeg, image/png" server="/async/upload"></file-pond>
+            {# <file-pond name="filepond[]" ref="pond" accepted-file-types="image/jpeg, image/png" server="/async/upload"></file-pond> #}
```
bobdenotter: We can make these work another way, right? I assume we also need to pull them in, instead of CDN'ing them.

templates/editcontent/fields/slug.twig
```diff
@@ -1,29 +1,12 @@
...
-                    <i class="fas fa-link fa-fw"></i> Generate from: {{ field.slugUseFields|join(', ') }}
-                </a>
-                <a class="dropdown-item" data-action="edit">
-                    <i class="fas fa-pencil-alt fa-fw"></i> Edit
-                </a>
-            </div>
-        </div>
-    </div>
-
+    <editor-slug
```
bobdenotter: So much nicer! ^_^

templates/editcontent/fields/image.twig
jackiboy: Yes I will make a component for it then import it inside that

templates/editcontent/fields/slug.twig
jackiboy: \o/

.gitignore
jackiboy: I have resolved this :-)

PR #33, opened 2018-10-23
Added Vendor and Manifest To Frontend

jackiboy:
Added The Vendor and Manifest to the frontend. Added Async/Await Babel polyfill to use Es7 Functionality.


bobdenotter:
👍

PR #115, opened 2018-11-16
Rename index.vue for Case Sensitive filesystems

bobdenotter:
This ought to fix your build, @xiaohutai


xiaohutai:
Nice, this works for me!

Issue #41, opened 2018-10-23
Make 'setcontent'


bobdenotter:

JarJak:
I hope it's implementation will be less magical than in v3 ;)

bobdenotter:
@JarJak See #116 

It's being worked on by @xiaohutai.. It's definitely based on our Bolt 3 code, but converted to use ORM. I assume by "magic" you mostly mean the hydration, right? That should be handled by ORM now, and it seems much snappier than before.

JarJak:
With "magical" I mean using user_func_*, side effects, implicit changing object state etc.

bobdenotter:
Closed by #116

PR #116, opened 2018-11-16
[WIP] setcontent

xiaohutai:
Some basic features `setcontent` calls working, based on `Bolt\Storage\Query\Query` from Bolt 3.

This is the first of the `setcontent` iteration (based on @rossriley's work for Bolt 3).
Some complex situations arise because a content-type is not contained in one table anymore.

So we have `bolt_content` with a has-many relationship to `bolt_field`.

A distinction is to be made when query-ing:
- Fields from the base table `bolt_content`. This is very simple and remains similar to Bolt 3's way for fetching content.
- Fields from the fields table `bolt_field`. A bit more complex, because we need to use `JOIN`s.

### Basic

- ✔️ `{contenttype}`
- ✔️ `{contenttype}/first/{number}`
- ✔️ `{contenttype}/last/{number}`
- ❌ `{contenttype}/random/{number}`
  - `random` is not going to work with Doctrine ORM queries, see [answer on Stack Oveflow](https://stackoverflow.com/questions/10762538/how-to-select-randomly-with-doctrine) for info.
  - An alternative is to make a special Twig function/filter to handle random records.
- ✔️ `{contenttype}/{id}`
- ✔️ `{contenttype}/{slug}`

### WHERE

- ✔️ `{key}: {value}`
- ✔️ `{key}: {operator}{value}` where operator is one of `%` `<` `>` `<=` `>=`
  - ⚠️ Currently the operations will not work for numbers/dates/etc. because of JSON-encoded values.
- ✔️ `author: {id}` replaces `ownerid` from Bolt 3

### WHERE advanced

- ❌ `{key}: {value1} || {value2}` or `{key}: {value1} && {value2}`
  - Currently only works for columns in `bolt_content`
- ❌ `"{key1} ||| {key2}": "{value1} ||| {value2}"`
  - Currently only works for columns in `bolt_content`
- ❌ `{key}: '>now'` and other special datetime selectors (e.g. `now`, `today`, `yesterday`, `last year`, `next thursday`).
  - Currently do not work at all.

### Special

- ✔️ `limit {number}`
- ✔️ `orderby {field}` or `orderby -{field}`
  - ⚠️ This works for fields that have only a single value (think of it as a column)
  - ✔️ Multiple sortorders do work (comma-separated)
- ✔️ `returnsingle`
- ✔️ `printquery`
  - ⚠️ Currently shows DQL. Let's see if we can show a _runnable query_ instead.

### Advanced

- ❌ `allowpaging`
- ❌ `{contenttype}/search/{num} where filter: {term}`
- ❌ `{ct1},{ct2},{ctN}/search/{num} where filter: {term}`

### Misc

- Implementations for locales, taxonomies and relationships.
- Proper integration with Config + ContentType data (such as max results per page).
- Remove all legacy stuff for real.


bobdenotter:
See #41

JarJak:
Should I put some feature requests here or in the related Epic?

bobdenotter:
> Should I put some feature requests here or in the related Epic?

This is primarily a branch by @xiaohutai to get a baseline, ported version of `setcontent` from v3 into v4. 

Improvements and fixes should go into the accompanying epic and/or subsequent issues. :-)

xiaohutai:
@JarJak I'll have to read up on that.

I've tried getting this up and running with the code that was available as a starting point (this may not have been the best idea). I think the most important part is that the results returned is consistent. So if a refactor/re-write is better, then we should definintely go for that.

(Any help is, of course, appreciated)

JarJak:
If we really need full entities in frontend, how about using Criteria instead of DQL?
https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/working-with-associations.html#filtering-collections

src/Storage/Query/SearchConfig.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * Get the config of all fields for a given content type.
+     *
+     * @param string $contentType
+     *
+     * @return array|false
+     */
+    public function getConfig($contentType)
```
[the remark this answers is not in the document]
xiaohutai: It's the config for a contenttype, `fields` is only one part of it. But I've removed this, as we need to look at this again, when implementing search.

src/Storage/Query/FrontendQueryScope.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Iterates over the main config and sets up what the default ordering should be.
+     */
+    protected function parseContenttypes()
+    {
+        $contentTypes = $this->config->get('contenttypes');
+        foreach ($contentTypes as $type => $values) {
+            $sort = $values['sort'] ?: '-publishedAt';
```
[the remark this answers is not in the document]
xiaohutai: Changed from `?:` to `??`. PHP would've shown a Notice otherwise.

xiaohutai:
@JarJak Thanks for the resources. I'm don't have much Doctrine experience, so I need to read up on that. Both Doctrine ORM and the Query code from Bolt 3 is quite new for me (it's the non-legacy setcontent version).

---

Also: I've added a skip rule for `AssignmentInConditionSniff.FoundInWhileCondition` to `easy-coding-standard.yml`. Not sure how you feel about that. See: https://github.com/bolt/four/pull/116/files#diff-a8b950982764fcffe4b7b3acd261cf91

src/Storage/Query/SelectQuery.php
```diff
@@ -0,0 +1,335 @@
...
+     * query execution time.
+     */
+    public function build(): QueryBuilder
+    {
+        $query = $this->qb;
+        if ($this->getWhereExpression()) {
+            $query->where($this->getWhereExpression());
+        }
+        foreach ($this->getWhereParameters() as $key => $param) {
+            // xiao: `Connection::PARAM_STR_ARRAY` seems incorrect to me
```
JarJak: yes it is incorrect. use:
https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Query/ParameterTypeInferer.php#L31

src/Storage/Query/Handler/SelectQueryHandler.php
```diff
@@ -0,0 +1,74 @@
...
+            $qb = $repo->getQueryBuilder();
+            $query->setQueryBuilder($qb);
+            // $query->setContentType($contentType);
+            $query->setContentType('content');
+
+            $query->setParameters($contentQuery->getParameters());
+            $contentQuery->runScopes($query);
+            $contentQuery->runDirectives($query);
+
+            // This is required. Not entirely sure why.
```
JarJak: because it builds all the parameters in query. the name of method is wrong if it is not meaningful.
rename to e.g. buildWhereParameters()

src/Storage/Query/Handler/SelectQueryHandler.php
```diff
@@ -0,0 +1,74 @@
...
+        $query->setSingleFetchMode(false);
+
+        foreach ($contentQuery->getContentTypes() as $contentType) {
+            $contentType = str_replace('-', '_', $contentType);
+
+            $repo = $contentQuery->getContentRepository();
+            $qb = $repo->getQueryBuilder();
+            $query->setQueryBuilder($qb);
+            // $query->setContentType($contentType);
+            $query->setContentType('content');
```
JarJak: rename to setAlias

src/Storage/Query/Handler/SelectQueryHandler.php
```diff
@@ -0,0 +1,74 @@
...
+            $query->build();
+
+            // Bolt4 introduces an extra table for field values, so additional
+            // joins are required.
+            $query->doReferenceJoins();
+            $query->doFieldJoins();
+
+            $query
+                ->getQueryBuilder()
+                ->andWhere('content.contentType = :ct')
```
JarJak: this is wrong place for this. it should be:
$query->setContentType($contentType);

src/Storage/Query/Handler/SelectQueryHandler.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        $set = new QueryResultset();
+        /** @var SelectQuery $query */
+        $query = $contentQuery->getService('select');
+        $query->setSingleFetchMode(false);
+
+        foreach ($contentQuery->getContentTypes() as $contentType) {
+            $contentType = str_replace('-', '_', $contentType);
+
+            $repo = $contentQuery->getContentRepository();
```
JarJak: is there a single content repo for all content types? then it should be constructed inside getService('select'):
inside SelectQuery:
private setQueryBuilder()
private setContentType() (setAlias?) (here also the where part from https://github.com/bolt/four/pull/116/files#diff-5ee7c092d66585260210d467fb97c416R50)

src/Storage/Query/Query.php
```diff
@@ -0,0 +1,102 @@
...
+        return null;
+    }
+
+    /**
+     * Helper to be called from Twig that is passed via a TwigRecordsView rather than the raw records.
+     */
+    public function getContentForTwig($textQuery, array $parameters = [])
+    {
+        // fix BC break
+        if (func_num_args() === 3) {
```
JarJak: this was for legacy, can be removed

src/Storage/Query/Handler/SelectQueryHandler.php
```diff
@@ -0,0 +1,74 @@
...
+            $query->doReferenceJoins();
+            $query->doFieldJoins();
+
+            $query
+                ->getQueryBuilder()
+                ->andWhere('content.contentType = :ct')
+                ->setParameter('ct', $contentType);
+
+            $result = $query
+                ->getQueryBuilder()
```
JarJak: i'm not sure query builder should ever leave boundaries of QueryInterface

src/Storage/Query/Handler/SelectQueryHandler.php
```diff
@@ -0,0 +1,74 @@
...
+
+            if ($result) {
+                $set->setOriginalQuery($contentType, $query);
+                $set->add($result, $contentType);
+            }
+        }
+
+        if ($query->getSingleFetchMode()) {
+            if ($set->count() === 0) {
+                return false;
```
JarJak: imo it should return null instead of false

src/Storage/Query/Query.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public function addScope($name, QueryScopeInterface $scope): void
+    {
+        $this->scopes[$name] = $scope;
+    }
+
+    /**
+     * @param string $name
+     */
+    public function getScope($name): ?QueryScopeInterface
```
JarJak: move param type from phpdoc (apply everywhere)

src/Storage/Query/QueryParameterParser.php
```diff
@@ -0,0 +1,344 @@
...
+            throw new \Exception('Mixed && and || operators are not supported', 1);
+        }
+    }
+
+    /**
+     * This handler processes 'triple pipe' queries as implemented in Bolt
+     * It looks for three pipes in the key and value and creates an OR composite
+     * expression for example: 'username|||email':'fred|||pete'.
+     */
+    public function multipleKeyAndValueHandler(string $key, $value, Expr $expr): ?Filter
```
JarJak: string $value

src/Storage/Query/QueryParameterParser.php
```diff
@@ -0,0 +1,344 @@
...
+    {
+        if (! mb_strpos($key, '|||')) {
+            return null;
+        }
+
+        $keys = preg_split('/ *(\|\|\|) */', $key);
+        $inputKeys = $keys;
+        $values = preg_split('/ *(\|\|\|) */', $value);
+        $values = array_pad($values, count($keys), end($values));
```
JarJak: it should throw exception if key and value count does not match

src/Storage/Query/QueryParameterParser.php
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * This handler processes multiple value queries as defined in the Bolt 'Fetching Content'
+     * documentation. It allows a value to be parsed to and AND/OR expression.
+     *
+     * For example, this handler will correctly parse values like:
+     *     'username': 'fred||bob'
+     *     'id': '<5 && !1'
+     */
+    public function multipleValueHandler(string $key, $value, Expr $expr): ?Filter
```
JarJak: string $value

src/Storage/Query/QueryParameterParser.php
```diff
@@ -0,0 +1,344 @@
...
+                $filter = $this->expr->{$exprMethod}($this->alias . $key, ':' . $placeholder);
+            }
+
+            $parts[] = $filter;
+            ++$count;
+        }
+
+        $filter = new Filter();
+        $filter->setKey($inputKeys);
+        $filter->setExpression(call_user_func_array([$expr, 'orX'], $parts));
```
JarJak: $filter->setExpression($expr->orX($parts));

src/Storage/Query/Handler/LatestQueryHandler.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ *  eg: 'pages/latest/10'
+ */
+class LatestQueryHandler
+{
+    public function __invoke(ContentQueryParser $contentQuery): QueryResultset
+    {
+        $contentQuery->setDirective('order', '-id');
+
+        return call_user_func($contentQuery->getHandler('select'), $contentQuery);
```
JarJak: return $contentQuery->getHandler('select')($contentQuery);

using call_user_func* is sooo PHP5

src/Storage/Query/QueryParameterParser.php
```diff
@@ -0,0 +1,344 @@
...
+            $comparison = 'orX';
+        }
+
+        $values = array_diff($values, ['&&', '||']);
+
+        $filterParams = [];
+        $parts = [];
+        $count = 1;
+
+        while ($val = array_shift($values)) {
```
JarJak: foreach ($values as $val)

src/Storage/Query/QueryParameterParser.php
```diff
@@ -0,0 +1,344 @@
...
+            $placeholder = $key . '_' . $count;
+            $filterParams[$placeholder] = $val['value'];
+            $exprMethod = $val['operator'];
+            $parts[] = $this->expr->{$exprMethod}($this->alias . $key, ':' . $placeholder);
+            ++$count;
+        }
+
+        $filter = new Filter();
+        $filter->setKey($key);
+        $filter->setExpression(call_user_func_array([$expr, $comparison], $parts));
```
JarJak: $filter->setExpression($expr->{$comparison}($parts));

src/Storage/Query/QueryParameterParser.php
xiaohutai: Note: can also be `integer`.

src/Storage/Query/QueryParameterParser.php
xiaohutai: can also be `integer`

src/Storage/Query/QueryParameterParser.php
xiaohutai: `...$parts` works

src/Storage/Query/Handler/SelectQueryHandler.php
xiaohutai: This pull request is based on Bolt 3's code, and the original did not use this `build` function. And I'm still not sure why it is needed for this, but not for Bolt _4_.

JarJak: Well, I don't know exactly how it worked in Bolt 3, but there was DBAL QB and here is the ORM one so it works differently ;)

xiaohutai: Yeah, that may be it actually 🤔

xiaohutai:
To be honest I think that I am not not qualified for this. Doctrine ORM and the Query layer are both rather new to me. With the rest of the changes open and still features not working, I am not sure how to solve those properly.

Hopefully someone else with more experience can work on it and/or refactor it. I believe that's why Bob asked me to make a start with it. @bobdenotter, you'll have the final say.

Thank you @JarJak for all the feedback. Appreciated!

bobdenotter:
As discussed, merging in right now! This issue is quite large heavy already, so let's split off the lingering issues into separate issues, to keep them smaller and more manageable. 

Thanks for the work @xiaohutai, and thanks @JarJak for the review and the offer to help out further! 👍 👍

JarJak:
@bobdenotter IMO this wasn't ready for merge... at least don't remove this branch please

PR #128, opened 2018-11-16
Travis eslint prettier

Ichag:
This PR configures eslint + prettier to lint js files.
To run it use `npm run lint` and for auto-fixing `npm run lint-fix`


Ichag:
prettier has a strong opinion about styles, it could be that is clashes with your opinion about how the code should look:

> By far the biggest reason for adopting Prettier is to stop all the on-going debates over styles. It is generally accepted that having a common style guide is valuable for a project and team but getting there is a very painful and unrewarding process. People get very emotional around particular ways of writing code and nobody likes spending time writing and receiving nits.

https://prettier.io/docs/en/why-prettier.html

PR #139, opened 2018-11-17
Tweaking PHPDOC for backend controllers

bobdenotter:


JarJak:
IMO those comments are not needed.

PR #142, opened 2018-11-18
ECS Config

JarJak:
`make cscheck` should be added to Travis


JarJak:
Now Travis fails, which is fine :)

PR #143, opened 2018-11-18
ECS Fixes

JarJak:
Not everything is fixed yet.

Includes config from #142

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,7 @@
         "api-platform/api-pack": "^1.1",
         "bolt/common": "^2.0",
         "cocur/slugify": "^3.1",
+        "doctrine/common": "2.8.1",
         "erusev/parsedown": "^1.7",
         "ezyang/htmlpurifier": "^4.10",
         "fzaninotto/faker": "^1.8",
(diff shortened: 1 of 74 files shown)


JarJak:
TBD: If we want a space after exclamation mark.

bobdenotter:
I see it removes a lot of `@param` and `@returns` from the docblocks, in favor of Type Hinting.. I think it's fine, but doesn't this leave people out in the cold who don't have PHPstorm or another "modern" IDE?

bobdenotter:
> TBD: If we want a space after each exclamation mark.

I'm fine with or without. This is one of these things that are easy to just adopt, and go with the flow.

bobdenotter:
> Also I don't know which IDE reads phpdoc, but skips type hints?

Neither do I, was just wondering out loud. Did a quick test and both PHPstorm and VSCode do it, so IMHO that is a good baseline: Both "professional" developers who use paid tools, as well as novices on VScode will be able to see the hints.. 

Let's roll with this! Should I merge #142 first, and then work on this until it's green?

JarJak:
Yes, let's merge #142 first (if it's config is fine for everyone). Additionally Travis config needs to be tweaked a bit.

JarJak:
37 errors ECS errors and 33 PHPStan errors. Not bad ;)

PR #152, opened 2018-11-22
PHPStan level bump

JarJak:
😈


JarJak:
>[ERROR] Found 15 errors   
Not that much! :)

PR #170, opened 2018-11-27
Remove references to Label component

### LINKED ISSUE #169: Errors when running `doctrine:fixtures:load`
```
> bin/console doctrine:fixtures:load -n
```

```
  > purging database
  > loading Bolt\DataFixtures\UserFixtures
  > loading Bolt\DataFixtures\TaxonomyFixtures
  > loading Bolt\DataFixtures\ContentFixtures

In ReferenceRepository.php line 141:

  Reference to: () does not exist
```

image

xiaohutai:
I _think_ it needs to be removed. If not, close this pull request. Partial fix for #169 

Edit: Hmmm... maybe someone needs to add the `_Partials/Label` instead 🤔

### DIFF
--- a/assets/js/Views/user.js
+++ b/assets/js/Views/user.js
@@ -6,7 +6,6 @@ import Theme from "../Components/User/Theme";
 import Password from "../Components/Password/Password";
 import Email from "../Components/Email/Email";
 import Text from "../app/editor/Components/Editor/Text/Text";
-import Label from "../app/editor/Components/Editor/_Partials/Label";
 
 /**
  * Register Components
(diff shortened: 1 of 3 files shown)


bobdenotter:
Actually, it might've been me who broke that.. the `Label` vue component was literally outputting a string of text, and I don't think it _should_ have more "behaviour" than that, so i've converted it back into a "plain" template partial in Twig. 

References can be removed, yes. :-)

nestordedios:
I think it should also be removed here https://github.com/bolt/four/blob/master/templates/_partials/fields/_label.html.twig. 
The Vue component is still being called there

nestordedios:
Thanks @xiaohutai 🎉

PR #173, opened 2018-11-27
Bump security-checker version to ^5.0

nestordedios:
Bump security-checker version as recommended here https://twitter.com/fabpot/status/1067085006542118913

Here #157 we changed the domain but the package version remained at v4.1.


bobdenotter:
Thanks! :-)

PR #230, opened 2019-01-02
Fixing empty titles on overview

bobdenotter:


src/Entity/Content.php
```diff
@@ -188,7 +189,7 @@ public function getSummary(): array
             'contenttype' => $this->getDefinition()->get('slug'),
             'slug' => $this->getSlug(),
             'title' => $this->magicTitle(),
-            'excerpt' => $this->magicexcerpt(),
```
JarJak: name of this method is magicExcerpt

JarJak:
Aren't magic traits deprecated?

bobdenotter:
@JarJak Not unless you come up with a viable alternative. Currently, we _need_ them in a number of places.

PR #234, opened 2019-01-04
Delete composer.lock

bobdenotter:


JarJak:
I would leave this file here until we get a stable version, then we can test both install and update scenarios.

bobdenotter:
@JarJak Yeah, let's close this for now. I do have an upcoming PR that greatly reduces the memory footprint of `composer update`, though.

Issue #208, opened 2018-11-30
Configure testing environment


JarJak:
PHPUnit, Behat, maybe PHPSpec?

bobdenotter:
See also #168

PR #236, opened 2019-01-04
Testing env with 1 simple test

JarJak:
First part of #208


JarJak:
Ok so symfony.lock is an important file :)

PR #240, opened 2019-01-08
Examples how to write Unit tests in 4 different ways

JarJak:
This shows 4 different ways to write Unit tests for `LoginFormAuthenticator`:
- PHPUnit + MockBuilder
- PHPUnit + Mockery
- PHPUnit + Prophecy
- PHPSpec + Prophecy

I personally prefer using PHPSpec + Prophecy (we use it a lot in our projects) but I have found that PHPUnit + MockBuilder is only a bit harder to use. The other two I would not recommend as they are overly complicated.


bobdenotter:
> I personally prefer using PHPSpec + Prophecy (we use it a lot in our projects)

In that case, PHPSpec + Prophecy has my vote. I have no personal preference, so experience with one of these is an important factor. 👍

JarJak:
Ok so let's stick with phpspec + prophecy then :)

PR #237, opened 2019-01-04
Some minor JS cleanup (with Prettier)

bobdenotter:
`prettier --single-quote  --trailing-comma --write assets/js/app/editor/Components/**/*`

PR #242, opened 2019-01-09
A bit of code cleanup

JarJak:


config/bolt/config.yaml
```diff
@@ -454,4 +454,4 @@ compatibility:
     # Setting this value to false is deprecated.
     template_view: true
     # Set to 'false' to enable using a newer version of the setcontent parser.
-    setcontent_legacy: true
+    setcontent_legacy: false
```
bobdenotter: This can be removed. Bolt 4 will have no option for the "legacy". 

In fact, before beta, we'll need to thoroughly go through the config, and determine what needs to be culled.

config/packages/security.yaml
```diff
@@ -32,4 +32,5 @@ security:
     access_control:
         # this is a catch-all for the admin area
         # additional security lives in the controllers
-        - { path: '^/(%app_locales%)/admin', roles: ROLE_ADMIN }
+        - { path: '^/bolt', roles: ROLE_ADMIN }
```
bobdenotter: We shouldn't hardcode the path to `/bolt. Right now we have it defined in `routes.yaml` and the annotations like `@Security("has_role('ROLE_ADMIN')")` in `src/Controller/Backend` make sure of the access control. If we do that here, it will make it easier for users to mess up by changing one, and leacing this as-is, leaving their install wide open to the world.

src/Configuration/Areas.php
```diff
@@ -23,7 +23,10 @@ public function __construct(Config $config)
         $this->initAreas();
     }
 
-    public function get($area, $key = null)
+    /**
+     * @return Collection|string
+     */
+    public function get(string $area, ?string $key = null)
```
bobdenotter: This change is fine for now, but we should strive to have only one return type, so we can use `: Collection` instead.

src/Controller/BaseController.php
```diff
@@ -49,8 +50,8 @@ protected function renderTemplate($template, array $parameters = [], ?Response $
         $parameters['user'] = $this->getUser();
 
         // Resolve string|array of templates into the first one that is found.
-        if ($template instanceof Collection) {
-            $template = $twig->resolveTemplate($template->toArray());
+        if (is_array($template)) {
```
bobdenotter: ~~Does this work? I think by `$template` is always `Collection`. If not, i'd like to work on that to make it so. :-)~~

I know see you've changed TemplateChooser to always use `->toArray()`. What's the benefit?

webpack.config.js
```diff
@@ -1,50 +1,48 @@
-var webpack = require('webpack');
-var WebpackBar = require('webpackbar');
-var Encore = require('@symfony/webpack-encore');
-var path = require('path');
-
-Encore
-
-  .addPlugin(new WebpackBar({
-    profile: Encore.isProduction() ? true:false,
+const webpack = require("webpack");
```
bobdenotter: I was working on these formatting changes in #237. I think I'd prefer `'` instead of `"`, like we do in our PHP code. 

(apart from that: All of our JS and SCSS files need some tidying up. :-) )

config/bolt/config.yaml
JarJak: How is this old config used in the new version?

config/packages/security.yaml
JarJak: so can we move it to a variable?

src/Controller/BaseController.php
JarJak: IMO Collection should rarely be returned. It is only an OOP wrapper for array-* functions. So if you need to do something with an array, wrap it with collect() and call methods, but after all return just an array with ->toArray()

webpack.config.js
JarJak: this is what linting did. I'm not sure how to configure it.

config/bolt/config.yaml
bobdenotter: I'd say the current version is about 95% copied from bolt 3, and needs some serious work. I wouldn't spend too much time on it in this PR, but we'll clean it up soon. Both for what Bolt 4 should be, and also to make it more "Symfony 4 app"-like. :-)

config/packages/security.yaml
bobdenotter: Yes, if you think that's better than keeping it in routing, i have no problems with it.

src/Controller/BaseController.php
bobdenotter: Ok, fine for me! 👍

webpack.config.js
bobdenotter: Ok, feel free to leave it like this, and i'll tweak ESlint / Prettier

 (or maybe Marcin can do that?)

config/packages/security.yaml
JarJak: See #245

PR #244, opened 2019-01-09
E2E tests

JarJak:


tests/e2e/features/login.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Logging
```
bobdenotter: This should be "Feature: Logging in"

PR #258, opened 2019-01-11
Hotfix: Store locale correctly from Profile

bobdenotter:


src/Controller/Backend/ProfileController.php
```diff
@@ -50,7 +51,7 @@ public function profileEditPost(Request $request, UrlGeneratorInterface $urlGene
         $user = $this->getUser();
         $displayName = $user->getDisplayName();
         $url = $urlGenerator->generate('bolt_profile_edit');
-        $locale = current($request->get('locale'));
+        $locale = Json::findScalar($request->get('locale'));
```
JarJak: locale is a json?

Issue #38, opened 2018-10-23
API: Flesh out API endpoints for frontend


bobdenotter:

JarJak:
Do we know which endpoints will be needed?

PR #268, opened 2019-01-16
Content API Endpoint and Behat tests

JarJak:
First part of #38.

Custom git repository added to composer.json because I wait for https://github.com/coduo/php-matcher/pull/143

### DIFF
--- a/src/Controller/Frontend/DetailController.php
+++ b/src/Controller/Frontend/DetailController.php
@@ -10,6 +10,7 @@ use Bolt\Repository\ContentRepository;
 use Bolt\Repository\FieldRepository;
 use Bolt\TemplateChooser;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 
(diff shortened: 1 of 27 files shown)


src/Controller/Frontend/DetailController.php
```diff
@@ -52,6 +56,9 @@ public function record(ContentRepository $contentRepository, FieldRepository $fi
             $recordSlug => $record,
         ];
 
+        dump($record);
+        dump($record->getFieldValues());
```
bobdenotter: This looks like something i could've put in there. ;-)

src/Entity/ContentMagicTrait.php
```diff
@@ -78,11 +101,13 @@ public function __call(string $name, array $arguments = [])
...
     private function magic(string $name, array $arguments = [])
     {
-        $magicName = 'magic' . $name;
+        $magicName = 'magic' . ucfirst($name);
 
         if (method_exists($this, $magicName)) {
             return $this->{$magicName}(...$arguments);
         }
+
+        throw new \RuntimeException(sprintf('Invalid field name or method call on %s: %s', self::class, $name));
```
bobdenotter: Ah, excellent! 👍

src/Entity/ContentMagicTrait.php
```diff
@@ -164,13 +189,7 @@ public function magicTitle(): string
...
         }
 
-        $title = trim(implode(' ', $titleParts));
-
-        if (empty($title)) {
-            return '(untitled)';
-        }
-
-        return $title;
+        return trim(implode(' ', $titleParts));
```
bobdenotter: I see why you did this, but this means i'll have to fix it in the Backend UI. 

It's better though, so i'll put it on my list. :-)

src/Controller/Frontend/DetailController.php
bobdenotter: True. :-)

PR #269, opened 2019-01-16
Cache config files

### LINKED ISSUE #263: Make Configuration files cachable.

bobdenotter:
Gives ~10% or ~6ms performance gain on simple pages. 

Before: 
screenshot 2019-01-16 at 22 20 42

After: 
screenshot 2019-01-16 at 22 19 53

Fixes #263

### DIFF
--- a/src/Configuration/Config.php
+++ b/src/Configuration/Config.php
@@ -27,11 +27,15 @@ class Config
     /** @var FilesystemCache */
     private $cache;
 
-    public function __construct(Stopwatch $stopwatch)
+    /** @var string */
+    private $projectDir;
+
+    public function __construct(Stopwatch $stopwatch, $projectDir)
     {
         $this->stopwatch = $stopwatch;
         $this->cache = new FilesystemCache();
-        $this->pathResolver = new PathResolver(dirname(dirname(__DIR__)), []);
+        $this->pathResolver = new PathResolver($projectDir, []);
+        $this->projectDir = $projectDir;
         $this->data = $this->getConfig();
     }
 
@@ -116,7 +120,7 @@ class Config
             }
         }
 
-        $envFilename = dirname(dirname(__DIR__)) . '/.env';
+        $envFilename = $this->projectDir . '/.env';
         if (file_exists($envFilename)) {
             $timestamps[$envFilename] = filemtime($envFilename);
         }
--- a/src/Configuration/Config.php
+++ b/src/Configuration/Config.php
@@ -34,9 +34,11 @@ class Config
     {
         $this->stopwatch = $stopwatch;
         $this->cache = new FilesystemCache();
-        $this->pathResolver = new PathResolver($projectDir, []);
         $this->projectDir = $projectDir;
         $this->data = $this->getConfig();
+
+        // @todo PathResolver shouldn't be part of Config. Refactor to separate class
+        $this->pathResolver = new PathResolver($projectDir, []);
     }
 
     private function getConfig(): Collection
--- a/src/Configuration/Config.php
+++ b/src/Configuration/Config.php
@@ -9,7 +9,7 @@ use Bolt\Configuration\Parser\BaseParser;
 use Bolt\Configuration\Parser\ContentTypesParser;
 use Bolt\Configuration\Parser\GeneralParser;
 use Bolt\Configuration\Parser\TaxonomyParser;
-use Symfony\Component\Cache\Simple\FilesystemCache;
+use Psr\SimpleCache\CacheInterface;
 use Symfony\Component\Stopwatch\Stopwatch;
 use Tightenco\Collect\Support\Collection;
 
@@ -24,16 +24,16 @@ class Config
     /** @var Stopwatch */
     private $stopwatch;
 
-    /** @var FilesystemCache */
+    /** @var CacheInterface */
     private $cache;
 
     /** @var string */
     private $projectDir;
 
-    public function __construct(Stopwatch $stopwatch, $projectDir)
+    public function __construct(Stopwatch $stopwatch, $projectDir, CacheInterface $cache)
     {
         $this->stopwatch = $stopwatch;
-        $this->cache = new FilesystemCache();
+        $this->cache = $cache;
         $this->projectDir = $projectDir;
         $this->data = $this->getConfig();
 
--- a/src/Configuration/Config.php
+++ b/src/Configuration/Config.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Bolt\Configuration;
 
 use Bolt\Common\Arr;
+use Bolt\Configuration\Parser\BaseParser;
 use Bolt\Configuration\Parser\ContentTypesParser;
 use Bolt\Configuration\Parser\GeneralParser;
 use Bolt\Configuration\Parser\TaxonomyParser;
@@ -26,9 +27,6 @@ class Config
     /** @var FilesystemCache */
     private $cache;
 
-    /** @var array */
-    private $timestamps = [];
-
     public function __construct(Stopwatch $stopwatch)
     {
         $this->stopwatch = $stopwatch;
@@ -45,7 +43,7 @@ class Config
             $data = $this->getCache();
         } else {
             $data = $this->parseConfig();
-            $this->setCache();
+            $this->setCache($data);
         }
 
         $this->stopwatch->stop('bolt.parseconfig');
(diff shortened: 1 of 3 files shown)


src/Configuration/Config.php
```diff
@@ -18,17 +20,71 @@ class Config
     /** @var PathResolver */
     private $pathResolver;
 
-    public function __construct()
+    /** @var Stopwatch */
+    private $stopwatch;
+
+    /** @var FilesystemCache */
```
JarJak: should be CacheInterface

src/Configuration/Config.php
```diff
@@ -18,17 +20,71 @@ class Config
...
+    private $cache;
+
+    /** @var array */
+    private $timestamps = [];
+
+    public function __construct(Stopwatch $stopwatch)
     {
+        $this->stopwatch = $stopwatch;
+        $this->cache = new FilesystemCache();
```
JarJak: please don't. cache driver must be replaceable. use cache interface as dependency or create cache decorator class

src/Configuration/Config.php
```diff
@@ -18,17 +20,71 @@ class Config
...
+
+    private function getConfig(): Collection
+    {
+        $this->stopwatch->start('bolt.parseconfig');
+
+        if ($this->validCache()) {
+            $data = $this->getCache();
+        } else {
+            $data = $this->parseConfig();
+            $this->setCache();
```
JarJak: setCache($data, $timestamps)

src/Configuration/Config.php
```diff
@@ -49,9 +105,27 @@ public function parseConfig(): Collection
...
         //'extensions' => $this->parseConfigYaml('extensions.yml'),
 
+        $this->setTimestamps($general, $taxonomy, $contentTypes);
+
         return $config;
     }
 
+    private function setTimestamps(...$configs): void
+    {
+        $this->timestamps = [];
```
JarJak: why timestamps are stored both in cache and here?

src/Configuration/Config.php
```diff
@@ -49,9 +105,27 @@ public function parseConfig(): Collection
...
+
+        foreach ($configs as $config) {
+            foreach ($config->getFilenames() as $file) {
+                $this->timestamps[$file] = filemtime($file);
+            }
+        }
+
+        $envFilename = dirname(dirname(__DIR__)) . '/.env';
```
JarJak: project root dir path must be replaceable

src/Configuration/Config.php
```diff
@@ -49,9 +105,27 @@ public function parseConfig(): Collection
         //'permissions' => $this->parseConfigYaml('permissions.yml'),
         //'extensions' => $this->parseConfigYaml('extensions.yml'),
 
+        $this->setTimestamps($general, $taxonomy, $contentTypes);
```
JarJak: $data['timestamps'] = $this->getConfigFilesTimestamps()

src/Configuration/Config.php
```diff
@@ -18,17 +20,71 @@ class Config
...
+    }
+
+    private function getConfig(): Collection
+    {
+        $this->stopwatch->start('bolt.parseconfig');
+
+        if ($this->validCache()) {
+            $data = $this->getCache();
+        } else {
+            $data = $this->parseConfig();
```
JarJak: [$data, $timestamps] = $this->parseConfigFiles();

src/Configuration/Config.php
```diff
@@ -18,17 +20,71 @@ class Config
...
+
+    /** @var array */
+    private $timestamps = [];
+
+    public function __construct(Stopwatch $stopwatch)
     {
+        $this->stopwatch = $stopwatch;
+        $this->cache = new FilesystemCache();
         $this->pathResolver = new PathResolver(dirname(dirname(__DIR__)), []);
+        $this->data = $this->getConfig();
```
JarJak: there are 2 places where $this->data is filled, remove it from getConfig()

src/Configuration/Config.php
```diff
@@ -18,17 +20,71 @@ class Config
...
+    private $cache;
+
+    /** @var array */
+    private $timestamps = [];
+
+    public function __construct(Stopwatch $stopwatch)
     {
+        $this->stopwatch = $stopwatch;
+        $this->cache = new FilesystemCache();
         $this->pathResolver = new PathResolver(dirname(dirname(__DIR__)), []);
```
JarJak: looks like PathResolver is not needed anymore here

JarJak:
Maybe use this?
https://symfony.com/doc/current/components/config/caching.html

src/Configuration/Config.php
bobdenotter: I tried to use `CacheInterface` as DI, but couldn't get it to work. I'll update the other points you mention, and then maybe you could take a look to see how to do that?

src/Configuration/Config.php
bobdenotter: We gather them here, in order to store them in the cache. 

On the next invocation, the values are fetched from the cache, and compared to the current timestamps to determine if the cache is still valid. 
This way, if _any_ of the used config files are modified through any means, the cache is invalidated. The overhead of this is very minimal.

src/Configuration/Config.php
bobdenotter: Agree completely, but I suggest we do that at once for all of the places where we do this. (and after we've discussed _how_ to do this. :-) )

src/Configuration/Config.php
bobdenotter: Correct, will do.

src/Configuration/Config.php
JarJak: but why they are stored in the class property? afair it can be just a variable, because checking timestamps is done only once

src/Configuration/Config.php
JarJak: you can just use %kernel.project_dir% here, no?

src/Configuration/Config.php
bobdenotter: Actually, `this->config->getPath` is used in a number of places, but it should be moved out of `Config`, but that will be a separate PR.

src/Configuration/Config.php
bobdenotter: Ah, yes.. upcoming push will remove it as class property.

src/Configuration/Config.php
bobdenotter: How would I do that?

src/Configuration/Config.php
bobdenotter: Removed it from `parseConfig()` :-)

src/Configuration/Config.php
```diff
@@ -18,40 +21,109 @@ class Config
...
+                $timestamps[$file] = filemtime($file);
+            }
+        }
+
+        $envFilename = dirname(dirname(__DIR__)) . '/.env';
+        if (file_exists($envFilename)) {
+            $timestamps[$envFilename] = filemtime($envFilename);
+        }
+
+        $this->cache->set('config_timestamps', $timestamps);
```
JarJak: $this->cache is not yet set here, that's why parseConfigFiles should return it too

src/Configuration/Config.php
JarJak: ok, add a todo then

src/Configuration/Config.php
```diff
@@ -18,40 +21,109 @@ class Config
...
+
+    private function getConfig(): Collection
+    {
+        $this->stopwatch->start('bolt.parseconfig');
+
+        if ($this->validCache()) {
+            $data = $this->getCache();
+        } else {
+            $data = $this->parseConfig();
+            $this->setCache($data);
```
JarJak: $this->updateCache($data, $timestamps)

src/Configuration/Config.php
JarJak: inject it in constructor

src/Configuration/Config.php
```diff
@@ -18,40 +21,109 @@ class Config
...
+    {
+        if (! $this->cache->has('config_cache') || ! $this->cache->has('config_timestamps')) {
+            return false;
+        }
+
+        $timestamps = $this->cache->get('config_timestamps');
+
+        foreach ($timestamps as $filename => $timestamp) {
+            if (filemtime($filename) > $timestamp) {
+                return false;
```
JarJak: IMO cache should store each config file separately and the file has been modified and then just set cache->expiresAt() as PSR-6 says

bobdenotter: @JarJak That won't work in this case, if only for the example pointed out by Marcin yesterday: Let's say the `.env` file changes. In that case not only _that_ cached part should expire, but the entire cached config. 

In this case, i think it'll also make it quite. bit more complex. Let's do that only if we're a bit further along, and we actually measure there's performance to be gained here. :-)

src/Configuration/Config.php
bobdenotter: Done. :-)

PR #270, opened 2019-01-17
Fix Taxonomy tests

JarJak:


src/DataFixtures/TaxonomyFixtures.php
```diff
@@ -41,45 +41,19 @@ private function loadTaxonomies(ObjectManager $manager): void
...
-                );
-
-                $manager->persist($taxonomy);
-                $reference = 'taxonomy_' . $taxonomyDefinition['slug'] . '_' . Str::slug($key);
-                $this->addReference($reference, $taxonomy);
             }
         }
     }
-
-    private function getOptions()
```
bobdenotter: Why remove these?

JarJak: are they needed for something?

bobdenotter: I think it makes sense for the dummy content "fixtures" to have a bunch of tags. Later on, it'll help when we need to add tests for dealing with tags too. So, i'd prefer to keep them.

bobdenotter: @JarJak True, but during development, it's useful to just have a "bunch of tags" available. Can we add a flag to the fixtures code, that allows toggling it?

PR #272, opened 2019-01-17
Fixes for JSON placement on Dashboard

bobdenotter:


tests/e2e/pages/dashboard.js
```diff
@@ -8,6 +8,8 @@ class DashboardPage extends BasePage {
 
     this.profile_text = $('#toolbar .is-profile');
     this.header = $('.admin__header--title strong');
+    this.records = element.all(by.css('.listing__row'));
+    this.first_record = element(by.css('.listing__row'));
```
JarJak: can we stick to $ ?

bobdenotter: @JarJak I was meaning to ask you.. If I use `$('.admin__header--title strong');`, I get the following notices: 

```
W/element - more than one element found for locator By(css selector, aside ul) - the first result will be used
```

Even if we ignore, that, I'm not sure how to distinguish (as in the above example) between one element and an array of elements. Do you know how to do that?

bobdenotter: Or else, i'm getting this, for `And there are "equal 6" "records" elements`

```
   ✖ And there are "equal 6" "records" elements # node_modules/kakunin/dist/step_definitions/elements.js:279
       TypeError: this[elementName].count is not a function
```

JarJak: And about that notice... You can be more specific in CSS selector (use ID, :first-child etc.) or just ignore this notice :)

bobdenotter: Ah, cool.. I think `$$()` is what i was looking for. Will work on this PR.

bobdenotter: Updated and rebased to master

PR #275, opened 2019-01-18
Fix pagination links in admin

JarJak:
needs E2E update


JarJak:
I'll merge those 2 PRs into one as they are fixing themselves :D

PR #278, opened 2019-01-20
[hotfix] Don't break when trying to save record without Taxonomies

bobdenotter:


JarJak:
could you add a test for it?

bobdenotter:
> could you add a test for it? 

Will be done de facto when I add tests for the new `GetContenFromFormdata` that'll be part of #279  :-)

PR #281, opened 2019-01-21
[hotfix] Don't break if Contenttype has no image

bobdenotter:


src/Entity/ContentMagicTrait.php
```diff
@@ -200,11 +200,7 @@ public function magicImage(): array
             }
         }
 
-        return [
-            'filename' => '',
-            'alt' => '',
-            'path' => '',
-        ];
+        return [];
```
JarJak: can we return null instead?

bobdenotter: It breaks Vue, if we do. I've added a `@todo` instead. :-)

Issue #151, opened 2018-11-22
MagicContentTrait DX improvements


JarJak:
Splitted from https://github.com/bolt/four/pull/148

- [x] I really don't like all those "magic" methods in ContentMagicTrait. Can we at least rename them to be less magical? :)
- [x] if we could make $this->fields a name-indexed array, using this method be much faster (simple isset instead of foreaching).
- [x] rename to get to getField, has to hasField

JarJak:
Also in MediaFactory:
- can you replace it with an \Exception and add @todo? At least for getting trace where it happened :)

bobdenotter:
> I really don't like all those "magic" methods in ContentMagicTrait. Can we at least rename them to be less magical? :)

In Bolt 3 we also had these, only they weren't named as such. And IMHO that gave a bigger problem. When you'd get `title`, you'd _either_ get the magic title-like result, _or_ the actual field named `title`. It was a source of confusion. 

I've named them like this to make it _explicit_.. You can now fetch what you need, and you'll know what to expect as the result. 
What would you change it to, instead? 

> if we could make $this->fields a name-indexed array, using this method be much faster (simple isset instead of foreaching).

Yes, certainly.. What we have now is basically the generated piece of code from `make:entity`. 
I was thinking about turning it into a `Collection`, too. 

> rename to get to getField, has to hasField

Will do. 

> can you replace it with an \Exception and add @todo? At least for getting trace where it happened :)

Allright, allright.. ;-)

JarJak:
>What would you change it to, instead?
I would like to get rid of them, I don't feel they are really needed (at least inside an entity, they can be moved to some service).

bobdenotter:
>  I don't feel they are really needed 

The are needed on the overview pages to show titles, excerpts, thumbnails. They are needed in the sidebar menu to link the 'last 5 items'.. In the "frontend" they're used for listing pages as well as 'generic content pages'. 

I'm open to renaming them, but they are most certainly required. :-)

bobdenotter:
> if we could make $this->fields a name-indexed array, using this method be much faster (simple isset instead of foreaching).

Over the weekend i worked on `getLocalisedFields()` and related methods. These return named fields, as a `Collection`.

JarJak:
I know they are used here and there. But an Entity is not a good place for logic like this. That's why I'd vote for moving them away.

bobdenotter:
> That's why I'd vote for moving them away.

[...]

PR #285, opened 2019-01-21
Implement Translatable Doctrine Extension

JarJak:
- refactor $this->fields to a name-indexed array to make field existence check much faster (simple isset instead of foreaching)
- rename get to getField, has to hasField
- implement Translatable instead of reinventing the wheel

First part of #151


src/Entity/Field.php
```diff
@@ -190,7 +196,7 @@ public function getValue(): ?array
      *
      * @return array|mixed|null
      */
-    public function getFieldValue()
+    public function getFlatenValue()
```
bobdenotter: Really like where this PR is going. But, this is a bit of an odd Method name.

JarJak: What do you propose? It returns single value if count($this->getValues()) === 1, array if it's >1 and null if it is zero.

bobdenotter: Then it should be `getFlatValue` or `getFlattenedValue`. "Flaten" looks like a typo. 😇

bobdenotter:
Works like a charm for me! 

Could we change the name of the new table from `ext_translations` to perhaps `bolt_translations? I found some info about that here: https://github.com/Atlantic18/DoctrineExtensions/issues/1129

bobdenotter:
To me, "a lot of queries" is no problem. "slow queries" are. ;-)

If O look at it, i see: 

schermafbeelding 2019-01-23 om 09 30 52

So, about 5% of resources are taken by doctrine now (4.3 ms / 86 ms). Of those 4.3 ms the majority happens in the "sibebar", which we'll be caching. So, i am fine with this solution. 

schermafbeelding_2019-01-23_om_09_30_39

I see `KnpLabs/DoctrineBehaviors` is "looking for maintainers", which doesn't say anything conclusive, but it's not a good sign that it's actively maintained. 

Finally, `Atlantic18/DoctrineExtensions` has `Tree`, which is something i would really, really love to have for Content, so we can make hierarchical content out of the box.

JarJak:
Interesting... this is what I see:
zrzut ekranu z 2019-01-23 09-47-58

bobdenotter:
Hmm, I'm not sure if I was looking at the correct branch earlier.. Just now i tried again, and i'm getting similar numbers as you.

Looks like it does a separate query for each field that is translated. In my case it's especially the sidebar that does a lot of queries. I wonder if this is "the price we pay" for baking in i18n, or if there's something we can do about it. 

I propose we keep it like this for now, then I add caching to the sidebar, and then it'll be easier to properly inspect which queries it does exactly. Is that a plan?

JarJak:
@bobdenotter I've made db tables prefix fully configurable

src/Entity/ContentLocalizeTrait.php
```diff
@@ -19,71 +19,68 @@ public function getLocales()
...
-
-        // Fourth: see if we can create the field on the fly
-        if (! empty($definition)) {
-            return Field::factory($definition, $name);
-        }
-
-        // Alas, return an empty generic Field
-        return Field::factory(['type' => 'generic']);
-    }
+//    public function getLocalizedFields(string $locale = '', bool $fallback = true): Collection
```
bobdenotter: If it's working correctly, shouldn't we just remove this chunk of commented-out code?

JarJak: removed

Issue #282, opened 2019-01-21
Setcontent bug `setcontent homepage = "homepage/1"`


bobdenotter:
If i do this, I get an error: 

```
    {% setcontent homepage = "homepage/1" %}
```
![screenshot 2019-01-21 at 19 48 06](https://user-images.githubusercontent.com/1833361/51493383-81793080-1db5-11e9-8ea0-7bb8572498b2.png)

xiaohutai:
Fixed in #286

Issue #283, opened 2019-01-21
Wrong result for `setcontent page = "pages" where {'id': 16}`


bobdenotter:
If i go to `/page/16`, with page # 16 being a `published` page, and I edit `record.twig` to look like this: 

```twig

{% block main %}

    <h1>{{ record.title }}</h1>

    {{ dump(record.title) }}
    {{ dump(record.id) }}

    {% setcontent page = "pages" where {'id': 16} %}
    {{ dump(page|first.title) }}
    {{ dump(page|first.id) }}
```

I would expect the two dumps to look similar. This is the result: 

![screenshot 2019-01-21 at 19 55 18](https://user-images.githubusercontent.com/1833361/51493723-8094ce80-1db6-11e9-8906-ed815719d220.png)

So, they give a different result, coming from a different record, and the second one has an additional `, UTF-8`, like it's passed through `Twig_Markup` twice.

bobdenotter:
Partial fix (for the `, UTF-8` issue) coming up.

About the other, it seems to "skip" the `where`: 

```
{% setcontent page = "pages" where { 'id': 16 } printquery %}

=> SELECT content FROM Bolt\Entity\Content content WHERE content.status = :status_1 
```

xiaohutai:
Fixed in #286

PR #286, opened 2019-01-22
Repair `setcontent`

xiaohutai:
Fixes issues #282 and #283

This PR removes some type hints which is not a nice thing to do in general, but I don't know how to circumvent these issues in a nice way. Below is a summary.

**Notes**

- At the moment, `setcontent` can return `null` or `Content` for single item when using `{contenttype}/{id}` or `QueryResultSet` (empty or not) for multiple items.
  
  This was changed in https://github.com/bolt/four/commit/82244ab#diff-5ee7c092d66585260210d467fb97c416

- Function `parseValue` can get an integer when using `where { id: 1 }`.

  This was changed in https://github.com/bolt/four/commit/ce98e9bbaa993ba452d23bcae463f07178a3f3c6#diff-9501bf2bce92196e20149014404e59fb

- Function `getContentForTwig` actually gets three parameters:
  1. the base part like `pages` or `pages/1`
  2. parameters like `printquery`, `paging`
  3. actual `where` parameters taken from `... where { foo: bar } ...`

  This was changed in https://github.com/bolt/four/commit/1624cdc#diff-47ee526d7c161782611f5afd2e66c76d

### DIFF
--- a/src/Storage/Query/Handler/SelectQueryHandler.php
+++ b/src/Storage/Query/Handler/SelectQueryHandler.php
@@ -14,7 +14,7 @@ use Bolt\Storage\Query\SelectQuery;
  */
 class SelectQueryHandler
 {
-    public function __invoke(ContentQueryParser $contentQuery): ?QueryResultset
+    public function __invoke(ContentQueryParser $contentQuery)
     {
         $set = new QueryResultset();
         /** @var SelectQuery $query */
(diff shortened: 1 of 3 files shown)


JarJak:
>the base part like pages or pages/1
>parameters like printquery, paging
>actual where parameters taken from ... where { foo: bar } ...

Could you explain those e.g. in the comment above getContentForTwig?

xiaohutai:
No need for typehinting in phpdoc if they already exists, right?

JarJak:
@xiaohutai if checkers won't complain, yes

xiaohutai:
It's stricter than I thought 😅

bobdenotter:
(And my apologies for breaking stuff ;-) )

xiaohutai:
No, worries. It's actually to show that it's PHPStan complaining about things. Unless those parts are coded differently we can't really add those nice type hints yet.

PR #293, opened 2019-01-24
Add `allowempty` flag to prevent save record with empty status

### LINKED ISSUE #291: [BUG] Status select allows selecting empty status
Steps to reproduce:
---

1. Go to edit page
2. Change status to same as it was before
3. Click save

Expected result:
----

Status stays unchanged.

Actual result:
---

Selector makes the field empty. Clicking save gives 500:
```
Argument 1 passed to Bolt\Entity\Content::setStatus() must be of the type string, bool given, 
called in src/Controller/Backend/ContentEditController.php on line 109
```

bobdenotter:
Fixes #291

### DIFF
--- a/templates/_partials/fields/select.html.twig
+++ b/templates/_partials/fields/select.html.twig
@@ -15,10 +15,10 @@
     {% set multiple = 'false' %}
 {% endif %}
 
-{% if field.allowempty is defined %}
+{% if allowempty is not defined and field.definition is defined %}
     {% set allowempty = field.definition.get('allowempty') ? 'true' : 'false' %}
 {% else %}
-    {% set allowempty = allowempty ? 'true' : 'false' %}
+    {% set allowempty = allowempty|default() ? 'true' : 'false' %}
 {% endif %}
 
 {% block field %}
--- a/templates/_partials/fields/select.html.twig
+++ b/templates/_partials/fields/select.html.twig
@@ -1,9 +1,12 @@
 {% extends '@bolt/_partials/fields/_base.html.twig' %}
 
-{# Check Attributes #}
-{% if form is not defined %}
-    {% set form = '' %}
-{% endif %}
+{# On top of the attributes that get set/defined in `_base.html.twig`,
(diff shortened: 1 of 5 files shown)


templates/_partials/fields/select.html.twig
```diff
@@ -15,6 +15,12 @@
     {% set multiple = 'false' %}
 {% endif %}
 
+{% if field.allowempty is defined %}
```
JarJak: Shouldn't it be `if allowempty is not defined` ?

bobdenotter: Yeah, this logic is wonky.. Fixed in latest commit! 👍

templates/_partials/fields/select.html.twig
```diff
@@ -15,6 +15,12 @@
     {% set multiple = 'false' %}
 {% endif %}
 
+{% if allowempty is not defined and field.definition is defined %}
```
JarJak: is it possible to have field without it's definition?

bobdenotter: Currently, yes. The language switcher, and the `<select>` for taxonomies, as well as the forms on the "profile page" and the "edit media" page use the same field twig components, only they aren't tied to a Field, so they don't have a definition per se.

JarJak: Ok, so this can only check `field is defined` instead of `field.definition is defined` (because it will probably break if `field` is not set at all).

This check should be probably moved to `_base.html.twig` as mentioned in https://github.com/bolt/four/issues/294#issuecomment-457337734 or, if `allowempty` is something special for select field, document it with comment on top of this template.

bobdenotter: The `allowempty` check is specific for select fields, so i've kept it here, but I've added a comment describing it. 👍

JarJak: It will still break if field is not defined

bobdenotter:
Cleaned up!

PR #301, opened 2019-01-26
Translation cleanup

### LINKED ISSUE #167: ⭐️ Set up Translation workflow

bobdenotter:
Cleans up translations, allows for easier translating in Bolt itself, and integrates with Loco (localize.biz) for community translations.

Closes #167


composer.json
```diff
@@ -23,6 +23,8 @@
         "miljar/php-exif": "^0.6.4",
         "nelmio/cors-bundle": "^1.5",
         "nesbot/carbon": "^1.34",
+        "php-http/httplug-pack": "^1.1",
```
JarJak: should be in require-dev

config/bundles.php
```diff
@@ -23,4 +23,6 @@
     Knp\Bundle\MenuBundle\KnpMenuBundle::class => ['all' => true],
     Translation\Bundle\TranslationBundle::class => ['all' => true],
     Symfony\WebpackEncoreBundle\WebpackEncoreBundle::class => ['all' => true],
+    Translation\PlatformAdapter\Loco\Bridge\Symfony\TranslationAdapterLocoBundle::class => ['all' => true],
```
JarJak: only dev and local

tests/e2e/features/edit_record.feature
```diff
@@ -55,10 +54,6 @@ Feature: Edit record
             | _locale | nl |
         Then there is element "title" with text "Changed title NL"
 
-        When I visit the "single_record" page with parameters:
```
JarJak: why removed?

composer.json
bobdenotter: In that case the php-translation/loco-adapter needs to go into `dev` too, and I'll need to make sure bundles.php and .yml files don't break on 'non dev'. Will do.

tests/e2e/features/edit_record.feature
bobdenotter: Because it's incorrect. The correct version is sligthly below, line 67/62.

I assumed this was a copy/paste mistake.

config/packages/dev/translation.yaml
```diff
@@ -0,0 +1,27 @@
...
+translation:
+    locales: ['en', 'nl', 'es', 'fr', 'de', 'pl', 'it', 'pt_BR']
+    edit_in_place:
+        enabled: false
+        config_name: app
+    configs:
+        app:
+            dirs: ["%kernel.project_dir%/templates", "%kernel.project_dir%/src"]
+            output_dir: "%kernel.project_dir%/translations"
+            excluded_names: ["*TestCase.php", "*Test.php"]
```
JarJak: we don't need to exclude anything. tests are in separate dir, same with logs, cache.

bobdenotter:
PR Cleaned up, should be GTG!

bobdenotter:
🚀

PR #303, opened 2019-02-01
[WIP] Setcontent refactor

stryxx:


src/Twig/Node/SetcontentNode.php
```diff
@@ -41,7 +43,7 @@ public function __construct($name, Node $contentType, ArrayExpression $arguments
     /**
      * {@inheritdoc}
      */
-    public function compile(Compiler $compiler): void
+    public function compile(\Twig_Compiler $compiler): void
```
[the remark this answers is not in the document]
stryxx: Ok, but IDE highlighted it in red

src/Storage/Query/Query.php
```diff
@@ -11,7 +11,7 @@ class Query
     /** @var ContentQueryParser */
     protected $parser;
 
-    /** @var array */
+    /** @var array<string> */
```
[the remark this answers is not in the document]
stryxx: Yes, PhpStorm points here a warning, and this entry removes the warning

public/theme/skeleton/partials/_sub_fields.twig
```diff
@@ -41,6 +41,8 @@
 
     {#{{ dump(record.fields) }}#}
 
+    {% setcontent test = 'homepage/dolores-maxime-eveniet-delectus' %}
```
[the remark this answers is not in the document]
stryxx: Ok I'm going to test during implementation.

src/Entity/Content.php
```diff
@@ -458,4 +459,42 @@ public function __call(string $name, array $arguments = []): Field
             throw new \RuntimeException(sprintf('Invalid field name or method call on %s: %s', $this->__toString(), $name));
         }
     }
+
+    public function serialize(): string
```
JarJak: use jsonSerialize() from ContentExtrasTrait

src/Repository/ContentRepository.php
```diff
@@ -27,6 +27,13 @@ public function __construct(RegistryInterface $registry)
         parent::__construct($registry, Content::class);
     }
 
+    public function createQueryBuilder($alias, $indexBy = null): QueryBuilder
```
JarJak: it's the same as in the extended class

src/Storage/Query/Parser/ContentFieldParser.php
```diff
@@ -0,0 +1,163 @@
...
+                )),
+            ],
+        ];
+
+        return array_merge($filterFields, $contentFields[$contentType]);
+    }
+
+    private function getTypeForField(string $contentType, DeepCollection $fieldConfiguration): Type
+    {
+        switch ($fieldConfiguration['type']) {
```
JarJak: what about user-defined field types?

JarJak:
@stryxx rebase to master

public/theme/skeleton/partials/_sub_fields.twig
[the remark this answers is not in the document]
stryxx: I done it.

src/Entity/Content.php
stryxx: This is not used now.

src/Storage/Query/Parser/ContentFieldParser.php
stryxx: For user-defined field types we should implement some mapper, but it is impossible now, because we cannot implement dynamic DiscriminatorMap for Entity. If new version of doctrine will make it possible, we have to do it, or in the nearest future we have to do by us.

JarJak:
Checklist copied from #116. API won't be 100% the same, but functionality should be kept.


### Basic

- [ ] `{contenttype}`
- [ ] `{contenttype}/first/{number}`
- [ ] `{contenttype}/last/{number}`
- [ ] `{contenttype}/random/{number}`
- [ ] `random` is not going to work with Doctrine ORM queries, see [answer on Stack Oveflow](https://stackoverflow.com/questions/10762538/how-to-select-randomly-with-doctrine) for info. 
  - ⚠️ An alternative is to make a special Twig function/filter to handle random records.
- [ ] `{contenttype}/{id}`
- [ ] `{contenttype}/{slug}`

### WHERE

- [ ] `{key}: {value}`
- [ ] `{key}: {operator}{value}` where operator is one of `%` `<` `>` `<=` `>=`
  - ⚠️ Currently the operations will not work for numbers/dates/etc. because of JSON-encoded values.
- [ ] `author: {id}` replaces `ownerid` from Bolt 3

### WHERE advanced

- [ ] `{key}: {value1} || {value2}` or `{key}: {value1} && {value2}`
- [ ] `"{key1} ||| {key2}": "{value1} ||| {value2}"`
- [ ] `{key}: '>now'` and other special datetime selectors (e.g. `now`, `today`, `yesterday`, `last year`, `next thursday`).

### Special

- [ ] `limit {number}`
- [ ] `orderby {field}` or `orderby -{field}`
  - ⚠️ This works for fields that have only a single value (think of it as a column)
  - [ ] Multiple sortorders (comma-separated)
- [ ] `returnsingle`
- [ ] `printquery`
  - ⚠️ Currently shows DQL. Let's see if we can show a _runnable query_ instead.


### Advanced

- [ ] `allowpaging`
- [ ] `{contenttype}/search/{num} where filter: {term}`
- [ ] `{ct1},{ct2},{ctN}/search/{num} where filter: {term}`


### Misc

- [ ] Implementations for locales, taxonomies and relationships.
- [ ] Proper integration with Config + ContentType data (such as max results per page).

tests/api/content/get_content.feature
```diff
@@ -43,84 +37,24 @@ Feature: Get content with API
...
-     "extras": {
-        "link": @string@,
-        "editLink": @string@,
-        "@*@": "@*@"
-     },
      "@*@": "@*@"
   }
   """
 
-  Scenario: As a user I fetch contents in JSON+LD format
```
JarJak: Why jsonld has been deleted?

bobdenotter:
@stryxx Gentle nudge.. Any progress on this PR? :-)

PR #310, opened 2019-02-04
[hotfix] Fix "Unused binding" breakage introduced in SF 4.2.3 / Terser 3.16

bobdenotter:
After updating to SF 4.2.3, I get this error: 

```
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!
!!  In ResolveBindingsPass.php line 48:
!!
!!    Unused binding "$defaultLocale" in service "Doctrine\ORM\Query\Expr".
!!
!!
!!
Script @auto-scripts was called via post-update-cmd
```

This PR comments out the line, with an added `@todo` to check if we need it in the future.


config/services.yaml
```diff
@@ -19,7 +19,8 @@ services:
                             # The best practice is to be explicit about your dependencies anyway.
         bind:               # defines the scalar arguments once and apply them to any service defined/created in this file
             $locales: '%app_locales%'
-            $defaultLocale: '%locale%'
+            # @todo: This breaks, starting with SF 4.2.3. Check if needed, and if so, fix properly.
+            # $defaultLocale: '%locale%'
```
JarJak: this can be removed it tests won't fail

bobdenotter: Tests pass locally, so i've removed it.

PR #311, opened 2019-02-04
Lock Terser version

JarJak:
Due to https://github.com/vuejs/vue-cli/issues/3407


JarJak:
Looks like #310 needs to get merged first

PR #317, opened 2019-02-07
Improve API Response and remove Magic from Content

### LINKED ISSUE #38: API: Flesh out API endpoints for frontend

### LINKED ISSUE #151: MagicContentTrait DX improvements

### LINKED ISSUE #182: Make "Content" less magical

### LINKED ISSUE #271: Remove jsonrecords Twig helper
Content serialization should be done with Serializable or JsonSerializable interface. JSONed Content should look the same in Twig as well as in API response. 

So:
```twig
{{ jsonrecords(records) }}
```
Should be replaced with:
```twig
{{ records|json_encode }}
```

JarJak:
Provides changes described here: https://github.com/bolt/four/issues/151#issuecomment-456311890

I've probably broke some Vue components here, @marcingajda please help :)

RFC: `{{ record|excerpt }}` vs `{{ record | excerpt }}`. Out frontend devs prefer the latter but I want to know your opinion too. @xiao @anketwokings @sbonardt @jadwigo
 
Closes #38 
Closes #151
Closes #182
Closes #271


sbonardt:
Hi @JarJak 

What's the rationale behind {{ record | excerpt }} ? And why do your frontend dev's prefer that RFC?

JarJak:
@sbonardt I guess it is more readable with spaces. On the other hand Twig docs doesn't recommend using spaces.

sbonardt:
@JarJak Ah, ok. Then I oppose the RFC. 😄 I like to stick to the conventions/recommendations

xiaohutai:
I prefer `foo|filter` without the spaces:

- `foo | bar` looks like two separate variables (compare with `foo & bar`, and `foo || bar`).
- I also wouldn't put places around `foo . bar` (weak argument as it's not really the same thing, but still).
- Twig docs also do it without spaces.
- If we would do it, we also need to re-write all (upcoming) Bolt docs to do that too? I feel there will be a gap between Twig docs and Bolt docs.

JarJak:
Ok then, I will revert filters to not use spaces.

JarJak:
Finally green :D

Issue #163, opened 2018-11-26
⭐️ Make "Relations" work


bobdenotter:

JarJak:
Can we make it easy to fetch relations easy from owning and inversed side? Like:
`{{ record.related(ct, bidirectional=false) }} -> Content[]`
And fetching single (first) related content, like:
`{{ record.related(ct, first=true) }} -> Content`

JarJak:
Have in mind relations in preview mode

JarJak:
`related` filter will be done with #151

bobdenotter:
Relations as such work. Closing this one for now. Feel free to re-open for more specific functionality.

PR #321, opened 2019-02-11
[RFC] Related content

### LINKED ISSUE #192: More elastic relations between content types
(Multiple relations to same content type but with different names or at least multiple sortable and limitable)

JarJak:
Closes #192 

Features:
---
- relations are decoupled from Content
- named relations allowed
- multi contenttype relations allowed
- sortable relations allowed
- relations are bidirectional by default, but owning side takes precedence

Configuration:
---
Config (from contentypes.yml):
```yaml
entries:
    relations:
        related_pages_and_blogposts: # name of relation, defaults to related content type if to_contenttypes is not set
            to_contenttypes: [pages, blogposts] # optional, see above
            sortable: true # default: false
            multiple: true # default: false
            min: 1
            max: 5
```
Or just simple config with defaults:
```yaml
entries:
    relations:
        pages: ~
        blogposts: ~
```

Twig examples:
---
```twig
{{ record|related(ct) }} -> Content[]
{{ record|related(name, bidirectional=false, limit=10) }} -> Content[]
{{ record|related_first(name) }} -> Content|null
{{ record|related_all }} or {{ related_content_all(record) }} -> Array<relation-name><Content>
```

API examples:
---
TBD: I thought about something like `/related/{from-id}/{to-contenttype}`

TODO:
---
- [x] write examples in skeleton
- [x] fixtures
- [x] readme
- [x] api endpoints
- [x] api tests with examples
- [x] vue components for admin
- [x] admin e2e tests
- [x] relations in preview mode
- [ ] rebase to master
- [x] split #163


JarJak:
Will be rebased to master after https://github.com/bolt/four/pull/317 gets merged.

package.json
```diff
@@ -65,13 +65,13 @@
...
     "babel-eslint": "^10.0.1",
-    "babel-jest": "^23.6.0",
+    "babel-jest": "^24.1.0",
     "eslint": "5.12.0",
     "eslint-config-prettier": "^3.6.0",
     "eslint-plugin-prettier": "^3.0.1",
     "eslint-plugin-standard": "^4.0.0",
     "eslint-plugin-vue": "^5.1.0",
-    "jest": "^23.6.0",
+    "jest": "^24.1.0",
```
bobdenotter: After we merge in this PR, #336 will give a conflict. I'll fix that one after this one is merged.

public/theme/skeleton/partials/_recordfooter.twig
```diff
@@ -34,13 +34,19 @@
         </p>
     {% endif %}
 
-    {# @todo See Github Issue https://github.com/bolt/four/issues/252 #}
-    {% set relatedrecords = related_content(record) %}
-    {% if relatedrecords is not empty %}
+    {% set related_cts = record|related_all %}
```
bobdenotter: Not sure if I like `related_cts`. Why not simply use `related_content` ?

src/Entity/Content.php
```diff
@@ -46,7 +50,7 @@ class Content implements \JsonSerializable
     /**
      * @var string
      *
-     * @ORM\Column(type="string", length=191, name="contenttype")
+     * @ORM\Column(type="string")
```
bobdenotter: Removing the `length=191` will break on MySQL 5.6. See https://github.com/symfony/symfony-docs/issues/6398

src/Entity/Content.php
```diff
@@ -63,15 +67,15 @@ class Content implements \JsonSerializable
     /**
      * @var string|null
      *
-     * @ORM\Column(type="string", length=191)
+     * @ORM\Column(type="string")
```
bobdenotter: Same here

src/Entity/Content.php
JarJak: Thanks, didn't know that! But this happens only on columns with `unique=true`, right? https://github.com/symfony/symfony-docs/pull/10055/files

bobdenotter: AFAIK it happens on all columns that are either `unique`, used as an index or as a Foreign Key.

public/theme/skeleton/partials/_recordfooter.twig
JarJak: Renamed to related_content_types

src/Entity/Content.php
JarJak: fixed

src/Entity/Content.php
JarJak: fixed

PR #335, opened 2019-02-24
Don't break on TightenCo collections, if Illuminate/support is present.

### LINKED ISSUE #333: Remove usages of collect()
Because they will conflict if someone would use Bolt and Laravel in same project. It creates errors like
>  Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Return value of Bolt\Configuration\Parser\BaseParser::parseConfigYaml() must be an instance of Tightenco\Collect\Support\Collection, instance of Illuminate\Support\Collection returned

bobdenotter:
Fixes #333 

I choose to do it like this, because `symplify/easy-coding-standard` requires `illimunate/support` already, and this is a simple way to fix the conflict between the two, who _both_ define a `collect()` function in the global namespace.


JarJak:
I would vote :-1: on this one. We can just lock simplify version.

JarJak:
See https://github.com/tightenco/collect/issues/104, they are working on some contracts there

JarJak:
After reading https://github.com/laravel/framework/pull/25421 I think the easiest way to solve this conflict would be to stop using Laravel's `collect()` be either replacing it with own `collect()` method or using `new Collection()`.

We main problem is that `collect()` is declared in the main namespace (which is one of the many reasons why I don't like Laravel code style).

JarJak:
Also, worth reading: https://mattallan.org/posts/dont-use-illuminate-support/

bobdenotter:
> We can just lock simplify version.

I don't like locking versions unless we have too. It can lead to confusing conflicts for users who don't understand why _their_ versions don't resolve. :-)

> `new Collection()` (I prefer the latter).

That's fine with me! I'll work on that.

Nevertheless, I'm not too happy about having _both_ `illuminate/support`, and `tightenco/collection`. How about we do use `illuminate/support`, and _not_ use `collect()` but `new Collection()`? That should prevent conflicts, right?

bobdenotter:
No more `\collect()`! :-D

JarJak:
Import changes should be reverted :)

src/Controller/Backend/ContentEditController.php
```diff
@@ -235,15 +236,15 @@ private function updateField(Content $content, string $fieldName, $value, ?strin
...
+        $taxonomy = new Collection(Json::findArray($taxonomy));
 
         // Remove old ones
         foreach ($content->getTaxonomies($key) as $current) {
             $content->removeTaxonomy($current);
         }
 
         // Then (re-) add selected ones
-        foreach ($taxonomy as $slug) {
+        foreach ($taxonomy->filter() as $slug) {
```
JarJak: ?

bobdenotter: Before: 

```
        $taxonomy = collect(Json::findArray($taxonomy))->filter();
…
        foreach ($taxonomy as $slug) {
```

after: 

```
        $taxonomy = new Collection(Json::findArray($taxonomy));
…
        foreach ($taxonomy->filter() as $slug) {
```

So the `->filter()` is in a different location, but the functionality is unaltered. All the `->filter()` does is skip over the 'empty' elements that might be in the collection.

JarJak: can you move it where it was?
I wonder why empty taxonomies are even passed to this method :thinking:

bobdenotter: I moved it back. :-)

Empty ones can be in there, because the Vue component _always_ passes in the value of the selection, so if the user clears an existing taxonomy, there's no selection and that means it gets one that's empty. /shrug

LarBearrr:
I know this issue has been on-going for sometime now, but I desperately need to figure out how to use this package with a Laravel 5.5 application, specfically through `okta/sdk`. Does anyone have suggestions on how to go about doing this? Okta says they have fixed the issue with their latest release, however I am still seeing composer errors. I have tried reaching out to their devs on numerous occasions but the response is slow:
```

  Problem 1
    - don't install tightenco/collect v5.8.31|don't install laravel/framework v5.5.46
    - don't install tightenco/collect v5.8.31|remove laravel/framework v5.5.46
    - don't install tightenco/collect v5.8.31|don't install laravel/framework v5.5.46
    - okta/sdk 1.2.0 requires tightenco/collect 5.8.31 -> satisfiable by tightenco/collect[v5.8.31].
    - Installation request for okta/sdk 1.2.0 -> satisfiable by okta/sdk[1.2.0].
    - Installation request for laravel/framework 5.5.46 -> satisfiable by laravel/framework[v5.5.46].
```

Any feedback is greatly appreciated, thank you :)

bobdenotter:
@LarBearrr Are you sure this is an issue with Bolt? I think this is a better fit in the okta/sdk repo.

LarBearrr:
🤦‍♂ Not sure how I ended up in this repo. My apologies!

PR #344, opened 2019-03-02
Be more lenient when calling non-existant object properties from within Twig

### LINKED ISSUE #343: Twig breaks on `is defined`
This should work: 

```twig
    {% if record.foo is defined %}
        It should be defined if we get here
        {{ dump(record.foo) }}
    {% endif %}
```

Instead I get this error: 

screenshot 2019-03-02 at 15 15 20

bobdenotter:
Fixes #343

### DIFF
--- a/src/Entity/Content.php
+++ b/src/Entity/Content.php
@@ -445,6 +445,14 @@ class Content implements \JsonSerializable
     /**
      * Generic getter for a record fields. Will return the field with $name.
      *
+     * If $name is not found, throw an exception if it's invoked from code, and
+     * return null if invoked from within a template. In templates we need to be
+     * more lenient, in order to do things like `{% if record.foo %}..{% endif %}
+     *
+     * Note: We can not rely on `{% if record.foo is defined %}`, because it
+     * always returns `true` for object properties.
+     * See: https://craftcms.stackexchange.com/questions/2116/twig-is-defined-always-returning-true
+     *
      * - {{ record.title }} => field named title
      * - {{ record|title }} => value of guessed title field
      * - {{ record.image }} => field named image
@@ -455,6 +463,14 @@ class Content implements \JsonSerializable
         try {
             $field = $this->getField($name);
         } catch (\InvalidArgumentException $e) {
+            $backtrace = new LaravelCollection(debug_backtrace());
+
+            if ($backtrace->contains('class', \Twig_Template::class)) {
+                // Invoked from within a Template render, so be lenient.
+                return null;
+            }
+
+            // Invoked from code, throw Exception
             throw new \RuntimeException(sprintf('Invalid field name or method call on %s: %s', $this->__toString(), $name));
         }
 
--- a/src/Entity/Content.php
+++ b/src/Entity/Content.php
@@ -463,7 +463,7 @@ class Content implements \JsonSerializable
         try {
             $field = $this->getField($name);
         } catch (\InvalidArgumentException $e) {
-            $backtrace = new LaravelCollection(debug_backtrace());
+            $backtrace = new LaravelCollection($e->getTrace());
 
             if ($backtrace->contains('class', \Twig_Template::class)) {
                 // Invoked from within a Template render, so be lenient.
--- a/public/theme/skeleton/record.twig
+++ b/public/theme/skeleton/record.twig
@@ -2,7 +2,7 @@
 
 {% block main %}
 
-    {% if record.hasField('title') %}
+    {% if record.title %}
         <h1 class="title">{{ record.title }}</h1>
     {% else %}
         <h1 class="heading">{{ record|title }}</h1>
@@ -14,6 +14,7 @@
         <a href="{{ record.image }}">
             <img src="{{ thumbnail(record.image, 400, 260) }}">
         </a>
+
         <hr>
     {% endif %}
 


src/Entity/Content.php
```diff
@@ -445,16 +445,32 @@ public function removeTaxonomy(Taxonomy $taxonomy): self
     /**
      * Generic getter for a record fields. Will return the field with $name.
      *
+     * If $name is not found, throw an exception if it's invoked from code, and
+     * return null if invoked from within a template. In templates we need to be
+     * more lenient, in order to do things like `{% if record.foo %}..{% endif %}
+     *
+     * Note: We can not rely on `{% if record.foo is defined %}`, because it
+     * always returns `true` for object properties.
+     * See: https://craftcms.stackexchange.com/questions/2116/twig-is-defined-always-returning-true
```
JarJak: Have you tried using http://php.net/manual/en/language.oop5.overloading.php#object.isset instead?

JarJak:
Can you add `is defined` check in sample template and a very simple e2e test to it?

src/Entity/Content.php
bobdenotter: I hadn't before, but I did try now.. When doing something like this: 

```
    public function __isset($name)
    {
        try {
            $this->getField($name);
            return true;
        } catch (\InvalidArgumentException $e) {
            return false;
        }
    }
```

The behaviour remains unchanged. If I break it down, it seems that the following happens with this twig snippet: 

```twig
    {% if record.foo is defined %}
        It "record.foo" should be defined, we get here
        {#{{ dump(record.foo) }}#}
    {% endif %}
```

When `record.foo` is parsed in the `{% if %}`, it correctly calls my overridden `__isset()`. My overridden isset _correctly_ returns `false` to twig. Then the `is defined` test is parsed, and it still *incorreclty* returns `true`, causing the part within the `{% if %}` to be output. :-/

bobdenotter:
> Can you add `is defined` check in sample template and a very simple e2e test to it?

Yes, when we have a working solution (or agreed on a workaround), i'll happily add a e2e test for it.

src/Entity/Content.php
bobdenotter: Upon further investigation, overriding `__isset()` is no good, for two reasons: 

 1) It's ignored in Twig's `defined` test. Even if I do this, the results get ignored:

```php
    public function __isset(string $name)
    {
        return null;
    }
```

2) The problem is in Twig. Overriding it, will have side effects for when calling it from PHP code, making it more complex and potentially buggy. 

So, let's stick with the general direction of the workaround I've made above. I've added an e2e test to make sure it works as expected. 

As an additional benefit, it'll make cleaner, more expressive templates for implementors, because we can do `{% if record.foo %}…`, even in strict mode. If you really need to test if the _field_ exists, you can do `{% if record.foo is null %}`.

JarJak: How about changing __call to __get, will that work with __isset ?

JarJak: > If you really need to test if the field exists, you can do {% if record.foo is null %}.

Looks like a nightmare if you want to check if field has the value of null...

JarJak: If you just want to check if field has been filled, you can use `record.foo|default`

JarJak:
I really want to avoid mixing up M and V layers... but if nothing else works, we need to do it that way.

bobdenotter:
Tests are failing because of 

```
PHP Fatal error:  Uncaught Symfony\Component\DependencyInjection\Exception\RuntimeException: Cannot autowire service "Symplify\TokenRunner\DocBlock\DescriptionAnalyzer": argument "$typeNodeAnalyzer" of method "__construct()" references class "Symplify\BetterPhpDocParser\PhpDocParser\TypeNodeAnalyzer" but no such service exists. in /Users/bob/Sites/bolt-four/vendor/symfony/dependency-injection/Compiler/DefinitionErrorExceptionPass.php:54
Stack trace:
```

This looks like an issue with EasyCodingStandard at first glance, because on current master it also breaks. :-)

> I really want to avoid mixing up M and V layers... but if nothing else works, we need to do it that way.

I agree, but from what I can tell, our friends in Drupal and Craft bumped into the same problem, and no clean solution there either.

src/Entity/Content.php
```diff
@@ -445,16 +445,32 @@ public function removeTaxonomy(Taxonomy $taxonomy): self
...
      * - {{ record.image }} => field named image
      * - {{ record|image }} => value of guessed image field
      */
-    public function __call(string $name, array $arguments = []): Field
+    public function __call(string $name, array $arguments = []): ?Field
     {
         try {
             return $this->getField($name);
         } catch (\InvalidArgumentException $e) {
+            $backtrace = new LaravelCollection(debug_backtrace());
```
JarJak: $e->getTrace()

bobdenotter: Fixed

JarJak:
https://twig.symfony.com/doc/2.x/recipes.html#using-dynamic-object-properties

I should work with __get and __isset

bobdenotter:
> I should work with __get and __isset

That is basically what we're doing here: Using these we can influence how `{% if record.foo %}` and `{{ record.foo }}` behave. This PR makes _that_ work as expected in Twig. 

The underlying issue here is in the `is defined` test. Even if `record.foo` returns `null`, the `is defined` returns `true`, making the test give the wrong behaviour. Pseudocode: 

`if ((record.foo) is defined)` -> `if (null is defined)` -> `if (true)` -> incorrectly evaluates to `true`

JarJak:
But we use __call, not __get.

How it works in Bolt 3?

src/Entity/Content.php
bobdenotter: Rebased this PR to master, and it's quite a bit simpler as a result. 

> How about changing __call to __get, will that work with __isset ? 

Nope, it doesn't. :-)

> Looks like a nightmare if you want to check if field has the value of null...

Thinking some more on this.. the correct syntax in Twig would be: `{% if record.foo is null %}`. But, the common fields _never_ return `null`. Only '``' or `false`. 

I can't really think of a usecase where you'd need to distinguish in Twig between a field that doesn't exist, _or_ a field that does exist but has value `null`. In these cases, you're writing the template so you most likely know which case you'll be dealing with.

JarJak: If you really need to test if the field exists, you can do `{% if record.hasField('foo') %}`.

^ that's the correct way :)

And I agree that usecases for it would be rather rare.

Issue #351, opened 2019-03-07
Default 'escaping' of field output in Twig.


bobdenotter:
In Bolt 3, if you do `{{ record.fieldname }}`, it'll output HTML. At the same time, it's not *explicitly defined* how it should behave. One of the causes of this is how the fields are stored internally. 

We should make sure that Bolt 4 does this sensibly, consistent and robust. In my opinion, it should work like the following. Let's say we have a title field, with a `<i>` tag in it.

![schermafbeelding 2019-03-07 om 13 59 32](https://user-images.githubusercontent.com/1833361/53958406-4458cb80-40e1-11e9-9c73-87f0dc955248.png)

In Twig we should be able to do: 

```twig
<hr>
    <p>{{ record.title }}</p>
    <p>{{ record.title|raw }}</p>
    <p>{{ record.title|e('html') }}</p>
    <p>{{ dump(record.title) }}</p>
    <p>{{ dump(record.title.getValue()) }}</p>
<hr>
```

And it should output: 

![schermafbeelding 2019-03-07 om 14 07 37](https://user-images.githubusercontent.com/1833361/53958875-66068280-40e2-11e9-9565-c9b91e9a8a40.png)

This way, the _implementor_ can use `{{ record.foo }}` and get the HTML output they expect 90% of the time, and at the same time, you still have access to the underlying data structure to do with it what you need.

stryxx:
@bobdenotter What does happen there when you add to field <script src="../link/to/bitcoin.miner.js"></script>? :)

JarJak:
IMO, since Bolt is a WYSIWYG editor, if an admin puts `<i>Something</i>` in a text or textarea field, he should see `<i>Something</i>` on website. That's because this type of field was not made to store html data but raw text.
On the other hand, if the field has type of html, as admin expects html output on website, so he should see *Something*.

BUT we could add an option to make simple text field behave as html one (e.g. a field to paste some js stuff). Example:
```
ga_script:
    type: text
    safe: true # default: false
```

bobdenotter:
@stryxx Bolt 3 had a whitelist-based filtering for HTML-like fields, where input would be sanitized before going into the database. In Bolt 3's config.yml, you had: 

```yaml
# Define the HTML tags and attributes that are allowed in 'cleaned' HTML. This
# is used for sanitizing HTML, to make sure there are no undesirable elements
# left in the content that is shown to users. For example, tags like `<script>`
# or `onclick`-attributes.
[...]

PR #353, opened 2019-03-10
Default escaping in fields

bobdenotter:
Make field output like `{{ record.title }}` behave like we'd expect: 

1. "semantically pretty" as `{{ record.title }}` outputs HTML for fields where it makes sense ('html', 'markdown', 'textarea')
2. Add `allow_markup` flag to overrule the default behaviour in `contenttypes.yml`
3. add `.raw('foo')` and `.escape('foo')` methods to `Content` for fine-grained control in output.

Additionally, the PR adds a new way to preset records to be added, which is convenient for our automated tests.

See discussion in #351

### DIFF
--- a/config/bolt/contenttypes.yaml
+++ b/config/bolt/contenttypes.yaml
@@ -336,12 +336,10 @@ tests:
             uses: title
         text_markup:
             type: text
-            allow:
-                markup: true
+            allow_markup: true
         text_plain:
             type: text
-            allow:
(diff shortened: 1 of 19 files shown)


bobdenotter:
Let's retire this one, and go with the one in #354 instead.

JarJak:
So this can be closed, right?

bobdenotter:
yes :-)

PR #354, opened 2019-03-10
Default escaping in Fields, attempt 2

bobdenotter:
Make field output like {{ record.title }} behave like we'd expect:

 - "semantically pretty" as {{ record.title }} outputs HTML for fields where it makes sense ('html', 'markdown', 'textarea')
 - Add `allow_markup` flag to overrule the default behaviour in contenttypes.yml,
use standard `|raw` and `|e` methods to Content for fine-grained control in output.
 - Additionally, the PR adds a new way to preset records to be added, which is convenient for our automated tests.
 - No hack needed to change the default or current `autoescape: false`.

See discussion in #351

This PR is less hackish than #353, but does the same thing basically.

### DIFF
--- a/config/bolt/contenttypes.yaml
+++ b/config/bolt/contenttypes.yaml
@@ -336,12 +336,10 @@ tests:
             uses: title
         text_markup:
             type: text
-            allow:
-                markup: true
+            allow_markup: true
         text_plain:
             type: text
-            allow:
-                markup: false
+            allow_markup: false
         selectfield:
             type: select
             values:
--- a/config/bolt/contenttypes.yaml
+++ b/config/bolt/contenttypes.yaml
@@ -336,10 +336,10 @@ tests:
             uses: title
         text_markup:
             type: text
-            allow_markup: true
+            allow_html: true
         text_plain:
             type: text
-            allow_markup: false
+            allow_html: false
         html_field:
             type: html
         markdown_field:
--- a/config/bolt/contenttypes.yaml
+++ b/config/bolt/contenttypes.yaml
@@ -322,9 +322,10 @@ blocks:
     icon_one: "fa:cube"
 
 
-dummies:
-    name: Dummies
-    singular_name: Dummy
+# This contenttype is here to use for (automated) tests.
+tests:
+    name: Tests
+    singular_name: Test
     fields:
         title:
             type: text
@@ -333,6 +334,14 @@ dummies:
         slug:
             type: slug
             uses: title
+        text_markup:
+            type: text
+            allow:
+                markup: true
+        text_plain:
+            type: text
+            allow:
+                markup: false
         selectfield:
             type: select
             values:
@@ -345,6 +354,10 @@ dummies:
             multiple: true
             postfix: "Select your favourite turtle(s)."
     taxonomy: [ groups, categories, tags, foobars ]
+    record_template: test.twig
+    listing_template: listing.twig
+    listing_records: 10
+
 
 # Possible field types:
 #
--- a/config/bolt/contenttypes.yaml
+++ b/config/bolt/contenttypes.yaml
@@ -340,6 +340,12 @@ tests:
         text_plain:
             type: text
             allow_markup: false
+        html_field:
+            type: html
+        markdown_field:
+            type: markdown
+        textarea_field:
+            type: textarea
         selectfield:
             type: select
             values:
--- a/src/Configuration/Parser/ContentTypesParser.php
+++ b/src/Configuration/Parser/ContentTypesParser.php
@@ -209,6 +209,10 @@ class ContentTypesParser extends BaseParser
                 $hasGroups = true;
             }
 
+            if (! isset($field['allow_markup'])) {
+                $field['allow_markup'] = in_array($field['type'], ['html', 'markdown', 'textarea'], true);
+            }
+
             // Make sure we have these keys and every field has a group set.
             $field = array_replace(
                 [
--- a/src/Configuration/Parser/ContentTypesParser.php
+++ b/src/Configuration/Parser/ContentTypesParser.php
@@ -209,8 +209,8 @@ class ContentTypesParser extends BaseParser
                 $hasGroups = true;
             }
 
-            if (! isset($field['allow_markup'])) {
-                $field['allow_markup'] = in_array($field['type'], ['html', 'markdown', 'textarea'], true);
+            if (isset($field['allow_markup']) === false) {
+                $field['allow_markup'] = in_array($field['type'], ['html', 'markdown'], true);
             }
 
             // Make sure we have these keys and every field has a group set.
--- a/src/Configuration/Parser/ContentTypesParser.php
+++ b/src/Configuration/Parser/ContentTypesParser.php
@@ -209,8 +209,8 @@ class ContentTypesParser extends BaseParser
                 $hasGroups = true;
             }
 
-            if (isset($field['allow_markup']) === false) {
-                $field['allow_markup'] = in_array($field['type'], ['html', 'markdown'], true);
+            if (isset($field['allow_html']) === false) {
+                $field['allow_html'] = in_array($field['type'], ['html', 'markdown'], true);
             }
 
             // Make sure we have these keys and every field has a group set.
--- a/src/DataFixtures/ContentFixtures.php
+++ b/src/DataFixtures/ContentFixtures.php
@@ -84,7 +84,7 @@ class ContentFixtures extends Fixture implements DependentFixtureInterface
                     $field->setName($name);
 
                     if (isset($preset[$name])) {
-                        $field->setValue((array) $preset[$name]);
+                        $field->setValue($preset[$name]);
                     } else {
                         $field->setValue($this->getValuesforFieldType($name, $fieldType));
                     }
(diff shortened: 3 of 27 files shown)


JarJak:
>"semantically pretty" as {{ record.title }} outputs HTML for fields where it makes sense ('html', 'markdown', 'textarea')

I disagree that textarea should output raw html by default.

src/DataFixtures/ContentFixtures.php
```diff
@@ -129,4 +141,36 @@ private function getValuesforFieldType($name, $field)
...
+            'text_markup' => 'Text with <em>markup allowed</em>.',
+            'text_plain' => 'Text with <strong>no</strong> markup allowed.',
+        ];
+
+        return $records;
+    }
+
+    private function getPreset(string $slug): array
+    {
+        if (isset($this->presetRecords[$slug]) && ! empty($this->presetRecords[$slug])) {
```
JarJak: return $this->presetRecords->has($slug) ? $this->presetRecords->get($slug)->toArray() : null;

src/DataFixtures/ContentFixtures.php
```diff
@@ -129,4 +141,36 @@ private function getValuesforFieldType($name, $field)
...
+        $records['tests'][] = [
+            'selectfield' => 'bar',
+            'multiselect' => 'Michelangelo',
+            'slug' => 'title-of-the-test',
+            'title' => 'Title of the test',
+            'text_markup' => 'Text with <em>markup allowed</em>.',
+            'text_plain' => 'Text with <strong>no</strong> markup allowed.',
+        ];
+
+        return $records;
```
JarJak: return new Collection($records);

src/DataFixtures/ContentFixtures.php
```diff
@@ -71,11 +76,18 @@ private function loadContent(ObjectManager $manager): void
...
+                $preset = $this->getPreset($contentType['slug']);
+
                 $sortorder = 1;
                 foreach ($contentType['fields'] as $name => $fieldType) {
                     $field = Field::factory($fieldType, $name);
                     $field->setName($name);
-                    $field->setValue($this->getValuesforFieldType($name, $fieldType));
+
+                    if (isset($preset[$name])) {
+                        $field->setValue((array) $preset[$name]);
```
JarJak: casting to array should be done inside setValue()

src/Entity/Content.php
```diff
@@ -450,12 +450,18 @@ public function removeTaxonomy(Taxonomy $taxonomy): self
...
+    public function __call(string $name, array $arguments = [])
     {
         try {
-            return $this->getField($name);
+            $field = $this->getField($name);
         } catch (\InvalidArgumentException $e) {
             throw new \RuntimeException(sprintf('Invalid field name or method call on %s: %s', $this->__toString(), $name));
         }
+
+        if ($field->getDefinition()->get('allow_markup')) {
```
JarJak: move to Field class (probably to getFlatValue? or new getTwigValue()?)

src/Entity/ContentExtrasTrait.php
```diff
@@ -71,4 +71,11 @@ public function jsonSerialize(): array
             'depublishedAt' => $content->getDepublishedAt(),
         ];
     }
+
+    public function var_export(): string
```
JarJak: remove

templates/_partials/fields/html.html.twig
```diff
@@ -2,7 +2,7 @@
 
 {% block field %}
     <editor-html
-        :value="'{{ value|json_encode() }}'"
+        :value="{{ value|json_encode() }}"
```
JarJak: remove brackets

templates/_partials/fields/markdown.html.twig
```diff
@@ -2,7 +2,7 @@
 
 {% block field %}
     <editor-markdown
-        :value="'{{value|json_encode()}}'"
+        :value='{{ value|json_encode() }}'
```
JarJak: remove brackets

src/DataFixtures/ContentFixtures.php
bobdenotter: Think so? Then we'd need to remove the typehint from `setValue`:

```php
    public function setValue(array $value): self
    {
        $this->value = $value;

        return $this;
    }
```

src/Configuration/Parser/ContentTypesParser.php
```diff
@@ -209,6 +209,10 @@ protected function parseFieldsAndGroups(array $fields): array
                 $hasGroups = true;
             }
 
+            if (! isset($field['allow_markup'])) {
```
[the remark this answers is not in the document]
bobdenotter: Why? :-)

Not saying we shouldn't, but if we do, we should probably update it in a bunch more places.

src/DataFixtures/ContentFixtures.php
JarJak: Yes, that makes sense. Only Field class can we aware of that value is stored as array.

src/Configuration/Parser/ContentTypesParser.php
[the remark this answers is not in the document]
JarJak: Why? - better readability.
We can update it in places step by step.

src/DataFixtures/ContentFixtures.php
bobdenotter: Ok, will change it, and cast it to array in the method.

config/bolt/contenttypes.yaml
```diff
@@ -333,6 +334,12 @@ dummies:
         slug:
             type: slug
             uses: title
+        text_markup:
+            type: text
+            allow_markup: true
+        text_plain:
+            type: text
+            allow_markup: false
```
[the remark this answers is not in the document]
bobdenotter: Done!

src/Entity/Field/MarkdownField.php
```diff
@@ -5,11 +5,18 @@
...
 /**
  * @ORM\Entity
  */
 class MarkdownField extends Field implements Excerptable
 {
+    public function __toString(): string
+    {
+        $markdown = new Markdown();
+
+        return $markdown->toHtml(implode(', ', $this->getValue()));
```
JarJak: how can markdown have more that 1 item in value array?

src/Entity/Content.php
```diff
@@ -450,12 +450,18 @@ public function removeTaxonomy(Taxonomy $taxonomy): self
...
+            $field = $this->getField($name);
         } catch (\InvalidArgumentException $e) {
             throw new \RuntimeException(sprintf('Invalid field name or method call on %s: %s', $this->__toString(), $name));
         }
+
+        if ($field->getDefinition()->get('allow_html')) {
+            $field = new \Twig_Markup($field, 'UTF-8');
+        }
+
+        return $field;
```
JarJak: can we assume that magic __call will be used only in Twig and return $field->getTwigValue() here?
Then we can move checking for allow_html there.

src/Entity/Field/MarkdownField.php
bobdenotter: Future-proofing. The value of _each_ field is basically an asociative array. Some fields (like 'embed' or 'image' have more items. Apart from simply treating them all the same this way, it allows us later to perhaps add the compiled HTML for performance, or a 'flavor' value for normal or GFM markdown.

src/Entity/Content.php
bobdenotter: Yes, I think so.. Will look into that!

src/Entity/Field/MarkdownField.php
bobdenotter: Thinking about this some more.. The `implode` is ugly.. I'll change it to `reset($value)`.

src/DataFixtures/ContentFixtures.php
bobdenotter: Nah, in this instance (arrays in arrays), using Collection just makes it more cumbersome and less legible. Maybe (if/when) we refactor to using Bolt\Collection we might, but for now i'd like to leave it as is.

src/DataFixtures/ContentFixtures.php
bobdenotter: I don't think that does the same as the current code.

src/Entity/ContentExtrasTrait.php
bobdenotter: Done

src/Entity/Field.php
```diff
@@ -198,9 +198,20 @@ public function getFlattenedValue()
         return $value;
     }
 
-    public function setValue(array $value): self
+    public function getTwigValue()
```
JarJak: add phpdoc with return

src/Entity/Field.php
```diff
@@ -198,9 +198,20 @@ public function getFlattenedValue()
...
     {
-        $this->value = $value;
+        $value = $this->__toString();
+
+        if ($this->getDefinition()->get('allow_html')) {
+            $value = new \Twig_Markup($value, 'UTF-8');
```
JarJak: use namespaced \Twig\Markup

src/Entity/Field.php
bobdenotter: Done

src/Entity/Field.php
bobdenotter: And done!

bobdenotter:
Thanks for review, @JarJak 👍

PR #356, opened 2019-03-13
Fix `Twig\Template` -> `Twig\TemplateWrapper`, because of Twig 2.7

bobdenotter:


JarJak:
Actually, do we need this extension?
Doesn't symfony demo need a fix too?

bobdenotter:
> Actually, do we need this extension?

We don't really use it at the moment, but I would like to. 

> Doesn't symfony demo need a fix too?

Yes. :-) https://github.com/symfony/demo/pull/955

PR #355, opened 2019-03-13
Add paging in frontend / `{{ pager() }}`

bobdenotter:
Fixes #254

PR #358, opened 2019-03-13
[hotfix] Return correct value for images (array with meta)

bobdenotter:


public/theme/skeleton/test.twig
```diff
@@ -62,13 +62,28 @@
...
+    <h2>Testing output for images</h2>
+
+    {% if record.image %}
+        We have a field named `image`: {{ dump(record.getField('image').getValue()) }}
+        {{ dump(record.image) }}
+        <img src="{{ thumbnail(record.image, 400, 260) }}">
+        <hr>
+    {% endif %}
+
+    {% if record.image %}
```
JarJak: picture

src/Entity/Field/ImageField.php
```diff
@@ -33,6 +33,11 @@ public function getValue(): array
         return $value;
     }
 
+    public function getTwigValue()
+    {
+        return $this->getValue();
```
JarJak: it should defaults to getFlattenedValue() somewhere inside Field class

bobdenotter: Good point. Have done so, and removed the overridden method from `Field\ImageField`.

src/Entity/Field.php
```diff
@@ -200,11 +200,11 @@ public function getFlattenedValue()
...
     /**
-     * @return string|Markup
+     * @return string|array|Markup
      */
     public function getTwigValue()
     {
-        $value = $this->__toString();
+        $value = $this->getFlattenedValue();
 
         if ($this->getDefinition()->get('allow_html')) {
```
JarJak: `if (is_string($value) && $this->getDefinition()->get('allow_html')) {`

JarJak: or Markup allows arrays too?

bobdenotter: No, `Markup` is always a `string` with the contents, plus a `string` with the encoding.

JarJak: Ok but using allow_html on i.e. image field doesn't make sense and it will break here if it happens, no?

bobdenotter: Ah, yes, you are correct. Fixed.

bobdenotter:
Finally green lights. This one and #355 took waaay more effort than I expected. 😓

src/Entity/Field/MarkdownField.php
```diff
@@ -20,4 +21,18 @@ public function __toString(): string
 
         return $markdown->toHtml(reset($value));
     }
+
+    /**
+     * @return string|array|Markup
+     */
+    public function getTwigValue()
+    {
+        $value = (string) $this;
```
JarJak: How this one differs from the base Field class?

bobdenotter: By calling `(string) $this` it gets parsed as Markdown. Otherwise it would return the "stored" value, unparsed.

JarJak: same should happen inside getFlattenedValue as it is used in API response
(maybe rename to getParsedValue?)

bobdenotter: Hmm, makes sense.. I'll work on this some more.

bobdenotter:
Made the requested changes, and I think it's a bit cleaner now! 👍

bobdenotter:
🎊 🎉

bobdenotter:
Gihub says "1 check was pending", but it was finished with green lights: https://travis-ci.org/bolt/four/builds/508871291

PR #360, opened 2019-03-14
Fix Kakunin

JarJak:


JarJak:
@bobdenotter this time it breaks in another place...

bobdenotter:
The issue is in Kakunin still, i'm afraid.. It breaks here: 

```
        When I visit the "edit_record" page with parameters:
            | id          | 1  |
            | edit_locale | nl |
```

The URL in the browser becomes `http://127.0.0.1:8088/bolt/edit/1?0=edit_locale&0=nl`. So `?0=edit_locale&0=nl` instead of `?edit_locale=nl`

Screenshot 2019-03-14 at 19 16 39

bobdenotter:
Added a workaround, with an `@todo`. See also https://github.com/TheSoftwareHouse/Kakunin/pull/157

PR #364, opened 2019-03-23
Make search work in Frontend

### LINKED ISSUE #306: Make "Search" work
Splitted from #42

bobdenotter:
Fixes #306 

todo:

 - [x] Update search template
 - [x] Make search controller work 
 - [x] Implement naive search
 - [x] Add tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,15 +71,6 @@ e2e-wip:
 e2e-install:
 	cd tests/e2e && npm install
 	node ./tests/e2e/node_modules/protractor/bin/webdriver-manager update --gecko=false
-	cd tests/e2e/step_definitions
-	ln -s ../node_modules/kakunin/dist/step_definitions/elements.js kakunin-elements.js
-	ln -s ../node_modules/kakunin/dist/step_definitions/debug.js kakunin-debug.js
-	ln -s ../node_modules/kakunin/dist/step_definitions/file.js kakunin-file.js
-	ln -s ../node_modules/kakunin/dist/step_definitions/form.js kakunin-form.js
-	ln -s ../node_modules/kakunin/dist/step_definitions/email.js kakunin-email.js
-	ln -s ../node_modules/kakunin/dist/step_definitions/generators.js kakunin-generators.js
-	ln -s ../node_modules/kakunin/dist/step_definitions/navigation.js kakunin-navigation.js
-	cd ../../..
 
 db-create:
 	bin/console doctrine:database:create
--- a/Makefile
+++ b/Makefile
@@ -69,10 +69,9 @@ e2e-wip:
 	cd tests/e2e && npm run kakunin -- --tags @wip && cd ../..
 
 e2e-install:
-	cd tests/e2e
-	npm install
-	node ./node_modules/protractor/bin/webdriver-manager update --gecko=false
-	cd step_definitions
+	cd tests/e2e && npm install
+	node ./tests/e2e/node_modules/protractor/bin/webdriver-manager update --gecko=false
+	cd tests/e2e/step_definitions
 	ln -s ../node_modules/kakunin/dist/step_definitions/elements.js kakunin-elements.js
 	ln -s ../node_modules/kakunin/dist/step_definitions/debug.js kakunin-debug.js
 	ln -s ../node_modules/kakunin/dist/step_definitions/file.js kakunin-file.js
--- a/public/theme/skeleton/search.twig
+++ b/public/theme/skeleton/search.twig
@@ -22,7 +22,7 @@
             {{ block.content }}
         {% endif %}
 
-        <form method="get" action="{{ path('search') }}" id="searchform-inline" enctype="text/plain">
+        <form method="get" action="{{ path('search') }}" id="searchform-main" enctype="text/plain">
             <input type="search" value="{{ searchTerm|default() }}" placeholder="{{ __('general.phrase.search-ellipsis') }}" name="searchTerm" >
             <button type="submit">{{ __('general.phrase.search') }}</button>
         </form>
--- a/public/theme/skeleton/search.twig
+++ b/public/theme/skeleton/search.twig
@@ -48,7 +48,7 @@
 
         {% else %}
 
-            {% if search is not empty %}
+            {% if searchTerm is not empty %}
                 <p class="search-results-description">
                     {{ __('general.phrase.no-search-results-for', { '%search%': searchTerm|escape }) }}
                 </p>
--- a/public/theme/skeleton/search.twig
+++ b/public/theme/skeleton/search.twig
@@ -7,8 +7,8 @@
            records matching the current query. If there are no results, the
            code in the 'else' part of the for-loop is used. #}
         <h1 class="search-results">
-            {% if search is not empty %}
-                {{ __('general.phrase.search-results-for', { '%search%': search }) }}
+            {% if searchTerm is not empty %}
+                {{ __('general.phrase.search-results-for', { '%search%': searchTerm }) }}
             {% else %}
                 {{ __('general.phrase.search') }}
             {% endif %}
@@ -23,7 +23,7 @@
         {% endif %}
 
         <form method="get" action="{{ path('search') }}" id="searchform-inline" enctype="text/plain">
-            <input type="search" value="{{ search|default() }}" placeholder="{{ __('general.phrase.search-ellipsis') }}" name="search" >
+            <input type="search" value="{{ searchTerm|default() }}" placeholder="{{ __('general.phrase.search-ellipsis') }}" name="searchTerm" >
             <button type="submit">{{ __('general.phrase.search') }}</button>
         </form>
 
@@ -50,7 +50,7 @@
 
             {% if search is not empty %}
                 <p class="search-results-description">
-                    {{ __('general.phrase.no-search-results-for', { '%search%': search|escape }) }}
+                    {{ __('general.phrase.no-search-results-for', { '%search%': searchTerm|escape }) }}
                 </p>
             {% else %}
                 <p class="search-results-description">
--- a/public/theme/skeleton/search.twig
+++ b/public/theme/skeleton/search.twig
@@ -18,7 +18,7 @@
         {% setcontent block = "block/search-teaser" %}
 
         {# check if we have 'content'. If so, we know we have have a teaser to display. #}
-        {% if block is defined and block.hasField('content') %}
+        {% if block and block.content %}
             {{ block.content }}
         {% endif %}
 
@@ -57,6 +57,6 @@
         {% endfor %}
 
         {# If there are more records than will fit on one page, the pager is shown. #}
-        {{ pager(template = 'partials/_sub_pager.twig') }}
+        {{ pager(records, template = 'partials/_sub_pager.twig') }}
 
 {% endblock main %}
--- a/public/theme/skeleton/search.twig
+++ b/public/theme/skeleton/search.twig
@@ -48,9 +48,15 @@
 
         {% else %}
 
-                <p>
+            {% if search is not empty %}
+                <p class="search-results-description">
                     {{ __('general.phrase.no-search-results-for', { '%search%': search|escape }) }}
                 </p>
+            {% else %}
+                <p class="search-results-description">
+                    {{ __('general.phrase.no-search-term-provided') }}
+                </p>
+            {% endif %}
 
         {% endfor %}
 
--- a/public/theme/skeleton/search.twig
+++ b/public/theme/skeleton/search.twig
@@ -6,9 +6,9 @@
            we display an appropriate title. The 'records' array contains all of the
            records matching the current query. If there are no results, the
            code in the 'else' part of the for-loop is used. #}
-        <h1>
+        <h1 class="search-results">
             {% if search is not empty %}
-                {{ __('general.phrase.search-results-for-variable', { '%search%': search }) }}
+                {{ __('general.phrase.search-results-for', { '%search%': search }) }}
             {% else %}
                 {{ __('general.phrase.search') }}
             {% endif %}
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -69,10 +69,9 @@ e2e-wip:
 	cd tests/e2e && npm run kakunin -- --tags @wip && cd ../..
 
 e2e-install:
-	cd tests/e2e
-	npm install
-	node ./node_modules/protractor/bin/webdriver-manager update --gecko=false
-	cd step_definitions
+	cd tests/e2e && npm install
+	node ./tests/e2e/node_modules/protractor/bin/webdriver-manager update --gecko=false
+	cd tests/e2e/step_definitions
```
JarJak: what was the problem here? do symlinks below work?

public/theme/skeleton/search.twig
```diff
@@ -18,7 +18,7 @@
         {% setcontent block = "block/search-teaser" %}
```
JarJak: rename block to searchTeaser

public/theme/skeleton/search.twig
```diff
@@ -18,7 +18,7 @@
...
 
         {# check if we have 'content'. If so, we know we have have a teaser to display. #}
-        {% if block is defined and block.hasField('content') %}
+        {% if block and block.content %}
```
JarJak: rename content to something more meaningful

public/theme/skeleton/search.twig
```diff
@@ -47,16 +47,14 @@
...
-                    {{ __('general.phrase.no-result-for-search-term', { '%SEARCHTERM%': search|escape }) }}
+                    {{ __('general.phrase.no-search-results-for', { '%search%': search|escape }) }}
                 </p>
 
-            </article>
         {% endfor %}
 
         {# If there are more records than will fit on one page, the pager is shown. #}
-        {{ pager(template = 'partials/_sub_pager.twig') }}
+        {{ pager(records, template = 'partials/_sub_pager.twig') }}
```
JarJak: this template is default so can be omitted here

src/Repository/ContentRepository.php
```diff
@@ -67,6 +70,41 @@ public function findLatest(?ContentType $contentType = null, int $amount = 6): ?
         return $qb->getQuery()->getResult();
     }
 
+    public function findNaiveSearch(int $page = 1, string $search = '', int $amountPerPage = 6, bool $onlyPublished = true)
+    {
+        if (empty($search)) {
+            return null;
```
JarJak: move input validation to controller

src/Repository/ContentRepository.php
```diff
@@ -67,6 +70,41 @@ public function findLatest(?ContentType $contentType = null, int $amount = 6): ?
         return $qb->getQuery()->getResult();
     }
 
+    public function findNaiveSearch(int $page = 1, string $search = '', int $amountPerPage = 6, bool $onlyPublished = true)
```
JarJak: rename: searchNaive

src/Repository/ContentRepository.php
```diff
@@ -67,6 +70,41 @@ public function findLatest(?ContentType $contentType = null, int $amount = 6): ?
...
+        $qb = $this->getQueryBuilder()
+            ->select('partial content.{id}');
+
+        $qb->addSelect('f')
+            ->leftJoin('content.fields', 'f')
+            ->andWhere($qb->expr()->like('f.value', ':search'))
+            ->setParameter('search', '%' . $search . '%');
+
+        // These are the ID's of content we need.
+        $ids = new Collection($qb->getQuery()->getArrayResult());
```
JarJak: $ids = array_column($qb->getQuery()->getScalarResult(), "id")

src/Twig/RecordExtension.php
```diff
@@ -75,8 +75,12 @@ public function dummy_with_env(Environment $env, $input = null)
         return $input;
     }
 
-    public function pager(Environment $env, Pagerfanta $records, string $template = '_sub_pager.twig', string $class = 'pagination', string $theme = 'default', int $surround = 3)
+    public function pager(Environment $env, ?Pagerfanta $records, string $template = '_sub_pager.twig', string $class = 'pagination', string $theme = 'default', int $surround = 3)
```
JarJak: rename $env to $twig (everywhere in this class)

JarJak: how can $records be null instead of empty array / empty pagerfanta object?

src/Twig/RecordExtension.php
```diff
@@ -75,8 +75,12 @@ public function dummy_with_env(Environment $env, $input = null)
...
     {
+        if (! $records) {
```
JarJak: if (count($records) === 0)

tests/e2e/features/display_search.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Display search functionality
+    @wip
+    Scenario: As a user I want to display search results
+        When I visit the "homepage" page
+
+        Then I scroll to the "search_input" element
+        When I fill the "search_input" field with "consequatur"
+        And I click the "search_button" element
+
```
JarJak: add "search" page

tests/e2e/pages/homepage.js
```diff
@@ -8,6 +8,13 @@ class HomepagePage extends BasePage {
 
     this.title = $('header a');
     this.recent_pages_list = $('aside ul');
+
+    // For testing "search"
```
JarJak: move to "search" page

Makefile
bobdenotter: I'm not sure.. it was this: 

```
cd tests/e2e
npm install
```

Which resulted in `make` running `npm install` in the project root, instead of in the folder it `cd`'ed to in the previous command.

bobdenotter: The symlinks do work, yes.

public/theme/skeleton/search.twig
bobdenotter: The "Blocks" ContentType has a `content` field. It's generic on purpose :-)

public/theme/skeleton/search.twig
bobdenotter: In the past i've experienced that it's helpful for people who want to tweak it, to have this explicit. We've had a _lot_ of people asking "How do I change the pager template?"

src/Twig/RecordExtension.php
bobdenotter: I'll make a separate PR for that. See #368

bobdenotter: People invoking `{{ pager(records) }}` when records is null. For instance on an empty search page.

tests/e2e/features/display_search.feature
bobdenotter: What do you mean?

src/Twig/RecordExtension.php
JarJak: on empty search page, `records` should be an empty iterable, not null

tests/e2e/features/display_search.feature
JarJak: Add new PageObject and then check if you are on right page

tests/e2e/pages/homepage.js
bobdenotter: You'll have to show me how.. I tried making a search.js for it, but the "feature" starts on "homepage" -> `When I visit the "homepage" page`. From there it navigates to the search page, but the orignal 'page' still applies to it, as far as the javascript is concerned.

JarJak: `Then the "search" page is displayed` and it should switch current page object

tests/e2e/features/display_search.feature
bobdenotter: How? :-)

src/Repository/ContentRepository.php
bobdenotter: Yes, nice!

src/Repository/ContentRepository.php
bobdenotter: done

tests/e2e/pages/homepage.js
bobdenotter: Done!

src/Controller/Frontend/SearchController.php
```diff
@@ -37,12 +37,23 @@ public function __construct(Config $config, Environment $twig, TemplateChooser $
...
         $page = (int) $request->query->get('page', 1);
+        $search = $request->get('search', '');
+        $amountPerPage = $this->config->get('general/listing_records');
 
         // @todo implement actual Search Engine
-        $records = $contentRepository->findForListing($page);
+        if (! empty($search)) {
+            $records = $contentRepository->searchNaive($search, $page, $amountPerPage);
+        } else {
+            $records = null;
```
JarJak: it should be empty array, otherwise pagerfanta etc. will break ;)

JarJak: also add a test what will happen if query is empty

bobdenotter: I _think_ `null` is correct, and it should be `?Pagerfanta $records`. I've added a test to illustrate. :-)

Schermafbeelding 2019-03-27 om 16 08 14

JarJak: yes, it gives an exception because this line returns null instead of empty array

src/Controller/Frontend/SearchController.php
```diff
@@ -37,12 +37,23 @@ public function __construct(Config $config, Environment $twig, TemplateChooser $
...
+        }
+
+        $context = [
+            'search' => $search,
```
JarJak: rename search to searchQuery or searchedText

src/Controller/Frontend/SearchController.php
bobdenotter: No, it _breaks_ as soon as that line returns `[]`. If the method accepts only `Pagerfanta $records`, it does break if you're trying to pass in `[]`.. 

On the other hand, `?Pagerfanta $records` allows you to pass in `null` and handle it properly.

src/Controller/Frontend/SearchController.php
bobdenotter: I'd prefer not. The parameter (in `POST` or `GET`) is also `search`, and it has always been this way historically.

src/Repository/ContentRepository.php
```diff
@@ -67,6 +69,37 @@ public function findLatest(?ContentType $contentType = null, int $amount = 6): ?
         return $qb->getQuery()->getResult();
     }
 
+    public function searchNaive(string $search, int $page, int $amountPerPage, bool $onlyPublished = true)
```
JarJak: add return type

src/Controller/Frontend/SearchController.php
JarJak: `historically` ;) Let's move to the future!
We can break anything from past Bolt versions, can't we?
Also some badly named things?

src/Controller/Frontend/SearchController.php
JarJak: Right, so let's return empty Pagerfanta then, to keep context interface clean.

bobdenotter: Ok, sounds good. Done.

src/Controller/Frontend/SearchController.php
bobdenotter: No, we don't have to keep it, but I think it's a properly named thing, in this case. :-)

src/Repository/ContentRepository.php
bobdenotter: Done

src/Controller/Frontend/SearchController.php
stryxx: Gentlemen, please :D Variable name "searchValue" exactly says that it is a value which is searching :) "search" means "activity" we do.

bobdenotter: Alright, alright.. :-)

I've renamed it to `searchTerm` and made the same change in the default templates. I couldn't help myself from keeping the old one backwards compatible, though.

Makefile
```diff
@@ -69,18 +69,8 @@ e2e-wip:
 	cd tests/e2e && npm run kakunin -- --tags @wip && cd ../..
 
 e2e-install:
-	cd tests/e2e
-	npm install
-	node ./node_modules/protractor/bin/webdriver-manager update --gecko=false
-	cd step_definitions
```
JarJak: why removed?

public/theme/skeleton/search.twig
```diff
@@ -47,16 +47,20 @@
             {% endif %}
 
         {% else %}
-            <article>
 
-                <p>
-                    {{ __('general.phrase.no-result-for-search-term', { '%SEARCHTERM%': search|escape }) }}
+            {% if search is not empty %}
```
JarJak: searchTerm

src/Controller/Frontend/SearchController.php
```diff
@@ -37,13 +39,24 @@ public function __construct(Config $config, Environment $twig, TemplateChooser $
...
-        $records = $contentRepository->findForListing($page, $amountPerPage);
+        if (! empty($searchTerm)) {
+            $records = $contentRepository->searchNaive($searchTerm, $page, $amountPerPage);
+        } else {
+            $records = new Pagerfanta(new ArrayAdapter([]));
+        }
+
+        $context = [
+            'searchTerm' => $searchTerm,
+            'search' => $searchTerm, // Keep 'search' for Backwards Compatibility
```
JarJak: what breaks if you remove this line?

Makefile
bobdenotter: Because the files are in git, these symlinks weren't created properly, but we didn't notice yet, because they're already there: https://github.com/bolt/four/tree/master/tests/e2e/step_definitions

public/theme/skeleton/search.twig
bobdenotter: Done, thanks.

src/Controller/Frontend/SearchController.php
bobdenotter: Nothing in the default template. It'll break for people porting a Bolt 3 template, and/or people who are used to using `search`. It's a small effort to support it, and the impact is very low, but it will save some people some headaches.

bobdenotter:
Thanks!

PR #373, opened 2019-03-29
Ensure that we get unique random taxonomies for fixtures

bobdenotter:
This could cause tests to fail, randomly.. We test for "2 categories", but if the fixtures chose "Books" twice, it only stores one. This PR fixes that.


bobdenotter:
Thx! :shipit:

PR #385, opened 2019-04-04
Fixes #384

### LINKED ISSUE #384: Main title doesn't line up properly.
The Main heading in the backen should line up with the main content. Like this: 

Screenshot_2019-04-04_at_20_00_29

However if the viewport gets too wide, it starts "swimming": 

Screenshot_2019-04-04_at_20_00_39

This is probably because i'm not good with Bootstrap. It's perhaps fixable by applying the correct Bootstrap classes, or tweaking the CSS in `assets/scss` somewhere.

zomars:
Done like this to match [this line](https://github.com/bolt/four/blob/bfa8db9fc9f0b87dee3ccbc6ea8baeabffb6f72b/assets/scss/layout/_admin.scss#L83). Although IMHO this can get messy very sooner than later.

### DIFF
--- a/assets/scss/layout/_admin.scss
+++ b/assets/scss/layout/_admin.scss
@@ -51,6 +51,10 @@ $admin-header-height: 70px;
 
     .container {
       padding: 0;
+
+      @include media-breakpoint-up(xl) {
+        max-width: 1240px;
+      }
     }
   }
 


sablonier:
It might fix the alignment right now, but the problem of the file and grid/flex structure resist, right? Is that what you mean with 'messy'? There are parts in ` /assets/scss/modules/admin/admin.scss` with @import of _header.scss and the main `assets/scss/layout/_admin.scss`. In case all can reside in main `assets/scss/layout/_admin.scss` one could move the imported header to the main file, like it is actually i.e. for the body. Or should everything be splitted up in modules?

zomars:
I've been using an ITCSS approach for CSS for ~4 years. Im not saying that we should rewrite completely, but we could take some inspiration to tackle these issues. Mainly we should differentiate base styles, components, objects and utilities.

bobdenotter:
Thanks for this for now! 👍 

About the greater scheme of things: A lot of work on this structure has been done by @jackiboy, but currently he's been busy. If the two of you have suggestions on how to improve he structure, please feel welcome to suggest them. Changes don't have to be radical either. One smaller PR could move some files around, another smaller one changes combines two, and changes the SCSS naming, etc. :-)

PR #386, opened 2019-04-04
Fixes #383

### LINKED ISSUE #383: Sidebar menu has 'empty' submenu for Dashboard
See screenshot: 

Screenshot_2019-04-04_at_19_55_52

There's an 'empty' submenu next to `Dashboard`, noticable because of its dropshadow.. The "Dashboard" should be a plain link, not have a submenu. 

Probably fixable in `assets/js/app/sidebar/Components/Sidebar.vue`

zomars:
This was a tricky one. I don't know if camelCase is how it's supposed to be, but it was causing a false positive while checking `item.contentType !== null`.

### DIFF
--- a/assets/js/app/sidebar/Components/Menu/index.vue
+++ b/assets/js/app/sidebar/Components/Menu/index.vue
@@ -17,12 +17,12 @@
         :href="item.link"
         class="admin__sidebar--link"
         :class="{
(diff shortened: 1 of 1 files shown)


bobdenotter:
Good catch! 👍 

This fixes this particular issue for sure, I'll look into where this inconsistency is coming from!

PR #393, opened 2019-04-12
Make Menu classes use interfaces

### LINKED ISSUE #380: Use CacheWrapper pattern for Sidebarmenu construction.
See https://github.com/bolt/four/pull/376#discussion_r271526884

JarJak:
Make Menu classes use interfaces, chain decorators and add specs of BackendMenuBuilder.

Closes #380


composer.json
```diff
@@ -86,6 +85,7 @@
         "symfony/debug-bundle": "^4.2",
         "symfony/dotenv": "^4.2",
         "symfony/maker-bundle": "^1.11",
+        "symfony/stopwatch": "^4.2",
```
bobdenotter: We will need to sort these out at some point, but Stopwatch needs to be in `require`, not in `require-dev`. On a 'distribution' version of Bolt 4, you'll get the `require` dependencies, and the toolbar, stopwatch and a few others will need to be in there.

JarJak: We'll see if we would ever need a distibution version. Maybe `composer create-project` will be enough

PR #391, opened 2019-04-10
Update composer.json

bobdenotter:

PR #398, opened 2019-04-13
Update translation features and composer dependencies

JarJak:
See https://github.com/bolt/four/pull/391#issuecomment-482815115

Closes #391


composer.json
```diff
@@ -4,7 +4,7 @@
     "type": "library",
     "description": "Bolt 4 skeleton / prototype",
     "require": {
-        "php": ">=7.1.3",
+        "php": ">=7.2",
```
bobdenotter: SF 4 uses 7.1.3. We should stick with that. https://symfony.com/doc/current/reference/requirements.html

src/Menu/BackendMenuBuilder.php
```diff
@@ -199,14 +199,6 @@ private function createAdminMenu(): ItemInterface
             ],
         ]);
 
-        $menu->getChild('Maintenance')->addChild('Fixtures', [
```
bobdenotter: Why remove this one? It doesn't work yet, but we'll add it sooner or later.

composer.json
JarJak: PHPUnit 8 requires PHP 7.2 and 7.1 is close to EOL. But I can revert it, no problem.

src/Menu/BackendMenuBuilder.php
JarJak: I don't think we could use migrations in Bolt.

JarJak: reverted

composer.json
JarJak: reverted

bobdenotter: Thanks. I think it's beneficial to us to stick with SF's version requirements :-)

PR #400, opened 2019-04-13
Fix integration with API Platform 2.4, the other way around

JarJak:
Another approach.

See https://github.com/api-platform/core/pull/2732#discussion_r275135410

Closes #399.


bobdenotter:
I would prefer this one over #399. Don't forget the changes in composer.json in this PR, though.

PR #404, opened 2019-04-15
[WIP] bootstrap-vue

sablonier:
Adding bootstrap-vue in dependencies


JarJak:
>ModuleNotFoundError: Module not found: Error: Can't resolve '../static/brand/bolt4-icon.svg' in '/home/travis/build/bolt/four/assets/scss'

assets/scss/bolt.scss
```diff
@@ -1,5 +1,9 @@
+@import './themes/default';
 @import './vendor/bootstrap/bootstrap';
-@import './init/init';
+@import './vendor/bootstrap-vue/bootstrap-vue';
+
+/* old */
```
bobdenotter: One of _the_ main advantages of Git is version control. Feel free to delete these "old" things, because if we might need them one day, we can simply go back in git history. 

Doing so keeps the code base nice and tidy. 

(in other words: Comments are for comments. Not for deleted code)

assets/scss/themes/default.scss
```diff
@@ -42,9 +42,44 @@ $bootstrap-colors: (
...
+$enable-gradients: false;
+
+// CSS
+// -------------------
+
+// Toolbar
+.svg-brand {
+  min-width: 20px;
+  min-height: 20px;
+  background-image: url('../../static/brand/bolt4-icon.svg');
```
bobdenotter: Seems like this one's missing. :-)

webpack.config.js
```diff
@@ -28,9 +25,9 @@ Encore.addPlugin(
 
   .addEntry('bolt', './assets/js/bolt.js')
   .addStyleEntry('theme-default', './assets/scss/themes/default.scss')
-  .addStyleEntry('theme-light', './assets/scss/themes/light.scss')
+  /*.addStyleEntry('theme-light', './assets/scss/themes/light.scss')
```
bobdenotter: Why have you commented these out?

assets/scss/themes/default.scss
sablonier: git add is not my friend.

webpack.config.js
sablonier: because this is not working with my current changes. it's WIP. anyway thanks for your comments.

sablonier: Question: What do you think about light and dark theme here ? Is it really needed for 'lightweight and simple'? Or isn't creating custom themes already sufficient ?

assets/scss/themes/default.scss
bobdenotter: Ping me on slack, if you'd like some pointers! :-)

webpack.config.js
bobdenotter: This is for the _backend_, and doesn't really impact the way people _build_ sites with it. Right now, these are "baked in", but this feature will need work for sure. But, it will become better fleshed out sooner or later, so we shouldn't remove them. :-)

JarJak: We need to make way for devs to provide their own admin themes.

bobdenotter:
This PR has been opened a full half year ago. @sablonier, if you want to finish it, let me know what I can do to help. 

Otherwise let's close it, because keeping it open without it being actionable is just noise. :-/

sablonier:
Just switch it off. I don't want to make noise with this as I don't find the time to move forward :(

bobdenotter:
Ok, too bad, but it is what it is.. Let me know if/when you have time to pick up work on Bolt, We'll happily provide guidance! :-)

PR #405, opened 2019-04-18
Use default filter instead of is defined in Twig

JarJak:
Because it is safer


tests/spec/Bolt/Menu/BackendMenuBuilderSpec.php
```diff
@@ -10,7 +10,6 @@
 use Bolt\Twig\ContentExtension;
 use Knp\Menu\FactoryInterface;
 use Knp\Menu\ItemInterface;
-use Knp\Menu\MenuItem;
```
bobdenotter: strange that ECS didn't pick this one up, earlier.. 🤔

JarJak: It was looking only into src folder

Issue #150, opened 2018-11-20
Create Content API endpoints for Frontend


JarJak:
GET single and GET multiple, with sorting, search etc.
Can be tricky due to new Content - Field (EAV?) relation.
Hydra annotations should be added too.

bobdenotter:
Is this different from what we already have listed in `/api`? If so, and it's considered useful to have, i'd be in favor of adding it.. 

I believe @jackiboy also had an opinion on GraphQL and/or Hydra. Let's discuss on how to proceed with this. :-)

bobdenotter:
Added it to the Zenhub board: https://app.zenhub.com/workspaces/four-5befffc74b5806bc2bf587e7/boards?repos=135177880

JarJak:
@bobdenotter jest it is different: in frontend we need only GET endpoints (+ POST in boltforms extension) and /fields endpoint doesn't make sense IMO, /content should return data the same way setcontent does (actually setcontent could use API Models internally), so:
- contents with fields with data
- paging
- searching
- returnsingle
- relations (bidirectional too)
etc.

bobdenotter:
I know @nestordedios already did some work on this.. We can use the API to fetch Records, _with_ all its Fields.. @nestordedios, maybe you can chime in?

Futhermore, we might need `POST` (or `PUT`) endpoints for Fields. The main edit interface will be traditional server-request-response type interaction, but Jack made a start on the interface for "quick edit", so there it'd make sense if you could update a specific field like "title" or "status".

> Another thing is that IMO if we do headless AND traditional frontend, all API endpoints should be prefixed with /api 

Sure.. Perhaps even `/bolt/api/`?

nestordedios:
@JarJak I'm not very familiar with API Platform yet and there are indeed points to improve.

>2. /fields endpoint doesn't make sense IMO

The reason I added the `Fields` entity to the Model is because the `Content` Entity has a collection of `Fields` Entities, and by not adding them to the Model the fields where not fetched in `api/contents/{id}` How to get that endpoint working without exposing `api/fields` in the API I haven't managed.

PR #406, opened 2019-04-18
Fields API

JarJak:
As an addition to #150 I've added API endpoints for fields, so now you can make a request like 
`GET /api/contents/1/fields.json` 
and it will return all names and values fields of this content. Values are always arrays here, like:
```
[
  {
    "name": "Title",
    "value: [
      "Hello world!"
    ]
  }
]
```

TBD: I can also add type of field to it in case someone will find it is useful information in frontend for proper rendering.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
         "behat/mink-extension": "^2.3",
         "behat/mink-goutte-driver": "^1.2",
         "behatch/contexts": "^3.2",
-        "coduo/php-matcher": "^3.1.2",
+        "coduo/php-matcher": "dev-patch-2@dev",
         "dama/doctrine-test-bundle": "^5.0",
         "doctrine/doctrine-fixtures-bundle": "^3.1",
         "lakion/mink-debug-extension": "^1.2",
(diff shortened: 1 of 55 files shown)


composer.json
```diff
@@ -66,7 +66,7 @@
         "behat/mink-extension": "^2.3",
         "behat/mink-goutte-driver": "^1.2",
         "behatch/contexts": "^3.2",
-        "coduo/php-matcher": "^3.1.2",
+        "coduo/php-matcher": "dev-patch-2@dev",
```
bobdenotter: Not a blocking issue, but is this `coduo/php-matcher` robust? I think this is the second or third time we've had to use it as non-stable, right?

JarJak: Because I've made another PR to them ;)

JarJak:
>TBD: I can also add type of field to it in case someone will find it is useful information in frontend for proper rendering.

@bobdenotter wdyt? would it be useful for headless frontend?

bobdenotter:
> wdyt? would it be useful for headless frontend?

Yes, very much! it'll be great for context. If something has a value of `foo`, and it comes from a `type: text` you might want to display it differently than something that's a `type: select`. :-)

JarJak:
@bobdenotter type added in API response. Also I've made Dynamic DiscriminatorMap :D

PR #410, opened 2019-04-21
[hotfix] Make sure `{{ config }}` is set

bobdenotter:
Fixes a regression in #404, where the API and translations screens don't show, because they don't use our controllers, and don't have our `config` set.

Screenshot 2019-04-21 at 13 32 13

### DIFF
--- a/tests/e2e/features/backend_api.feature
+++ b/tests/e2e/features/backend_api.feature
@@ -1,4 +1,4 @@
-Feature: Visiting Dashboard
+Feature: Visiting API backend page
 
     Scenario: As an admin I want to see API page
         Given I am logged in as "admin"
--- a/tests/e2e/features/backend_api.feature
+++ b/tests/e2e/features/backend_api.feature
@@ -1,0 +1,7 @@
+Feature: Visiting Dashboard
+
+    Scenario: As an admin I want to see API page
+        Given I am logged in as "admin"
+        When I visit the "backend_api" page
+        Then there is element "header" with text "Bolt API"
+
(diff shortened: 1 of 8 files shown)


JarJak:
Isn't config global? I think it should be.

bobdenotter:
It gets added as global in our `TwigAwareController::renderTemplate()`, but templates that don't use it (like API and Translations) will not have it. 

Do you have a suggestion on how to insert it otherwise?

JarJak:
https://symfony.com/doc/current/templating/global_variables.html

But it is fine to use it only in TwigAwareController

bobdenotter:
> https://symfony.com/doc/current/templating/global_variables.html 

That's only for specific single variables. Not for things like our Config object. 

I think this PR is correct the way it is.

JarJak:
>That's only for specific single variables. Not for things like our Config object.

https://symfony.com/doc/current/templating/global_variables.html#referencing-services

bobdenotter:
Done.

tests/e2e/features/backend_translations.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Visiting Translations page
+
+    Scenario: As an admin I want to see API page
```
JarJak: see Translations page

tests/e2e/features/backend_api.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Visiting Dashboard
```
JarJak: Visiting API docs page

tests/e2e/features/backend_translations.feature
bobdenotter: Done

tests/e2e/features/backend_api.feature
bobdenotter: Done

PR #416, opened 2019-04-28
Expanding on fixtures

### LINKED ISSUE #252: Frontend: Make "Show relations for a record" work
Additional issue for #232 

Depends on #163

bobdenotter:
- Adding Relations as fixtures
 - Adding `with-images` and `without-images` groups, to speed up fixtures, if more images are not needed
 - Removing 'risky' tests, using them with `--group=without-images` instead. 

Also fixes #252, because tests broke now that Fixtures have relations


JarJak:
Code is OK but please wait with merging, I will add tests to this PR for fetching unpublished relations.

PR #420, opened 2019-05-09
[hotfix] Add missing attributes to json_records

### LINKED ISSUE #407: Missing attributes in `records|json_records`
After the introduction of #400, a few things are not displayed correctly in the backend. Namely this: 

Schermafbeelding 2019-04-19 om 18 00 05
Schermafbeelding 2019-04-19 om 18 00 16

The Vue component that breaks is `assets/js/app/listing/Components/Table/Row/_Meta.vue`. 

Especially these values: 

`{{ record.author.displayName }}` -> can be replaced by `{{ record.authorName }}`

These two should be added: 

 - `{{ record.icon }}` -> which is coming from the `icon_one` from the contenttype definition
 - `{{ record.status }}` -> Which is the current status of the record.

bobdenotter:
Fixes #407.

### DIFF
--- a/src/Entity/ContentExtrasTrait.php
+++ b/src/Entity/ContentExtrasTrait.php
@@ -38,7 +38,7 @@ trait ContentExtrasTrait
             'excerpt' => $this->contentExtension->getExcerpt($content),
             'link' => $this->contentExtension->getLink($content),
             'editLink' => $this->contentExtension->getEditLink($content),
-            'icon' => $this->contentExtension->getIcon($content),
+            'icon' => $this->getIcon(),
         ];
     }
 }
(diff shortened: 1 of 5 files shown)


src/Entity/ContentExtrasTrait.php
```diff
@@ -38,6 +38,7 @@ public function getExtras(): array
             'excerpt' => $this->contentExtension->getExcerpt($content),
             'link' => $this->contentExtension->getLink($content),
             'editLink' => $this->contentExtension->getEditLink($content),
+            'icon' => $this->contentExtension->getIcon($content),
```
JarJak: 'icon' => $content->getIcon()

bobdenotter: Good catch, PR updated.

PR #421, opened 2019-05-09
Allow for caching of widgets

bobdenotter:
Makes a big difference, especially for widgets like 'news'...

Before: 

Screenshot 2019-05-09 at 22 02 39

After: 

Screenshot 2019-05-09 at 22 02 54

### DIFF
--- a/src/Widget/CacheAware.php
+++ b/src/Widget/CacheAware.php
@@ -13,6 +13,4 @@ use Psr\SimpleCache\CacheInterface;
 interface CacheAware extends WidgetInterface
 {
     public function setCacheInterface(CacheInterface $config): void;
-
-    public function cachedInvoke(): string;
 }
--- a/src/Widget/CacheAware.php
+++ b/src/Widget/CacheAware.php
@@ -12,5 +12,5 @@ use Psr\SimpleCache\CacheInterface;
  */
 interface CacheAware extends WidgetInterface
 {
-    public function setCacheInterface(CacheInterface $config): void;
+    public function setCache(CacheInterface $config): void;
 }
--- a/src/Widget/CacheAware.php
+++ b/src/Widget/CacheAware.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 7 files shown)


src/Widget/CacheTrait.php
```diff
@@ -0,0 +1,55 @@
...
+    /** @var string */
+    private $key;
+
+    public function setCacheInterface(CacheInterface $cache): void
+    {
+        $this->cache = $cache;
+        $this->key = $this->createKey();
+    }
+
+    public function cachedInvoke(): string
```
JarJak: can't this method just overwrite __invoke() ?

bobdenotter: I'm going to try something else to make it more elegant. 

Overriding it would be nice, but won't work in this case. A trait can override methods from the base (extended) class, but then there's no such thing as a `::parent`.

src/Widget/CacheAware.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Psr\SimpleCache\CacheInterface;
+
+/**
+ * Interface CacheAware - Widgets that make use of caching need to implement
+ * this interface, in order to have their contents cached.
+ */
+interface CacheAware extends WidgetInterface
+{
+    public function setCacheInterface(CacheInterface $config): void;
```
JarJak: setCache* or setCacheAdapter*

src/Widget/SnippetWidget.php
```diff
@@ -27,7 +27,7 @@ public function __construct(
         $this->zone = $zone;
     }
 
-    public function __invoke(array $params = []): string
+    public function run(array $params = []): string
```
JarJak: protected

src/Widget/CacheAware.php
bobdenotter: Done

src/Widget/SnippetWidget.php
bobdenotter: Done

PR #428, opened 2019-05-12
Make Weather widget work

bobdenotter:


bobdenotter:
Note: Before 4.0.0 stable, this will move to an extension. Currently I have it in core, to be able to work on the widgets, but something silly like this has no place in core. :-)

PR #488, opened 2019-06-24
[WIP] Setcontent for front

stryxx:
Checklist copied from #116. API won't be 100% the same, but functionality should be kept.
 ### Basic
- [x] `{contenttype}`
- [x] `{contenttype}/first/{number}`
- [x] `{contenttype}/last/{number}`
- [x] `{contenttype}/random/{number}`
- `random` is not going to work with Doctrine ORM queries, see [answer on Stack Oveflow](https://stackoverflow.com/questions/10762538/how-to-select-randomly-with-doctrine) for info.
- warning An alternative is to make a special Twig function/filter to handle random records.
- [ ] Random should be refactored for better performance.

`{contenttype}/{id}` - **_it cannot be supported because `id` is not unique for every of contenttype. `id` is unique globally, so it not possible to get correct `contenttype` by `id`._**
- [x] `{contenttype}/{slug}`
 
 
 ### WHERE
- [x] `{key}: {value}`
- [x] `{key}: {operator}{value}` where operator is one of `%` `<` `<=` `=`
       
     warning Currently the operations will not work for numbers/dates/etc. because of JSON-encoded values.
 
 
 ### WHERE advanced
- [x] `{key}: {value1} || {value2}` or `{key}: {value1} && {value2}`
- [x] `"{key1} ||| {key2}": "{value1} ||| {value2}"`
- [x] `{key}: 'now'` and other special datetime selectors (e.g. `now`, `today`, `yesterday`, `last year`, `next thursday`).
 
 
 ### Special
- [x] `limit {number}`
- [x] `orderby {field}` or `orderby -{field}`
       
     warning This works for fields that have only a single value (think of it as a column)
- [x] Multiple sortorders (comma-separated)
- [x] `returnsingle`
- `printquery` - _**I'm not sure if we should have this option**_
       
     warning Currently shows DQL. Let's see if we can show a _runnable query_ instead.
 
 ### Advanced
 
- [ ] `allowpaging`
- [ ] `{contenttype}/search/{num} where filter: {term}`
- [ ] `{ct1},{ct2},{ctN}/search/{num} where filter: {term}`
 
 
 ### Misc
- [ ] Implementations for locales, taxonomies and relationships.
- [ ] Proper integration with Config + ContentType data (such as max results per page).

### New feature
- [ ] Add getcontent function for twig allows to get content by direct GraphQL query.


bobdenotter:
Good to see you back in action on this! 👍

stryxx:
I happy too ;) I'm back with full bag of good (imo :P) ideas ;)

stryxx:
I happy too ;) I'm back with full bag of good (imo :P) ideas ;)

bobdenotter:
I haven't commented much on this, but that doesn't mean i'm keeping an eye on it. ;-)

Just let me know when you think it's done, or ready for a second pair of eyes. 👍

stryxx:
Thanks Bob, I'm going to let you know ;) It comes to the end ;)

bobdenotter:
This PR has been lingering for two and a half years now. I think it's time to close it. 😔

PR #495, opened 2019-06-28
Wttr.in is down, and is breaking the bolt backend. Fix for a better check.

anketwokings:
If wttr.in is down, (like, now) it doesnt return an empty array, but might also be an error. so the check if empty does not work, the widget gets rendered and breaks bc the array is not what its expected to be.


anketwokings:
Oh! Sorry, it contains commits that shouldn't be in here. Wait! Will fix this! -- working on pull request foo..

anketwokings:
> Oh! Sorry, it contains commits that shouldn't be in here. Wait! Will fix this! -- working on pull request ..

Ok, fixed!

bobdenotter:
Tests, gotta love 'em and hate 'em at the same time. ;-) 

Your PR does put the finger on the sore spot, but needs to be fixed slightly differently. I'll make some adjustments.

PR #498, opened 2019-06-28
Adds hidden-danger class - #493

### LINKED ISSUE #493: Add a new “hidden-danger” button style
This new button style should look like a neutral button, but on `hover` or `focus`, it should get the “danger” state. Basically, so we can use it multiple times on one page, without it looking too overwhelming.

anketwokings:
fixes #493 
+ small extra comments on yiq


anketwokings:
Guess: Probably failing tests bc the wttr.in service is down and breaking Bolt. (timeouts)

Issue #474, opened 2019-06-18
Make backend responsive


bobdenotter:
The backend works OK-ish on Desktop formats. It looks crappy on mobile. Should be improved

bobdenotter:
Also: Check is how we use "Grid" is OK.. If not, either use Grid properly or refactor to Flexbox

sbonardt:
IMO we should rework this and start mobile first. @anketwokings, @sbonardt and anybody interested are thinking about streamlining this and getting this in properly

sbonardt:
Made changes for wide screens. Still WIP and under development, but already a usable solution

sbonardt:
The backend is now responsive, mobile first. I would still like to rewrite all classes as this isn't BEM, but I'll save that for another issue and PR.

The 'admin__body--container' is the main css Grid container for the content. It has 12 columns. The sub element 'admin_body--main' is by default 12 columns wide. If there is anything present in the sidebar region this container gets the additional class 'admin__body--container--has-sidebar' and a sub element 'admin_body--sidebar'. The 'admin__body--main' element will then be 9 columns wide and the 'admin__body--sidebar 3 columns'

Sub elements like content-listings or fields can inherit/work upon this grid.

There are still some tweaks and alignment improvements to be made in the near future.

PR #332, opened 2019-02-22
[RFC][WIP] Introduce Symfony Forms

JarJak:
Let's use Symfony Forms everywhere.

Advantages:
- CQRS security by default
- unit testable
- front-end and back-end validation
- allows adding custom validation to any field
- ability to use form themes for fields
- allow adding more field types (by extensions?)

TODO:
- frontend / Vue components need to follow up
- Manual tests: Check if all fields, relations, taxonomies are saved correctly. Test if validation works as expected.
- E2E tests won't pass until some frontend dev will fix forms in twig.

Closes #331

PR #499, opened 2019-06-28
Redo of layout backend, making it responsive and mobile first. MVP for now

sbonardt:
#474 

The initial setup for desktops only -in flexbox and grid- has been rewritten. Grid now only kicks in on wide screens for the admin-body-container only. The rest of the layout works without grid or flexbox (because it's not needed). Mobile has a base version with a slide in sidebar and toggle button.

This is rework to a minimal working version for now. It's work in progress but looks nice. Suitable to work with and further develop on.


sbonardt:
@bobdenotter Seems all checks fail at the same stuff. We think this is due to the fact the weather widget is broken. I had to comment it out to get my dev up and running. Seems the service used for the weather widget is not available due to a problem over there.

JarJak:
Hey @sbonardt! Could you also update all the fields views to be compatible with those  coming from Symfony Forms (https://github.com/bolt/four/pull/332)?

sbonardt:
@JarJak I'll check with @bobdenotter this monday if that is something I can pick up as well 👍🏻

bobdenotter:
> Hey @sbonardt! Maybe you could also update fields views to be compatible with those coming from Symfony Forms (#332)?

@JarJak We need to figure out how to progress on your PR. I still have no clue how I can work on that PR in it's current status. Maybe have a call again sometime in the upcoming week?

assets/js/app/notifications/index.js
```diff
@@ -2,4 +2,19 @@ import $ from 'jquery';
 
 $(document).ready(function() {
   $('.toast').toast('show');
+  // add a js class to indicate we have JS enabled. Needs a change to either modernizr or something custom @bobdenotter
```
bobdenotter: Why is this code in `notifications`? I think there ought to be a better place for this.

assets/scss/init/_variables.scss
```diff
@@ -14,7 +14,7 @@ $big-z-index: 1000;
 $font-family-base:              -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
 $headings-font-family:          'Source Sans Variable', sans-serif;
 
-$font-size-base:                0.9rem;
+$font-size-base:                1rem;
```
bobdenotter: Not sure if I like this. If we do re-set the base font size to 1 rem, we should also review line-heights and paddings here and there to keep it balanced.

assets/js/app/notifications/index.js
sbonardt: I deleted the styling for the narrow stuff because I thought we were going to ditch it for V4. But I understand differently now. I''ll put it back in place, but it needs work together with the HTML. For the location of the JS, it's temporary, so we have to discuss where to put this and other generic UI js.

assets/scss/init/_variables.scss
sbonardt: This was step 1 in a redo to come. I would like to set the base to 1rem/16px (assumed browser default) and work from there to the rest.

sbonardt: OK. The common js is now in common.js

assets/js/app/notifications/index.js
sbonardt: This has been transferred to common.js

bobdenotter:
It looks like the tests are now failinge because of your changes: 

>   ✖ Then there is element "header" with text "Dashboard" # step_definitions/steps.js:10
 >      TimeoutError: Wait timed out after 10002ms

>    ✖ And there is element "profile_text" with text "Hey, Admin!" # step_definitions/steps.js:10
>       TimeoutError: Wait timed out after 10001ms

Issue #508, opened 2019-07-08
Make tests run more smoothly


bobdenotter:
- [x] Make Kakunin output less warnings
- [x] Don't crawl images from Unsplash when runnin tests
- [x] Don't break if weatherwidget can't reach wttr.in, or if it's over quota
- [x] Make Kakunin run again. See: https://github.com/TheSoftwareHouse/Kakunin/issues/188

bobdenotter:
Fixed after transitioning to Behat.

PR #520, opened 2019-07-10
Update Kakunin, other deps, make it work again

bobdenotter:
Part of #508


sbonardt:
OK, so it only needed an update...
:thum

Issue #510, opened 2019-07-08
On a page with "widecontent" aside overlaps usable space


bobdenotter:
See screenshot: 

![Screenshot 2019-07-08 at 21 28 41](https://user-images.githubusercontent.com/1833361/60837017-77ae8280-a1c7-11e9-9db6-462916b0e4d4.png)

sbonardt:
I can fix this but I need some more info on the approach of the content. Currently the aside is written out although there is no content in it. I'm planning on running a check on the aside block if it has content set in the subtemplate. This way we can still define the HTML for the aside in the main template and not give others 'control' over the structure, which would happen if we move the HTML for the aside to the sub templates. I'll rework the styling for this as well

bobdenotter:
Fixed in #521

PR #521, opened 2019-07-10
Aside block var check, add class and rewrite grid layout. Removed widecontent class

sbonardt:
…econtent class


sbonardt:
Fixes #510

sbonardt:
@bobdenotter Removed the redundant container bloks

Issue #492, opened 2019-06-25
[RFC] Discussion starter: how could we best tackle Extensions?


doenietzomoeilijk:
_This relates to #134_ 

A couple of weeks ago I asked Bob how he figured Extensions would fit into B4, and he figured

> That's still up for debate.. _so far_ i've only made some preliminary tests, and i'm leaning towards going with the paradigm i've been using for most things with Bolt 4: Use what's there.
> 
> In this case: We have composer (and packagist) to handle installation, versioning, distribution and all
> 
> We have SF 4 (and its awesome autowiring) and autoloading to make sure stuff gets autoloaded.

after which I hinted at having implemented something similar in an application somewhere, and promising to do a little write-up about how that works. That got me thinking that the way I used in that application might not actually fullfil the entire need Bolt has, but at least it could get things started, people might have some ideas, and so on. So without further ado:

# How to get Symfony to do special things to classes implementing a given Interface

Actually the whole thing is quite simple, and consists of the following parts:

- An Interface for classes to implement, which will signify that the implementing class is (in this case) an extension;
- A repository (not the Doctrine kind) or registry to keep all these implementors in, which can be injected. Come to think of it, maybe you could call this a container within the container;
- A bit of glue to thet the former into the latter, at kernel boot time.

## The Interface

In fact, the interface doesn't need to _do_ anything to be useful. Bob mentioned that the Interface could define an initialize function to be run (possible); what I did in my project was make sure the implementor can identify itself. 

All code below is barely above pseudocode level, in real life this should all be properly namespaced (probably under \Bolt\Extension) et cetera.

```php
<?php

interface ExtensionInterface
{
    /** Defines the tag that Symfony will get to know our Extensions by */
    public const CONTAINER_TAG = 'bolt.extension';

    /** Let's say that all Extensions should be able to identify them with a polite name */
    public function getName(): string;
}
```

That's all.

## The registry

For the registry, we need to be able to add implementors, retrieve them, etc.

```php
<?php
class ExtensionRepository
{
    /** @var ExtensionInterface[] **/
    protected $extensions = [];

    public function addExtension(ExtensionInterface $extension): void
    {
        $this->extensions[\get_class($extension)] = $extension;
    }

    /** @return ExtensionInterface[] */
    public function getExtensions(): array
    {
        return $this->extensions;
    }
}
```

## Getting the implementors into a repository

Finally, at compile time, we need to grab all our implementing classes and have them stuffed into the repository. This is done using a CompilerPass:

```php
[...]

Issue #134, opened 2018-11-17
⭐️ Make "Extensions"


bobdenotter:
Things to do: 

- [x] Make extensions 'autoload' as a compiler pass
- [x] Make extensions call their 'initialising' method after booting
- [x] Make possible for extensions to have their own config file
- [x] Make extension add widgets/snippets
- [x] Make extensions add Twig funcions & filters
- [ ] Provide things like `Request`, `Response`, `Config` in extensions
- [ ] Make Autowiring work for extensions
- [x] Make command for `extensions:list`
- [ ] Make command for `extensions:info`
- [ ] Make command for `extensions:enable`
- [ ] Make command for `extensions:disable`
- [x] Make controller / UI screen in backend
- [ ] Create `make:extension` command
- [ ] Make it so extensions can add FieldTypes.

bobdenotter:
Note: This package might be convenient: 

 - https://www.reddit.com/r/PHP/comments/cl206o/new_package_composer_helper_classes_for_finding/
 - https://github.com/drupol/composer-packages

drupol:
> Note: This package might be convenient:
> 
>     * https://www.reddit.com/r/PHP/comments/cl206o/new_package_composer_helper_classes_for_finding/
> 
>     * https://github.com/drupol/composer-packages

Hey cool, thanks for this! I built this package especially for that purpose. Allowing people to easily find package extensions :-)

Thanks!!!

bobdenotter:
@drupol If i bump into any issues implementing it, i'll be sure to reach out! 👍

bobdenotter:
Closing, see #492

PR #559, opened 2019-08-03
Register composer installed extensions, remove Weather Widget

bobdenotter:
See #134, #492


src/Extension/ExtensionRegistry.php
```diff
@@ -14,11 +16,40 @@ class ExtensionRegistry
...
         $this->extensionClasses[] = $extensionClass;
     }
 
+    private function addComposerPackages(): void
+    {
+        // We do a try/catch here, instead of using `method_exists`. This is
+        // because PHPStan is being a smart-ass, and takes the state of the
+        // generated `Types` class into account. And that's exactly the point.
+        try {
+            $packages = Types::boltExtension();
```
drupol: I think you might remove the `try/catch` if you use:

<?php

iterator_to_array(Types::get('bolt-extension'))

bobdenotter: Ah, yes! much nicer. I've fixed it straight away! 👍

PR #563, opened 2019-08-05
Working on Extensions

bobdenotter:


harikt:
Hey a dump question,

I am seeing all these extensions, but where are they going to be displayed. The docs are for version 3, so not sure how it works in 4. ie why.

harikt:
Oh I noticed this now.

bobdenotter:
Hi @harikt , As you've probably seen, there's the "Extensions" screen in the UI, but you can also get a list on the command line using `bin/console extensions:list`. 

You'll be able to activate/deactivate them the samy way soon, without having to fully delete or uninstall them.

harikt:
To make it clear : Basically my problem was how the extensions and where it was rendered. But I see this on the themes. 

Thank you for the other commands and functionalities you mentioned. I will try to dig more into it.
If this is a right place to ask, is the docs for v4 some where or in a progress.

bobdenotter:
@harikt Documentation needs to be written yet.. I can see how it can be a bit confusing now, with "Extensions" that get autoloaded, and "Widgets" that are slightly different. (The main reason for this is, that _one_ extension can have _multiple_ Twig extensions or Widgets)

That said, docs need to be written, but as soon as there is something, i'll make it public as well. 

You can ask questions here, but feel free to join us on our slack channel as well: https://slack.bolt.cm

Issue #514, opened 2019-07-09
Make login accept (slugified) username, as well as email.


bobdenotter:
Because: 

 - It's more convenient for users
 - Mobile devices make it hard to do case-correct input in input fields, inputting `admin` as `Admin`, making it a hassle for users

bobdenotter:
Fixed in #633 by @fatguytyson

PR #633, opened 2019-09-17
Issue [#514](https://github.com/bolt/core/issues/514)

fatguytyson:
Changed the repository method to accept the credential(string) and has
the storage engine compare the lowercase values. At least this is
consistent to the strings compared so it should be a consistent
experience


bobdenotter:
Hi @fatguytyson ,

Thanks! Looks like this one partially fixes #514, right? 

Could it be adapted to accept email as well, so we can fix that in one go?

PR #642, opened 2019-09-24
fix(config): fix form_themes in twig config

fabienChaillou:
This PR fix this errors:

[X] Fix form_themes key in twig config
[X] Update `bootstrap_3_layout.html.twig`

bootstrap_3_layout html html twig


fabienChaillou:
Thanks @bobdenotter

PR #647, opened 2019-09-27
Explicitly use Debian Stretch as the base image

doenietzomoeilijk:
Otherwise we end up with Debian Buster, which results in a package error when running `docker-compose up`:

```E: Package 'ttf-freefont' has no installation candidate 
ERROR: Service 'php' failed to build```


doenietzomoeilijk:
The better fix is to use Buster instead of Stretch, as outlined in https://github.com/bolt/project/pull/8

PR #668, opened 2019-10-05
Fix singleton contentType menu bug when no record exists

DjordyNelissen:


assets/js/app/sidebar/Components/Menu/index.vue
```diff
@@ -40,7 +40,7 @@ export default {
   props: ['menu'],
   methods: {
     singleton(item) {
-      if (item.submenu !== null) {
+      if (item.submenu.lenght && item.submenu !== null) {
```
GDmac: typo in length property
```suggestion
      if (item.submenu.length && item.submenu !== null) {
```

PR #693, opened 2019-11-01
Display selects in frontend, add simple Notifications framework

bobdenotter:


templates/_partials/notification.html.twig
```diff
@@ -0,0 +1,107 @@
+<style>
```
I-Valchev: Do styles go here, rather than in assets?

bobdenotter: Yes, and note the `!importants` as well. These notices will show up in templates of users, so we can't pull in external CSS'es, nor do we know what it looks like. By inlining it, we can mostly ensure it'll be readable in most circumstances.

I-Valchev: OK! 👍

Issue #518, opened 2019-07-09
On "Edit users" screen, display session info


bobdenotter:

bobdenotter:
Fixed by #709

Issue #702, opened 2019-11-03
Update to Doctrine DBAL 2.9.3 breaks tests


bobdenotter:
Doctrine DBAL 2.9.3 update added this particular line: https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L365

```
$this->transactionNestingLevel = 0;
```

This causes Bolt's tests to break. For example, run `composer update`, and then: 

```
$ ./vendor/bin/phpunit --testdox tests/php/Repository --stop-on-failure                                                  (disable-tests✱)
PHPUnit 7.5.17 by Sebastian Bergmann and contributors.

Bolt\Tests\Repository\TaxonomyRepository
 ✔ Search by type

In StaticDriver.php line 142:

  There is no active transaction

```

The "There is no active transaction" is caused by `DAMA\DoctrineTestBundle` trying to rollback a DB, while there's no apparent transaction going. 

It's invoked here: https://github.com/dmaicher/doctrine-test-bundle/blob/29882b0d1a815f4819126ef714931bb24a31cbaa/src/DAMA/DoctrineTestBundle/PHPUnit/PHPUnitListener.php#L39

```
        public function endTest(\PHPUnit\Framework\Test $test, float $time): void
        {
            StaticDriver::rollBack();
        }
```

I suspect this is happening in Bolt, because of how we're using the DB in `DbAwareTestCase`.  

@JarJak Since you've initially written this, could you please look into this?

bobdenotter:
See also: https://github.com/dmaicher/doctrine-test-bundle/issues/94

dmaicher:
I'm on vacation over the weekend. Will look into this next week :+1:

bobdenotter:
@dmaicher Thanks in advance! Have a nice vacation! 👋

dmaicher:
I am able to reproduce this locally now and will look into the issue today :blush: 

```
it@f5e2da3f8c45:/var/www/core  (master) $ make test
vendor/bin/phpspec run
                                      100%                                       25
4 specs
25 examples (25 passed)
661ms

vendor/bin/phpunit
PHPUnit 7.5.17 by Sebastian Bergmann and contributors.

.........................................
In StaticDriver.php line 142:
                                  
  There is no active transaction  
                                  

Makefile:60: recipe for target 'test' failed
make: *** [test] Error 255

PR #709, opened 2019-11-06
session info shown on view users

I-Valchev:
What is the correct way to display dates on the Bolt backend?

[This twig extension](https://twig-extensions.readthedocs.io/en/latest/date.html) allows relative times such as `4 seconds ago` and `in one month`, but I could not install it correctly because it requires the higher `doctrine/dbal` version, on hold because of #702 .

For now this is shown e.g. `2019-11-06 12:35:19`


I-Valchev:
Fixes #518

Issue #334, opened 2019-02-22
Update .env files config


JarJak:
Details: https://symfony.com/doc/current/configuration/dot-env-changes.html#updating-my-application

More about it: https://symfony.com/blog/improvements-to-the-handling-of-env-files-for-all-symfony-versions

After all, call `composer sync-recipes --force` from https://symfony.com/blog/new-in-symfony-flex-1-2

bobdenotter:
Calling `composer sync-recipes --force` breaks a _lot_ of things. We should leave this as-is.

PR #711, opened 2019-11-07
Update/env config files

nestordedios:
This PR is related to #334 and is still work in progress.


nestordedios:
Closing since it does not seem to be very relevant at the moment.

PR #715, opened 2019-11-12
fix test database transactional isolation

dmaicher:
Fixes https://github.com/bolt/core/issues/702

This doctrine bugfix actually made this issue visible :blush: 
https://github.com/doctrine/dbal/commit/a9b7bf855c6f391f96773f844d7158500394bcef#diff-7b1e94fb7dd408985f7b7973330dc2edR359

Here the issue was that dropping/creating the database is not transactional anymore. It was implicitly committing any open transaction on the underlying static connection from `dama/doctrine-test-bundle`. This is also mentioned in the readme of the bundle.

Before the doctrine bugfix this issue was hidden as the internal connection state (transaction nesting) was wrong.

So this PR fixes it by refactoring the database creation to be done **before** running any test. So there is no need anymore to do this within `setUp` or `tearDown`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "cocur/slugify": "^3.2",
         "composer/composer": "^1.9",
         "doctrine/annotations": "^1.7",
-        "doctrine/dbal": "2.9.2",
+        "doctrine/dbal": "^2.9",
         "doctrine/doctrine-bundle": "^1.11",
         "doctrine/doctrine-cache-bundle": "^1.3.1",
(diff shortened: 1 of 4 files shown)


bobdenotter:
Hi @dmaicher ,

Thanks for your work on this! 🤗 

I see tests are passing on PHP 7.2 now.. Since SF 5 will be released soon, we'll drop support for PHP 7.1 (keeping parity with SF stable). 

The failing test on 7.3 is a fluke (in the embed)

PR #718, opened 2019-11-13
Feature/imagelist contenttype

I-Valchev:
Implements imagelist content type as:
```
example:
  type: imagelist
```

### DIFF
--- a/assets/js/app/editor/Components/Imagelist.vue
+++ b/assets/js/app/editor/Components/Imagelist.vue
@@ -1,79 +1,78 @@
 <template>
-    <div>
-       <div v-for="(child, index) in images" v-bind:key="index">
-            <editor-image
-                          :filename=child.filename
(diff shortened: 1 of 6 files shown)


I-Valchev:
This works.
Vue warns not to change props directly as it will be wiped on re-render (but for now this is what we want anyway). Maybe this can be refactored once we know more about Vue or in a subsequent Vue version.

assets/js/app/editor/Components/Imagelist.vue
```diff
@@ -0,0 +1,78 @@
...
+  </div>
+</template>
+
+<script>
+import Image from './Image';
+
+export default {
+  name: 'EditorImage',
+  components: { 'editor-image': Image },
+  //mixins: [field],
```
bobdenotter: Commented line can be removed. ;-)

PR #727, opened 2019-11-18
Feature/move images in imagelist

I-Valchev:


I-Valchev:
Not ready to merge yet.
Known issues:

- attributes not carried on move up/down
- Move empty image odd number of times, then move empty with non-empty: this triggers a bug in which both empty images are moved

PR #738, opened 2019-11-22
Makefile: fixup composer Memory limit errors

ttys3:
run `make start` result in error:
```
PHP Fatal error:  Allowed memory size of 1610612736 bytes exhausted (tried to allocate 20480 bytes) in phar:///home/my-username/bin/composer.phar/src/Composer/Repository/ComposerRepository.php on line 339

Fatal error: Allowed memory size of 1610612736 bytes exhausted (tried to allocate 20480 bytes) in phar:///home/my-username/bin/composer.phar/src/Composer/Repository/ComposerRepository.php on line 339

Check https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors for more info on how to handle out of memory errors.make[1]: *** [Makefile:21: install] Error 255
make[1]: Leaving directory '/home/my-username/repo/php/bolt'
make: *** [Makefile:15: start] Error 2

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 DC_RUN ?= docker-compose run --rm
 
+COMPOSER ?= COMPOSER_MEMORY_LIMIT=-1 composer
+
 .PHONY: help install server server-stop cache csclear cscheck csfix csfix-tests stancheck test \
 behat full-test db-create db-update db-reset docker-install docker-install-deps docker-start docker-assets-serve \
 docker-update docker-cache docker-csclear docker-cscheck docker-csfix docker-stancheck docker-db-create docker-db-reset \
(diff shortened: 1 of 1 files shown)


bobdenotter:
Hi @ttys3, 

Thanks for the fix, seems useful.. One question: should we really include the `-vvv` flag by default?

bobdenotter:
btw: failing test wil be addressed in #737

ttys3:
@bobdenotter  OK, I removed the `-vvv` flag
In my place, accessing to packgist is very slow, and sometimes I can't even download packages successfully.

ttys3:
thanks for your work
this project is really great.

bobdenotter:
@ttys3 Thanks! 

Maybe you know this already, but what I sometimes do on a slow connection is this: 

`composer update --dry-run` -> It downloads all packages that need to be updated, without applying them. Then if you run `composer update` again, it'll run much faster, because the files are already cached.

ttys3:
thanks for the tip,
my problem is network problem, so even use the `--dry-run` parameter, still I have to add an `-vvv` to see if there's any network failure

PR #748, opened 2019-11-24
Improve fieldtype factory creation

### LINKED ISSUE #743: If fields are removed from the contenttype, but are still in the database it breaks.
If anything, the factory method for fields should have a sensible default definition

bobdenotter:
Fixes #743


bobdenotter:
@I-Valchev This should fix the issue with the missing `localize` attribute.

PR #777, opened 2019-12-06
Fix jumping dropdowns

### LINKED ISSUE #775: Jumping dropdowns
While testing Bolt 4 i saw that the the dropdowns has a different location/offset if you hover or click on the "label".

dropdown2
dropdown1

I-Valchev:
Fixes #775 
There seem to be some [positioning issues](https://github.com/popperjs/popper.js/issues) with popper.js (the library Bootstrap uses to dynamically align dropdowns), so this disables dynamic positioning.


UtechtDustin:
You have to do the same on line 12 for the "Bolt"-Dropdown (visit bolt and Bolt Documentation).

I-Valchev:
Good catch, thanks!

PR #787, opened 2019-12-12
Feature/collections support all fields

### LINKED ISSUE #767: Improve collections
- [x] Add move up/down buttons to re-arrange collection items
- [x] Add support for 'simple' fields (fields that are not sets)

I-Valchev:
Fixes #767 
All fields are supported in collections.
@bobdenotter are you happy with the way they collections are displayed? Do you think the boundaries of a collection item are clear enough? 
For reference:
Screenshot 2019-12-12 at 09 56 13

### DIFF
--- a/assets/js/app/editor/Components/Collection.vue
+++ b/assets/js/app/editor/Components/Collection.vue
@@ -6,14 +6,18 @@
 
     <div class="d-flex">
       <editor-select
-        class="flex-grow-1 mr-2"
         ref="templateSelect"
+        class="flex-grow-1 mr-2"
         :value="initialSelectValue"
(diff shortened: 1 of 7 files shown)


codmdu:
The buttons to move down/up and to delete are very big in my opinion, it would make more sense to use only Icons with the title attribute and move them to the top right of each field.   
Also the Buttons of the departments field has no margin-top.

UtechtDustin:
The buttons to move down/up and to delete are very big in my opinion, it would make more sense to use only Icons with the title attribute and move them to the top right of each field.
Also the Buttons of the departments field has no margin-top.

PR #799, opened 2019-12-14
Don't break on existing records without sets.

bobdenotter:


bobdenotter:
@I-Valchev This seems like a harmless fix, but you can judge it better if i don't inadvertently break stuff with this.

I-Valchev:
@bobdenotter Looks good to me 👍

PR #872, opened 2020-01-13
Update `findOneByFieldValue` method to handle translated fields

bobdenotter:


bobdenotter:
Should pass after #874 is merged in.

PR #878, opened 2020-01-15
Log message does not log sensitive request info

### LINKED ISSUE #870: Log message (for example, when logging in) should _NOT_ log potentially sensitive info.
For example, sessions or POSTed data from the "Login" screen

I-Valchev:
Fixes #870 

1 line 👀


I-Valchev:
Oh! Much better, yes.

PR #898, opened 2020-01-21
Magic call returns field

### LINKED ISSUE #897: Magic `__call` for content doesn't work as expected
It used to be that `{{ record.title }}` output the "plain" value of a field, while `{{ dump(record.title) }}` dumped the underlying Entity. Somewhere in the past few weeks, this behaviour changed. 

Schermafbeelding_2020-01-21_om_12_21_18

I-Valchev:
Fixes #897


bobdenotter:
After discussion: let's close this, because it's not really feasible to fix.

PR #905, opened 2020-01-22
Rebuild our `services_bolt.yaml` less often

bobdenotter:

PR #912, opened 2020-01-24
Don't render widgets more than once

bobdenotter:


bobdenotter:
replaces part of #905

PR #994, opened 2020-02-03
Feature/required option

### LINKED ISSUE #930: Add required option to fieldtypes

I-Valchev:
Fixes #930 
Known issues: #995

### DIFF
--- a/assets/js/app/common.js
+++ b/assets/js/app/common.js
@@ -77,4 +77,12 @@ $(document).ready(function() {
       return false;
     });
   });
+
+  /*
+   ** Simulates disabled behavior for elements with data-readonly attribute.
+   * This is needed, because a disabled input cannot be required.
+   */
+  $("[data-readonly]").on('keydown paste', function(e){
+    e.preventDefault();
+  });
 });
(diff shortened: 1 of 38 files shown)


I-Valchev:
@bobdenotter  I think this can be merged now.

`List-type` fields (Collection, Imagelist, Filelist) cannot be 'required', because a required option does not make sense semantically for these. I'll update the documentation to reflect this.
In future, we may look at an option similar to "limit" for having a minimum number of items in those.
HTML and Markdown is a known issue, I've upvoted/posted issues on the relevant repos about it - although simplemde that Bolt uses for Markdown hasn't been updated in 4 years... anyway, this better be revisited on its own.

bobdenotter:
> List-type fields (Collection, Imagelist, Filelist) cannot be 'required', because a required option does not make sense semantically for these. I'll update the documentation to reflect this.

Thing is, we never had this for v3, either, and it wasn't a real problem. If we document it, i think we should be fine! 

> simplemde that Bolt uses for Markdown hasn't been updated in 4 years... anyway, this better be revisited on its own.

Yeah.. I usually like to check if something is kinda active, before we commit to it, but I missed out on that. Instead I looked at the lib that adds the vue wrapper. That's more active: https://github.com/F-loat/vue-simplemde/commits/master

Anyhow, maybe we should replace it with this, for example: https://ui.toast.com/tui-editor/

PR #1060, opened 2020-02-13
 fix dockerfile argument flag and extra env $PATH

flug:


bobdenotter:
@flug Cool, thanks! I assume we should make a similar change in the Project file, right? 

https://github.com/bolt/project/blob/master/docker/php/Dockerfile

flug:
@bobdenotter Yes, I made a pr for that, too.

bobdenotter:
@flug 🤦‍♂ Ok, i definetely need more coffee. Thanks!

flug:
Me too 😂

PR #1069, opened 2020-02-14
Always show slug in listing title, even if translations are available

I-Valchev:
Since slugs are localisable, pages with translatable slugs showed "Object object" for the hover on the listing page. This fixes it

### DIFF
--- a/assets/js/app/listing/Components/Table/Row/index.vue
+++ b/assets/js/app/listing/Components/Table/Row/index.vue
@@ -92,7 +92,6 @@ export default {
       ];
     },
     size() {
-      console.log(this.record);
       return this.$store.getters['general/getRowSize'];
     },
     sorting() {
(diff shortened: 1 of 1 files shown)


assets/js/app/listing/Components/Table/Row/index.vue
```diff
@@ -82,7 +82,17 @@ export default {
...
+      if (typeof this.record.fieldValues.slug === 'string') {
+        return this.record.fieldValues.slug;
+      }
+      // if slug has different locales, return the 0st one
+      return this.record.fieldValues.slug[
+        Object.keys(this.record.fieldValues.slug)[0]
+      ];
+    },
     size() {
+      console.log(this.record);
```
bobdenotter: ಠ_ಠ 

Doesn't ESlint trip on that?

I-Valchev: Oops! Seems like it doesn't... I'll look up if there's an option for it though!

PR #1104, opened 2020-02-21
ImageField uses injected RequestStack instead of request creating from globals

I-Valchev:


I-Valchev:
This doesn't work because:
When instantiating entities on the fly, the postLoad is not triggered. This is probably why the API doesn't work with it.

PR #1111, opened 2020-02-22
Allow configuration of web root folder

### LINKED ISSUE #1109: Option to change the 'public' folder
In Bolt 3 it's possible to add a bolt.yml file with a rule to use another foldername for Bolt's public accessible files, aka 'public'. Like this:

```
paths:
    web: '%site%/public_html'
```
How does one do this in Bolt 4?

bobdenotter:
Fixes #1109

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -34,19 +34,12 @@ DATABASE_URL=sqlite:///%kernel.project_dir%/var/data/bolt.sqlite
 ###< doctrine/doctrine-bundle ###
 
 ###> symfony/mailer ###
-# MAILER_DSN=smtp://localhost
+MAILER_DSN=smtp://localhost
 ###< symfony/mailer ###
 
 ###> nelmio/cors-bundle ###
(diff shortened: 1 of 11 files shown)


peterboorsma:
Thank you. I looked into the changed files in this commit but could not find an example of how and where to add the path to $publicFolder. In the .env file or config.yaml?

(The update to rc 5 broke my test installation.)

Screenshot 2020-02-23 at 10 24 24

bobdenotter:
Instructions for that can be found here: https://github.com/bolt/project/blob/master/UPDATE.md 

There is nothing to configure for the path. It looks for `public`, `public_html` and a few others, and pick whatever's present.

bobdenotter:
Although in your case, you might need to change it in a few other places. I didn't expect you to try it, before i had a chance to update the docs.

peterboorsma:
Thanks. I think you are right. Might take a few other changes. After adding the ' $publicFolder: '%bolt.public_folder%'' line Bolt breaks on something else. 
I'll just wait for the docs to be updated.

PR #1162, opened 2020-03-06
Make sure Imagelist and Filelist API values work on all endpoints

I-Valchev:
Ensures Imagelist and Filelist values work on `contents/1.json`, where the getApiValue on the images and files within the list is not called automatically by api-platform.


I-Valchev:
@nestordedios thank you!

PR #1166, opened 2020-03-09
Set default locale for existing fields

I-Valchev:
All existing fields have the default locale as the first locale from the content definition.
Fixes issue with 'Untitled' record when 'en' is not a locale.


I-Valchev:
🤷‍♂️ haha! thanks!

PR #1167, opened 2020-03-09
[WIP] current locale is initiated from defaultlocale

I-Valchev:
This is needed as well to ensure the current locale is also the default locale (KNP already sets 'en' by that time, so we need to override it)


bobdenotter:
As briefly discussed.. Let's park this one for now.

I-Valchev:
We can close this now.

PR #1180, opened 2020-03-14
Update Sirius Uploader, get rid of raw `$_FILES`

### LINKED ISSUE #402: Refactor "uploader" file handling
We currently use https://github.com/siriusphp/upload under the hood to handle uploads, but it requires us to pass in `$_FILES` instead of `$request->files->all()`. We should investigate if we want to remedy this. 

Three options: 

 - We fix sirius/upload, and upstream the fix
 - We use SF's upload mechanisms
 - We use VichUploaderBundle

bobdenotter:
Fixes #402


bobdenotter:
Merge after https://github.com/siriusphp/upload/pull/33 is merged in.

PR #1191, opened 2020-03-17
Don't break when trying to write thumbnails to non-writable folder

bobdenotter:


bobdenotter:
For now, we exclude PHPStan 0.12.15, because of this issue: https://github.com/phpstan/phpstan/issues/3090

PR #1195, opened 2020-03-17
Sorting correctly for MySQL

### LINKED ISSUE #1190: Unexpected results on Contenttype sorting
**Backend:**

- Sorting contenttypes on defined fields (eg `sort: title`) shows unexpected results. it is not apparent what sort is being based on.

- Sorting contenttypes using columns in Bolt_content table (eg `sort: id`, `sort: -createdAt`) works and gives correct results. 

**Frontend:**

- Listing.twig  - sort defaults to `sort: id` regardless of what is set in the `contenttype.yaml`.

- `{% setcontent entries = 'entries' orderby 'title' %}` also does not show correctly sorted results. This replicates the problems encountered at the backend. 

Details
-------

|----------------- | ------
| Relevant Bolt Version |  4.0.0 rc 6 / rc 7
| PHP version                 |  7.4 
| For UX/UI issues        | n/a

Reproduction
------------

### Bug summary

As above

### Specifics

screen shot 1: incorrect sort order using defined field
screenshot_1

screen shot 2: correct sort order using id, created date etc
screenshot 2

### Expected result

Ability to sort results based on defined fields of contenttype.

### LINKED ISSUE #1192: Sorting config does _NOT_ sort on frontend
The `sort` contenttypes.yaml config does not sort the records in the frontend, but it should.

bobdenotter:
Fixes #1190 
Fixes #1192

### DIFF
--- a/src/Controller/Frontend/ListingController.php
+++ b/src/Controller/Frontend/ListingController.php
@@ -44,8 +44,9 @@ class ListingController extends TwigAwareController implements FrontendZoneInter
         $contentType = ContentType::factory($contentTypeSlug, $this->config->get('contenttypes'));
         $page = (int) $request->query->get('page', 1);
         $amountPerPage = $contentType->get('listing_records');
+        $order = $contentType->get('sort');
 
-        $records = $this->query->getContent($contentTypeSlug, ['status' => 'published'])
+        $records = $this->query->getContent($contentTypeSlug, ['status' => 'published', 'order' => $order])
             ->setMaxPerPage($amountPerPage)
             ->setCurrentPage($page);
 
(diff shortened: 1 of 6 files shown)


src/Controller/Frontend/ListingController.php
```diff
@@ -44,8 +44,12 @@ public function listing(ContentRepository $contentRepository, Request $request,
         $contentType = ContentType::factory($contentTypeSlug, $this->config->get('contenttypes'));
         $page = (int) $request->query->get('page', 1);
         $amountPerPage = $contentType->get('listing_records');
+        $order = $contentType->get('sort');
```
JTNMW: $order = $request->query->get('order', $contentType->get('sort'));

JTNMW:
check if 'order' is set in the request first and if not default back to 'sort' config. 
$order = $request->query->get('order', $contentType->get('sort'));

Im not sure if this works correctly across the entire app, but this would be a preferred option if you can build this in.

bobdenotter:
> check if 'order' is set in the request first and if not default back to 'sort' config.
> $order = $request->query->get('order', $contentType->get('sort'));

I've added this in the PR. Good suggestion! 👍

PR #1223, opened 2020-03-25
Imagelist and Filelist work in API if not localized

I-Valchev:
Earlier I fixed the API values of Imagelists and Filelists, but it did not work for the ones that are not localized. This fixes that.


I-Valchev:
@bobdenotter  I have no idea why CI didn't trigger?

PR #1244, opened 2020-03-30
Embed field can be refreshed and cleared

### LINKED ISSUE #1243: Embed field cannot be cleared
The embed field cannot be cleared.

Removing the url (even after saving) keeps the other options as they were before.

I-Valchev:
Fixes #1243

### DIFF
--- a/assets/js/app/editor/Components/Embed.vue
+++ b/assets/js/app/editor/Components/Embed.vue
@@ -8,7 +8,7 @@
             <label for="embed-url">{{ labels.content_url }}</label>
             <div class="input-group">
               <input
-                v-model="url"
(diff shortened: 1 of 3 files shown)


bobdenotter:
Tests on the new part seem to be failing: 

```
--- Failed steps:

001 Scenario: As an Admin I want to be able to make use of the embed field # tests/e2e/edit_record_1.feature:46

      And the "fields[embed][title]" field should contain ""               # tests/e2e/edit_record_1.feature:60

        The field "fields[embed][title]" value is "Silversun Pickups - Nightlight (Official Video)", but "" expected. (Behat\Mink\Exception\ExpectationException)
```

PR #1245, opened 2020-03-30
Access fields inside a set by their name

I-Valchev:
Pretty sure I saw an issue about it today, but I can't find it :/ 

Basically:

`{{ set.title}}` does not work before this PR.


technicallyerik:
I created an issue in the documentation project, because I thought I was missing something under documented, rather than seeing a bug- https://github.com/bolt/docs/issues/1025

I-Valchev:
@technicallyerik right, yes, thanks for that.

Turns out the documentation showed how the set field _should_ behave, not how it _did_ behave. This will fix it :-)

PR #1248, opened 2020-03-30
Collections and Sets in API

### LINKED ISSUE #1241: Collection field in API
Make sure collection field works as expected in API

I-Valchev:
As with Imagelists and Filelists, make sure to return the values of the fields, not the object. Also, for collections, include the `name`, `type`, and `value`, as collections have different fields within them.

This PR is based on:
#1246 (for collections)
#1245 (for sets)

Fixes #1241


I-Valchev:
Wohoo!

PR #1249, opened 2020-03-30
Fix KNP bundle location (which breaks the install)

I-Valchev:


bobdenotter:
This breaks because of https://github.com/KnpLabs/DoctrineBehaviors/pull/534

Issue #1256, opened 2020-04-01
Unnecessary table prefixing when multiple db connection defined


rixbeck:
In case when some has multiple database connections with different entity managers the actual table prefixer solution will prefix all configured database tables not just in default Bolt DB.
   
Details
-------
Bolt table prefix settings in `services.yml` is intended for prefixing Bolt's own tables to avoid collisions in case when default db may contains tables already for different purposes. Prefixing tables with Doctrine now goes by event listener `\Bolt\Doctrine\TablePrefix:loadClassMetadata()` which invoked on `loadClassMetadata` event. Unfortunately currently it is not checking the contextual entity class belongs to another entity manager than default, so tables in different databases will be also prefixed which leads wrong queries has been built.

In general it causes problems when some would like to use Bolt as a manager for contents for his SF based application that uses different DB connections.       

| Question                      | Answer
|----------------- | ------
| Relevant Bolt Version | 4.0 / 4.1 / master
| Install type                   |  Composer install
| BC Break                     | no
| PHP version                 | unrelevant |

Reproduction
------------

I have the similar doctrine config like this:
```yaml
parameters:
    # Adds a fallback DATABASE_URL if the env var is not set. This allows you
    # to run cache:warmup even if your environment variables are not available
    # yet. You should not need to change this value.
    env(DATABASE_DRIVER): 'pdo_sqlite'
    env(DATABASE_URL): ''

doctrine:
# All your database connecctions used by the app should be defined here
    dbal:
        default_connection: default
        connections:
# Default Bolt Database connection
            default:
                charset: utf8mb4
                url: '%env(resolve:DATABASE_URL)%'
            
# Connection for a wireframe database which managed by a 3rd party application
            wireframe:
                driver:  pdo_mysql
# Connection url comes from env file doesn't necessary to describe here
                url: '%env(DATABASE_URL_WIREFRAME)%'
                charset: utf8
# Some of our tables has no primary fields needed to ORM so have to ignore them
# We are using non-standard db field types like ENUM so have to define how to cast a standard type if using reverse engineering mapping tables to entities
                mapping_types:
                    enum: string

# This is a db managed by our app, data here doesn't belongs either Bolt nor the 3rd party data 
            localdb:
                charset: utf8mb4
                url: '%env(resolve:DATABASE_URL_LOCALDB)%'

# ORM section
    orm:
        auto_generate_proxy_classes: '%kernel.debug%'
        default_entity_manager: default
        entity_managers:
            localdb:
[...]

PR #1265, opened 2020-04-02
Feature/em based table prefixing

rixbeck:
Possible solution for #1256

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -5,7 +5,8 @@ parameters:
     # This parameter defines the codes of the locales (languages) enabled in the application
     app_locales: en|nl|es|fr|de|pl|it|hu|pt_BR|ja|nb|nn|nl_NL|nl_BE|is|ru
     app.notifications.email_sender: anonymous@example.com
-    bolt.table_prefix: bolt_
+    app.table_prefix:
+        default: bolt_
     bolt.backend_url: /bolt
 
 services:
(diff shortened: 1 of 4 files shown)


config/services.yaml
```diff
@@ -21,7 +26,7 @@ services:
             $emailSender: '%app.notifications.email_sender%'
             $projectDir: '%kernel.project_dir%'
             $publicFolder: '%bolt.public_folder%'
-            $tablePrefix: '%bolt.table_prefix%'
+            $tablePrefixes: '%app.table_prefix%'
```
bobdenotter, line 29: If we could leave this as `$tablePrefix: '%app.table_prefix%'`, then we could probably make this change so it won't break for people who are updating.

src/Event/Subscriber/TimedPublishSubscriber.php
```diff
@@ -20,10 +23,22 @@ class TimedPublishSubscriber implements EventSubscriberInterface
...
     {
-        $this->entityManager = $entityManager;
-        $this->tablePrefix = Str::ensureEndsWith($tablePrefix, '_');
+        $this->entityManager = $managerRegistry->getManager('default');
+        $this->tablePrefix = $this
+            ->setTablePrefixes($tablePrefixes, $managerRegistry)
+            ->getTablePrefix($this->entityManager);
+    }
+
+    /**
```
bobdenotter, line 34: Did you move this method to this position on purpose?

rixbeck: That was by IDE formatter did, not me :)  ...reverted

PR #1267, opened 2020-04-02
Fix Preview by removing EntityManager flush

### LINKED ISSUE #1264: Create E2E test for making sure preview works as expected

I-Valchev:
We can avoid flushing by removing the fields from the EntityManager _and_ from the current content, which allows the Preview to work as expected.

And fixes #1264


I-Valchev:
@bobdenotter I had to clean up my own mess  😳

Issue #1296, opened 2020-04-14
Hide viewless content from API


technicallyerik:
Feature request / suggestion:  Viewless content should be hidden from the API.

A user might have either data submitted from Bolt Forms into a content type, or maybe a certain content type has content restricted to authenticated users via 3rd party code.

We can piggyback off the `viewless: false` parameter in content types to hide this content from the public API.

`curl -X GET "http://localhost/api/contents?contentType=viewless_content_type" -H "accept: application/ld+json"`

Details
-------

| Question                      | Answer
|----------------- | ------
| Relevant Bolt Version | 4.0-RC 15
| Install type                   |  Composer install
| PHP version                 | 7.3.9
| Web server                  | Apache

technicallyerik:
https://api-platform.com/docs/core/security/#filtering-collection-according-to-the-current-user-permissions

bobdenotter:
fixed by #1305

PR #1305, opened 2020-04-16
API shows published and viewless: false content only

### LINKED ISSUE #1295: Unpublished content accessible via front-end API
Unpublished content is accessible to unauthenticated users via the front-end API.

`curl -X GET "http://localhost/api/contents?status=draft" -H "accept: application/ld+json"`

A user might have a draft news release, product launch, embargoed product review, etc they might not want a crafty user accessing before it's public.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.0-RC 15
| PHP version                 | 7.3.9

Reproduction
------------

### Steps to reproduce

In your console, enter `curl -X GET "http://localhost/api/contents?status=draft" -H "accept: application/ld+json"`

(I assume the same is with the GraphQL endpoints)

### Expected result

Only content with a `published` status is accessible with or without a `status` parameter when accessing the API as an unauthenticated user.

### Actual result

Unpublished content can be accessed by unauthenticated users via the API.

I-Valchev:
API work is planned ahead for v4.3, however this seems important to fix now.

Fixes #1295 and #1296

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -90,6 +90,11 @@ services:
     Bolt\Twig\DumpExtension:
         arguments: ["%kernel.environment%"]
 
+    Bolt\Api\Extensions\ContentExtension:
+        tags:
+            - { name: api_platform.doctrine.orm.query_extension.collection }
+            - { name: api_platform.doctrine.orm.query_extension.item }
+
     Doctrine\ORM\Query\Expr: ~
 
     monolog.processor.request:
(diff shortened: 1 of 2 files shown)


technicallyerik:
Y'all are fast!

I tested these changes and the Unpublished filter looks good.

I wanted to get your thoughts on applying `$this->excludeViewlessContent($queryBuilder);` to `applyToCollection` as well, to make it consistent with how [viewless works on the front-end](https://docs.bolt.cm/4.0/howto/resource-contenttype#viewless).

I-Valchev:
Oops - this is what I intended originally. Nice spot, thanks it will get fixed :-)

PR #1314, opened 2020-04-20
Backend menu caches localized

I-Valchev:


bobdenotter:
Merging this in, and also created a followup issue: #1321

PR #1343, opened 2020-04-29
Require confirmation to delete collection item in editor

I-Valchev:


I-Valchev:
Hmm... rather than doing this (which will be messy anyway), it may be more useful to create a button "Restore saved version" instead, as this will also work for any sort of changes made? What this would do is just reload the page, without saving. @bobdenotter

PR #1384, opened 2020-05-18
Persist non-translatable field with correct default locale

I-Valchev:
So far all non-translatable fields were persisted with locale `en`.

In this PR, the fields are persisted with the provided `locale` from `services.yaml`


I-Valchev:
wow I'm shocked myself!

Issue #1437, opened 2020-06-03
Records locale issues


peterboorsma:
Bolt install updated to RC22 with 1 language (nl)

In config.yaml:` locale: nl`
No localizations in contenttype.yaml

The issues:
- The `{{ htmllang() }}` tag shows nothing in my frontend templates: `<html lang="">`

- When adding a new record, the slug remains empty. I have to 1) save the record, then 2) use "generate from: title" to add the correct slug. 
See screencast: https://www.dropbox.com/s/z9puc6i7pv1f7mp/Screen%20Recording%202020-05-31%20at%2022.14.41.mov?dl=0

- /nl/ appears in links to records in listing templates 

First I thought that it could be a caching thing, so I emptied cache in Bolt (Maintenance > Empty cache). No changes. The did a php ./bin/console cache:clear . No changes. Then I emptied the complete cache/dev directory.
Unfortunetaly the issues remain.

Remote system info:
```
 Bolt version: 4.0.0 RC 22
 
 * Install type: Composer install
 * Database: mysql 5.5.5-10.2.32-MariaDB - 127.0.0.1 via TCP/IP (with JSON)
 * PHP version: 7.3.18
 * Operating System: Linux - 2.6.32-042stab136.1
```

Local system info:
```
 Bolt version: 4.0.0 RC 22

 * Install type: Composer install
 * Database: mysql 5.7.26 - 127.0.0.1 via TCP/IP (with JSON)
 * PHP version: 7.3.12
 * Operating System: Darwin - 19.4.0
```

This is a new issue based on problems as described in https://github.com/bolt/core/issues/1398

I-Valchev:
hi @peterboorsma , can you try setting the `locale` option in `services.yaml` to `nl`, if it was different?

peterboorsma:
It was set to `'en'` in `services.yaml`. Changed it to `'nl'` but now all content 'dissapeared' from my records.
<img width="360" alt="Screenshot 2020-06-03 at 12 47 23" src="https://user-images.githubusercontent.com/2750653/83628201-6701ba80-a598-11ea-9133-b9da362d2651.png">

No luck with the {{ htmllang()}} either I'm afraid.

peterboorsma:
Update: if I save a new record (still have to add the slug manually) and go to my (frontend) listing page, all records in that listing now have url's without the /nl/ addition.

peterboorsma:
Just to see this is documented here: 
- Set the locale in services.yaml:` locale: 'nl'`
[...]

PR #1439, opened 2020-06-03
Fix htmllang() when no locales defined

I-Valchev:
Fixes htmllang part of #1437

### DIFF
--- a/src/Utils/LocaleHelper.php
+++ b/src/Utils/LocaleHelper.php
@@ -56,12 +56,12 @@ class LocaleHelper
             if ($all) {
                 $localeCodes = $this->localeCodes;
             } else {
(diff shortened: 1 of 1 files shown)


milosa:
My `htmllang()` is still empty in RC 23.

bobdenotter:
@milosa Does that install have `bobdenotter/configuration-notices` ? ;-)

milosa:
@bobdenotter Yes, it does. I tried it on a clean install.

peterboorsma:
I can confirm @milosa 's findings.
The `{{htmllang()}}` tag remains empty in RC23. In my case only on a site with 1 language and no locales in `contenttypes.yaml`.

PR #1445, opened 2020-06-05
[WIP] Unit tests for twig functions and filters

I-Valchev:


bobdenotter:
Tests pass, but looks like some branches got mixed up. Is the PR itself OK and good to go?

I-Valchev:
Oh, yeh I think this is the issue with GitHub not updating PR changes after rebase...

But there's too many things that should not be here. I'll write some more tests and submit another, cleaner PR :)

PR #1448, opened 2020-06-06
Fix for windows not handling `%k` in timestamp strings properly

### LINKED ISSUE #1447: strftime error on Windows
I tried to install Bolt on a Windows machine as a test, and used Laragon out of curiosity.
Everything seems to be working, except for an error 500 when editing an entry. This works perfect on my Linux machine though. 
It seems like the `%k` format for `strftime` isn't supported on my machine.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.0.0 RC 22
| Install type                   |   Composer install
| PHP version                 | 7.2  |
| Web server                  | Apache version 2.4.35
| For UX/UI issues        | Browser name and version

Reproduction
------------

### Bug summary

- TypeError in `Bolt\Twig\LocaleExtension::localedatetime()`: returns `false` instead of a string
- In fact it's the `strftime` that returns a false, so seems to be a Windows thing
  - > As the output is dependent upon the underlying C library, some conversion specifiers are not supported. On Windows, supplying unknown conversion specifiers will result in 5 E_WARNING messages and return FALSE.
  - see https://www.php.net/strftime
- The time string in https://github.com/bolt/core/blob/master/templates/_partials/fields/date.html.twig#L39 is resulting in this error
  - looks like to be the ´%k´ flag that doesn't work:  
```
λ php -a
Interactive shell

php > var_dump(strftime('%B %e, %Y - %k:%M<small>%p</small'));
bool(false)
php > var_dump(strftime('%B %e, %Y - %H:%M<small>%p</small'));
string(37) "June  6, 2020 - 13:11<small>PM</small"
php >
```
  - seems like  ´%k´ is used on other places as well

### Specifics

- OS: Windows, using Laracon
- Installed using composer
- Bolt version: 4.0.0 RC 22
- PHP version:  7.2.19

<details>
<summary>Screenshot of the error</summary>

afbeelding

</details>

Works perfectly on my Linux machine with Docker.  
Also checked on WSL, and works perfectly there as well.

### Steps to reproduce

- Happens whenever I change a contenttype in de admin (e.g. homepage)

### Expected result

No exception when trying to edit a contenttype

### Actual result

Exception when trying to edit a contenttype

bobdenotter:
Fixes #1447

### DIFF
--- a/templates/_macro/_macro.html.twig
+++ b/templates/_macro/_macro.html.twig
@@ -74,3 +74,10 @@
     {% endfor %}
     {{ fieldsHtml|json_encode }}
 {% endapply %}{% endmacro %}
+
+
+{% macro relative_datetime(datetime) %}{% apply spaceless %}
+    <abbr class="datetime-relative" title="{{ datetime|localedatetime('%A %B %e, %Y - %k:%M%p') }}">
+        {{- datetime|date('c') -}}
+    </abbr>
+{% endapply %}{% endmacro %}
--- a/templates/_partials/fields/date.html.twig
+++ b/templates/_partials/fields/date.html.twig
@@ -1,4 +1,5 @@
 {% extends '@bolt/_partials/fields/_base.html.twig' %}
+{% import '@bolt/_macro/_macro.html.twig' as macro %}
 
 {% block field %}
   {# check disabled #}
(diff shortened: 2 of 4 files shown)


bobdenotter:
@jnstr Yeah, It was a pretty easy fix, so.. Might as well get it out of the way! 

Thanks for bringing it to my attention! 👍

Issue #1182, opened 2020-03-14
Update to Symfony 5


bobdenotter:
Currently, Bolt 4 uses SF 4 components. We should bump these to v5. 

- [x] Replace Behat / MinkBrowserKitDriver with updated forks -> See #1474
- [x] Replace PagerFanta with `BabDev/BabDevPagerfantaBundle`
- [x] Remove `doctrine/doctrine-cache-bundle`
- [x] Fix deprecations in SF 4.4
- [x] use `error_controller` under "framework", instead of `exception_controller` under "twig" -> See #1477

bobdenotter:
Behat / MinkBrowserKitDriver needs to be updated, but it seems work is underway: 

https://github.com/minkphp/MinkBrowserKitDriver/pull/142

bobdenotter:
Same for PagerFanta: 

- https://github.com/whiteoctober/WhiteOctoberPagerfantaBundle/issues/224
- Fork here: https://github.com/BabDev/BabDevPagerfantaBundle

bobdenotter:
Other findings: 

We need to fix deprecations: 

```
!!
!!  Fatal error: Uncaught TypeError: Argument 2 passed to Symfony\Component\DependencyInjection\ContainerBuilder::set() must be an object or null, array given, called in /foo/src/Kernel.php on line 95 and defined in /foo/vendor/symfony/dependency-injection/ContainerBuilder.php:484
!!  Stack trace:
!!  #0 /foo/src/Kernel.php(95): Symfony\Component\DependencyInjection\ContainerBuilder->set('bolt.config.gen...', Array)
!!  #1 /foo/src/Kernel.php(64): Bolt\Kernel->setBoltParameters(Object(Symfony\Component\DependencyInjection\ContainerBuilder), '/Users/bob/Site...')
!!  #2 /foo/vendor/symfony/framework-bundle/Kernel/MicroKernelTrait.php(118): Bolt\Kernel->configureContainer(Object(Symfony\Component\DependencyInjection\ContainerBuilder), Object(Symfony\Component\Config\Loader\DelegatingLoader))
!!  #3 /foo/vendor/symfony/dependency-injection/Loader/ClosureLoader.php(38): Bolt\Kernel->Symfony\Bundle\FrameworkBundle\Kernel in /foo/vendor/symfony/dependency-injection/ContainerBuilder.php on line 484
[...]

PR #1477, opened 2020-06-14
Use new Symfony Error Controller, instead of Twig's old one.

### LINKED ISSUE #1476: Exceptions handled by Bolt itself, should show the Exception.
Exceptions handled by Bolt (currently `404`'s and `503`'s (for maintenance)) have the actual exception available in twig: 

```php
$twig->addGlobal('exception', $exception);
```

We should use these in the templates: 

```twig
{{ exception.message }}
{{ dump(exception) }}
```

bobdenotter:
Fixes #1476

Part of #1182

### DIFF
--- a/config/packages/framework.yaml
+++ b/config/packages/framework.yaml
@@ -29,3 +29,6 @@ framework:
     validation:
         email_validation_mode: 'html5'
         enable_annotations: true
+
+    # Override Symfony's error controller, so we can show custom 404's and 503's
+    error_controller: Bolt\Controller\ErrorController::showAction
--- a/config/packages/twig.yaml
+++ b/config/packages/twig.yaml
@@ -12,5 +12,6 @@ twig:
     globals:
         'config': '@Bolt\Configuration\Config'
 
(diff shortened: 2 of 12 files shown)


peterboorsma:
I think this commit causes problems when updating to RC25.
Frontend gives a Internal Server Error. Log says (no error message in my browser screen):

{"exception":"[object] (ErrorException(code: 0): User Deprecated: The \"Symfony\\Component\\Debug\\FatalErrorHandler\\ClassNotFoundFatalErrorHandler\" class is deprecated since Symfony 4.4, use \"Symfony\\Component\\ErrorHandler\\FatalErrorHandler\\ClassNotFoundFatalErrorHandler\" instead. at /Users/xxxx/localdev/xxxx/vendor/symfony/debug/FatalErrorHandler/ClassNotFoundFatalErrorHandler.php:20)"} []

bobdenotter:
The solution should be the top item in this file: https://github.com/bolt/project/blob/master/UPDATE.md 

After making these changes, run `bin/console cache:clear`, and you should be up and running again.

PR #1484, opened 2020-06-15
Add ability to override routing with record_route

### LINKED ISSUE #1470: Specify route name in ContentType config
Be able to specify the route name that should be used in `contenttypes.yaml`, so that the `|link` will work.

### LINKED ISSUE #1483: Overriding routes.yaml breaks |link unexpectedly
See Canonical.php @ line 213.

Also see #1470

I-Valchev:
Fixes #1483 
Fixes #1470

One outstanding issue:

```
termsandconditions:
  path: /terms-and-conditions
  defaults:
    _controller: Bolt\Controller\Frontend\DetailController::record
    contentTypeSlug: 'terms-and-conditions'
    slugOrId: 31
```

The default slugOrId is set here.
Trying to get the `|link` on the `/terms-and-conditions` page breaks the url, because the ContentExtension::getLink always overrides the slugOrId param, hence the link becomes:
`/terms-and-conditions?slugOrId=someUnexpectedSlug`

where someUnexpectedSlug is the slug from the content.

I'm not sure how to fix this elegantly? @bobdenotter

### DIFF
--- a/src/Canonical.php
+++ b/src/Canonical.php
@@ -9,6 +9,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\Routing\Exception\InvalidParameterException;
 use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 class Canonical
@@ -170,7 +171,14 @@ class Canonical
     {
         if (isset($params['_locale']) && $params['_locale'] === $this->defaultLocale) {
             unset($params['_locale']);
-            $route = str_replace('_locale', '', $route);
+            $routeWithoutLocale = str_replace('_locale', '', $route);
+
+            // If a route without the locale exists, use that. e.g. record_locale -> record
+            try {
+                $this->generateLink($routeWithoutLocale, $params);
+                $route = $routeWithoutLocale;
+            } catch (RouteNotFoundException $e) {
+            }
         }
 
         return $this->urlGenerator->generate(
--- a/src/Canonical.php
+++ b/src/Canonical.php
@@ -170,6 +170,7 @@ class Canonical
     {
         if (isset($params['_locale']) && $params['_locale'] === $this->defaultLocale) {
             unset($params['_locale']);
+            $route = str_replace('_locale', '', $route);
         }
 
         return $this->urlGenerator->generate(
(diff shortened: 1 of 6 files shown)


bobdenotter:
I see the problem! 

First, it works for the generated canonical just fine, it seems: 

Screenshot_2020-06-16_at_08_02_11

But, it doesn't work correctly when using `{{ record|link }}`: 
 
Screenshot 2020-06-16 at 08 11 07

This is because it's _configured_ with some defaults, and then you generate a route for something that has different values for these specific things. This is undesirable, but it's also the _correct_ behaviour, I think. 

If i'd set those values as defaults (and make sure there's an `entry` with slug `terms-and-conditions`), it doesn't add them: 

```yaml
termsandconditions:
  path: /terms-and-conditions
  defaults:
    _controller: Bolt\Controller\Frontend\DetailController::record
    contentTypeSlug: 'entry'
    slugOrId: terms-and-conditions
```

Screenshot 2020-06-16 at 08 15 58

I think this is not something we can fix, but it does have a workaround.. 

What's more annoying is the following: 

```
content_seo:
   path: /{slugOrId}
   methods: [GET|POST]
   defaults:
     _controller: Bolt\Controller\Frontend\DetailController::record
```

That'll make any `/lorem-ipsum` work, but if i add `record_route: content_seo`, the result of `{{ record|link }}` is `/`.. There it _doesn't_ substitute the `slugOrId`, while it really should do that.

I-Valchev:
I see now!
Yeah that makes sense, if it's configured in `routes.yaml` with the slug and not ID, they are the same, so no query params.

I am not sure I could replicate the `content_seo` example. You say that any `/lorel-ipsum` should work, but for me when I try anything `site.wip/lorem-ipsum` it says content not found. Which I think makes sense given the controller:

```
        if (is_numeric($slugOrId)) { // THIS IS FALSE
            $record = $this->contentRepository->findOneBy(['id' => (int) $slugOrId]);
        } else {
           // $contentTypeSlug is null here?
            $contentType = ContentType::factory($contentTypeSlug, $this->config->get('contenttypes'));
            $record = $this->contentRepository->findOneBySlug($slugOrId, $contentType);
        }

        return $this->renderSingle($record, $requirePublished);
```

I was able to get it working if I change `ContentRepository::findOneBySlug` on line 146 from `if ($contentType) {` to 
`if ($contentType && $contentType->get('slug')) {`

And then the `|link` is `/aut-error-cupiditate-magnam?contentTypeSlug=blog-post`

bobdenotter:
> And then the |link is /aut-error-cupiditate-magnam?contentTypeSlug=blog-post

Still, that's better than it was before. If you fix the tests (seems pretty straightforward in this case), let's merge and move to the next item on the list for now! :-)

PR #1497, opened 2020-06-17
Fix: Don't show spinners on invalid form submissions

bobdenotter:


assets/js/app/patience-is-a-virtue.js
```diff
@@ -18,7 +26,7 @@ $('*[data-patience]').on('click', function() {
 });
 
 window.reEnablePatientButtons = function() {
-  $('*[data-patience]').each(function() {
+  $('*[data-original-class]').each(function() {
```
I-Valchev, line 29: what is data-original-class?

bobdenotter: It gets set on line 20, when a button (or something else) gets a spinner.

PR #1518, opened 2020-06-19
Cleanup usage of `$request` in Controllers (since TwigAwareController already has it)

bobdenotter:


I-Valchev:
Tests are failing because, theoretically speaking, people can enter '  ' (two spaces) as display name to by-pass the required field. But we explicitly check for that in the backend, so that Bolt still shows that as an issue. The new `getFromRequest` returns null by default.

PR #1521, opened 2020-06-19
Datefield uses global app to access user locale

### LINKED ISSUE #1515: 'Variable "user" does not exist' with Date Fields inside Collection
When using Date fields within a Collection field, the following RuntimeError takes place in the Admin Panel:

```Variable "user" does not exist.```

The line causing the error is `vendor/bolt/core/templates/_partials/fields/date.html.twig (line 33)`

My field definition is:

```
        price_list:
            type: collection
            fields:
                start_date:
                    type: date
                    label: Start Date
                    required: true
                end_date:
                    type: date
                    label: End Date
                    required: true
                price:
                    type: text
                    required: true
```

Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.0.0-rc.25
| Install type                   | Composer install
| PHP version                 | 7.3.17

I-Valchev:
Fixes #1515

### DIFF
--- a/templates/_partials/fields/date.html.twig
+++ b/templates/_partials/fields/date.html.twig
@@ -30,7 +30,7 @@
     :readonly='{{ readonly|json_encode }}'
     form='{{ form }}'
     mode='{{ mode }}'
-    locale='{{user.locale}}'
+    locale='{{app.user.locale}}'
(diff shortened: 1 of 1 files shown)


I-Valchev:
haha not really. user works just fine outside collections/sets, but inside it's not available because the context is limited :-)

PR #1530, opened 2020-06-22
Make sure currentlocale on edit is always set

### LINKED ISSUE #1528: View saved version button does not work on homepage or singletons
The homepage contenttype 'View saved version' button seems broken, either because it's the homepage or because it's singleton.

I-Valchev:
Fixes #1528


src/Controller/Backend/ContentEditController.php
```diff
@@ -512,6 +512,10 @@ private function getEditLocale(Request $request, Content $content): string
             $locale = $content->getDefaultLocale();
         }
 
+        if (! $locale) {
+            $locale = $this->defaultLocale;
```
bobdenotter, line 516: If we rebase this on master, line 508 says: 

```php
  $locale = $this->getFromRequest('edit_locale', '');
```

Couldn't we just do this? 

```php
  $locale = $this->getFromRequest('edit_locale', $this->defaultLocale);
```

Or would that conflict with the `$locales->contains($locale)` below it?

I-Valchev: It will conflict with the `$locales->contains($locale)`. This fix is meant for cases when the content locales are not defined. In that case, `$locales` will always be empty, but `$content->getDefaultLocale()` will return null. The idea is to avoid returning null when no locales are defined.

bobdenotter: Ok, in that case: let’s roll with this! :-)

I-Valchev:
@bobdenotter Also had to change `_localeswitcher.html.twig` since after this change `currentlocale` is always defined, hence the switcher is always shown. I'm a bit unsure about this though: a CT could exist with just one locale _that is different_ to the default one, but that won't be visible in the admin panel with this latest change.

I can think of 3 possible options:
- always show the locale switcher (even if just 1 option exists)
- never show the locale switcher when just 1 option exists
- show the locale switcher when one option exists _only if_ that option is different than the default site locale

What do you think?

Issue #1286, opened 2020-04-08
search/setcontent on Select fields with "multiple: true"


JTNMW:
When using search, setcontent or url query on taxonomy fields or select fields with 'multiple: true', with more than one entry eg  ["love", "Music", "Movies"] only the first values in these fields are searchable. 

search on "love" would return results, search on "music" or "Movies" returns no content.
This bug occurs at both frontend (set content, search, url query) and backend in filter on content listing.

In selectQuery.php changing
$newLeftExpression = JsonHelper::wrapJsonFunction($translationsAlias . '.value', null, $em->getConnection());

to 

$newLeftExpression = $translationsAlias . '.value';

resolves the issue, but causes issues errors elsewhere.

Details
-------

| Question                      | Answer
|----------------- | ------
| Relevant Bolt Version | legacy / 4.0 / 4.1 / master
| Install type                   |  Zip or tar install / Composer install / GitHub checkout
| BC Break                     | yes/no
| PHP version                 | 7.2 / 7.3 / 7.4 |
| Web server                  | Apache / Nginx / Built-in, version [ version ]
| For UX/UI issues        | Browser name and version

Reproduction
------------

If you're filing a bug, please describe how to reproduce it. Include as much
relevant information as possible, such as:

bobdenotter:
Hmm, right.. This is not trivial to fix, i think. Marked as "bug", so it'll stay on the list!

JTNMW:
ive managed find a fix for this issue, though im not sure if there might be more elegant solution:

**changes to selectQuery.php**

<img width="667" alt="Screenshot 2020-06-18 at 10 24 18" src="https://user-images.githubusercontent.com/57093734/85004247-410b1700-b14f-11ea-9ff5-e26fc74b1c70.png">

<img width="1001" alt="Screenshot 2020-06-18 at 10 29 52" src="https://user-images.githubusercontent.com/57093734/85005396-a4497900-b150-11ea-87cc-6f7794b8caf0.png">


**changes to config/packages/doctrine.yaml**

<img width="729" alt="Screenshot 2020-06-18 at 10 30 34" src="https://user-images.githubusercontent.com/57093734/85004253-41a3ad80-b14f-11ea-99e0-d2b8a45fe67c.png">

this resolves issues frontend when filtering and %setcontent% for select fields with multiple entries, and all taxonomies (tags, catagories) with multiple entries.

[...]

PR #1541, opened 2020-06-23
fix for issue #1286 search/setcontent on selectfields with multiple entries. 

JTNMW:


bobdenotter:
@JTNMW You can run `make csfix` on the command line, and it'll fix the Code Style issues for you. :-)

JTNMW:
small addition to selectQuery.php. Enables taxonomies to be included in contenttype filtering at the backend UI.

JTNMW:
sorry, apologies for any coding style errors if they are causing problems. Do i run csfix on command line whilst in app project dir. this what i get back. any advice?
make csfix
make csclear
mkdir -p var/cache/ecs
chmod -R a+rw var/cache/ecs
rm -rf var/cache/ecs/*
bash: vendor/bin/ecs: No such file or directory
make: *** [csfix] Error 127

bobdenotter:
You'll need to run this in the root folder of a checkout of `bolt/core`, instead of `bolt/project`. 

I've run this now, and pushed to your branch.. You'll have to `git pull` to fetch it. 

Let's see what the tests say now. :-)

JTNMW:
You may want to revert back to the first of the selectQuery.php files i submitted 50708be . The changes in 775ad38 is causing an issue at the frontend which i wasn't aware of when making the pull request. I'll see if i can get this working correctly.

bobdenotter:
Updated.. :-)

bobdenotter:
Ok, seems like there's some quirks or side-effects. I'll see what I can find.

JTNMW:
@bobdenotter it seems that json_contains() function is not available with sqlite. I'm not sure what the equivalent function is in sqlite. Is this causing the issue you are finding? I'll see if i can find a work around.

bobdenotter:
@JTNMW Hmm, could be that that's not readily available, yes.. For JSON_EXTRACT, we've implemented our own method to add it to the https://github.com/ScientaNL/DoctrineJsonFunctions library. 

It _should_ be supported now, though. See here: https://github.com/ScientaNL/DoctrineJsonFunctions/issues/49 

could it be that _your_ local SQLite version doesn't support it, yet? We've found out there's no real way to see if SQLITE supports JSON or not, other than guessing by a few known cases. See here: https://github.com/bolt/core/blob/master/src/Doctrine/Version.php#L16

JTNMW:
@bobdenotter Workaround so that multiselect fields can be searched without the need for JSON_CONTAIN(). Ive tested this on SQLite and MySQL and seems to work fine. Snippet of code added to allow taxonomies to be searched as part of the backend contenttype Filter (anyField).

JTNMW:
Sorry for error on my part, but I discovered that the foreach loop of taxonomies was causing memeory exhuastion. Ive removed the loop and search/setcontent with anyField now performs fine. ive done a new pr with the modification #1619.

PR #1565, opened 2020-06-28
Singletons return single Content result with `{% setcontent %}`

### LINKED ISSUE #1557: {% setcontent %} tag should return one object for singleton CTs
Currently, always a paginated thing is returned. But for singleton CTs, that doesn't make sense.

bobdenotter:
Fixes #1557


bobdenotter:
*wipes sweat from brow* 

Well, that was more work than I anticipated. 😅

PR #1568, opened 2020-06-29
Ensure fields inside sets and collections have their correct definition

### LINKED ISSUE #1555: Image extensions property not always working inside set inside collection
For some reason, the extensions setting of an image does not _always_ work when the image is inside a set that is inside a collection.

Example output:

```
@bolt/_partials/fields/image.html.twig line 23
"collections[content][picture_story][1][image]"
@bolt/_partials/fields/image.html.twig line 109
Bolt\Configuration\Content\ContentType {#265 ▼
  #items: array:5 [▼
    0 => "gif"
    1 => "jpg"
    2 => "jpeg"
    3 => "png"
    4 => "svg"
  ]
}
@bolt/_partials/fields/image.html.twig line 23
"collections[content][picture_story][2][image]"
@bolt/_partials/fields/image.html.twig line 109
[]
```

The 2nd one should not be empty. It should be the same as the first.

### LINKED ISSUE #1559: In the Editor, the name of the collection is always the yaml key, even if the name is explicitly set / overriden

I-Valchev:
Fixes #1559 by properly getting the definition for every field
Fixes #1555

### DIFF
--- a/src/Event/Listener/ContentFillListener.php
+++ b/src/Event/Listener/ContentFillListener.php
@@ -30,6 +30,14 @@ class ContentFillListener
         $this->users = $users;
     }
 
+    /**
+     * @deprecated
+     */
+    public function preUpdate(LifeCycleEventArgs $args): void
+    {
+        return;
+    }
+
     public function prePersist(LifecycleEventArgs $args): void
     {
         $entity = $args->getEntity();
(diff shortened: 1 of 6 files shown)


src/Event/Listener/ContentFillListener.php
```diff
@@ -32,25 +30,6 @@ public function __construct(Config $config, ContentExtension $contentExtension,
         $this->users = $users;
     }
 
-    public function preUpdate(LifecycleEventArgs $args): void
```
bobdenotter, line 35: Should we leave this method in place, as an empty placeholder, with an `@deprecated`? Then it won't break for people who haven't updated their `services.yaml` yet.

I-Valchev: yes, I'll put that back.

PR #1594, opened 2020-07-02
Make sure form validation works with `data-patience` buttons

### LINKED ISSUE #1584: Clicking 'Save changes' does not shown validation errors
When editing a content that has validation, e.g. a required field or `pattern` setting, the validation pop up does not show on pressing Save changes. It only shows if I try to save by hitting enter.

I-Valchev:
Fixes #1584


I-Valchev:
@bobdenotter quiiicklyyyyyy

PR #1598, opened 2020-07-03
Sorry GitHub Actions 😢 you are breaking too often

I-Valchev:


TomasVotruba:
Hi guys,

I'm looking at Github Actions and it seems they were intergrate, but rather annoying. 

Could you share what exactly is breaking but should not?

I-Valchev:
hi @TomasVotruba , for me it narrows down to 3 issues:

1. No option to restart checks (except for _all_ checks): that was an issue for us because the chromedriver/selenium/behat setup that we use doesn't always behave and sometimes we need to restart a build.

2. Lack of enough resources and good documentation: GA is still relatively new, and there just aren't the resources online when you bump into stuff. One example I remember was that the console output didn't have colour. The docs said colour _works_, and that's it.

3. Travis (what we currently use) has a lot of config options for common use cases that don't exist in GA. For example, getting chrome out of the box. When we last used GA, the Chrome version was sort of arbitrary, updated every now and then but not really consistently. We always run the tests on the latest stable chrome, so yet another reason this was inconvenient.

GA sort of promised to be _much_ faster but I didn't experience that. Yes it was a bit faster (20% perhaps?) but given 1, 2 and 3 that wasn't enough to offset the drawbacks. That's my own experience though 🤷‍♂️ and at the end of the day since both Travis and GA did what we wanted, it narrowed down to which of the two @bobdenotter and I liked better. Have anything to add @bobdenotter ?
_

bobdenotter:
That's a good summary, yes. 

I mentioned to @TomasVotruba on Reddit we've tried it, and weren't very satisfied with the results so far. But, that might've been for lack of experience with GA on our side. If it can be fixed up to work properly, we're very open to switching.

TomasVotruba:
Thanks for sharing your experience. My "first" was actually very similar. I tried to switch my full 100 % Travis to Github Actions in a day, but failed. It was not perfect and I reverted everything with frustration.

Thanks to a friend of mine, who tried it in simple 1 step way, I gave it a try a week later. Just one job that installed coding standards, nothing more. It worked, I merged it, CI was green :heavy_check_mark: 

That was the approach I should've done in the beggining. Why? Travis and Github Actions use conflicting keywords for similar meaning, e.g. "matrix", "step", "job", "section" etc. I mean... English and Spanish language are closer :D

Saying that, I'd suggest to start **with 1 small change at a time**. So you're in the control, feel confident, see the speed improvement and enjoy it. Because that's the point of it all, right? :)

If you're open to do it at small controlled steps, I'll try to prepare first PR.

What do you think?

TomasVotruba:
@I-Valchev I've noticed the Chrome out of the box. 

I have similar experience with monorepo-splits. Some advanced cases Travis just handles them much better. First I thought it had to all or nothing. Then I realized, it's actually to performance benefit to drop part of work on Travis and rest on Github Actions. 

So we have 23 jobs on Github Actions and the heaviest 2 on Travis. Instead of 6 minuts, it takes 3 on both :) 

image

<br>

**tl;dr;** We use the best tool for its job and gain 50 % faster CI feedback as side-effect.

PR #1645, opened 2020-07-22
Pass on Query parameters to subrequest from Homepage

bobdenotter:


bobdenotter:
> Wonder how you found out about it?

@I-Valchev I updated phpnews.io to the latest, and discovered the frontpage wasn't paginating anymore. 

Took longer than I care to admit what was causing the `page` parameter not too show up in the subrequest. 😅

PR #1654, opened 2020-07-28
fixed default homepage menu item

ymarkus:
Fixed homepage menu item as discussed in the slack channel from `path: homepage` to `path: /`

But I don't know if `path`/`link` are the same or if one is preferred.

### DIFF
--- a/config/bolt/menu.yaml
+++ b/config/bolt/menu.yaml
@@ -4,7 +4,7 @@
 main:
   - label: Home
     title: This is the <b>first<b> menu item.
-    link: homepage
+    link: /
     class: homepage
   - label: About
     link: blocks/about


bobdenotter:
Hi, 

In FrontendMenubuilder, there's this special case: 

```php
        // Special case for "Homepage"
        if ($trimmedLink === 'homepage' || $trimmedLink === $this->config->get('general/homepage')) {
            return ['Home', $this->urlGenerator->generate('homepage')];
        }
```

And, at first glance it seems to work as expected. Could you give me some more details again about what breaks on your end?

ymarkus:
OK, sorry for the sparseness. Here's the relevant excerpts from my files:
`_menu_main.yaml`:
```
<div class="collapse navbar-collapse float-right" id="navbarSupportedContent">
  <ul class="navbar-nav mr-auto">
    {% for item in menu %}
      <li class="nav-item {%- if item|current %} active{% endif %}">
        <a class="nav-link" href="{{ item.link }}">{{ item.label }}{%- if item|current %}<span class="sr-only"> (current)</span>{% endif %}</a>
      </li>
    {% endfor %}
  </ul>
</div>
```

`menu.yaml`:
```
main:
  - label: Home
    title: homepage
    path: homepage
```

`config.yaml`:
```
homepage: homepage
homepage_template: index.twig
```

The resulting menu item from the html: `<a class="nav-link" href="homepage">Home</a>`, which when clicked throws this exception: 
```
Error:
Call to a member function setMaxPerPage() on null

  at vendor/bolt/core/src/Controller/Frontend/ListingController.php:47
  at Bolt\Controller\Frontend\ListingController->listing(object(ContentRepository), 'homepage')
     (vendor/symfony/http-kernel/HttpKernel.php:157)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw(object(Request), 1)
     (vendor/symfony/http-kernel/HttpKernel.php:79)
  at Symfony\Component\HttpKernel\HttpKernel->handle(object(Request), 1, true)
     (vendor/symfony/http-kernel/Kernel.php:196)
  at Symfony\Component\HttpKernel\Kernel->handle(object(Request))
     (public/index.php:39)
```

bobdenotter:
Dumb question: Do you have a homepage? As in "A record in the `homepage` contenttype"?

ymarkus:
Here's my `contenttypes.yaml`:
```
homepage:
  name: Homepage
  singular_name: Homepage
  fields:
    [...]
  singleton: true
  icon_many: "fa:home"
  icon_one: "fa:home"
  locales: ['de', 'en']
[...]
```

And I have created a homepage which works find when I open it, only the menu item itself shows the wrong link. Seems really weird to me...

PR #1659, opened 2020-07-29
[WIP] Use `current` filter to check if menu item is this page

### LINKED ISSUE #1655: {{ item|current }} or {{ item.current }} not working
When using {{ item|current }} in my menu template, an exception is thrown. However, if I use {{ item.current }}, there is no exception thrown, but it always returns false.

Details
-------

|----------------- | ------
| Relevant Bolt Version |  4.0-rc32
| BC Break                     | maybe?
| PHP version                 | 7.3.20 |
| For UX/UI issues        | N/A 

Reproduction
------------

Use the following code in your template:
`menu.yaml`:
```
main:
  - label: Home
    link: /
[...]
```
`_menu_main.twig`:
```
<div class="collapse navbar-collapse" id="navbarSupportedContent">
  <ul class="navbar-nav mr-auto">
    {% for item in menu %}
      <li class="nav-item {%- if item.current %} active{% endif %}"> {# item|current throws an exception #}
        {{ dump(item) }}
        <a class="nav-link" href="{{ item.link }}">{{ item.label }}<span class="sr-only"> (current)</span></a>
      </li>
    {% endfor %}
  </ul>
</div>
```

For {{ item|current }} following exception is thrown:
```
Argument 2 passed to Bolt\Twig\ContentExtension::isCurrent() must be an instance of Bolt\Entity\Content, array given
[...]
at vendor/bolt/core/src/Twig/ContentExtension.php:311
at Bolt\Twig\ContentExtension->isCurrent(object(Environment), array('label' => 'Home', 'title' => '', 'link' => '/', 'class' => '', 'submenu' => null, 'uri' => '/', 'current' => false))
```

For {{ item.current }} it simply always returns false.

Therefore, I can't find a way to use the current filter in menus.

I-Valchev:
Fixes #1655 
Use `|current` and not `.current`

### DIFF
--- a/src/Twig/FrontendMenuExtension.php
+++ b/src/Twig/FrontendMenuExtension.php
@@ -61,6 +61,6 @@ class FrontendMenuExtension extends AbstractExtension
     {
         $uri = $item['uri'] ?? '';
 
-        return $uri === $this->request->getUri();
+        return $uri === $this->request->getRequestUri();
     }
 }
--- a/src/Twig/FrontendMenuExtension.php
+++ b/src/Twig/FrontendMenuExtension.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Bolt\Twig;
 
 use Bolt\Menu\FrontendMenuBuilderInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Twig\Environment;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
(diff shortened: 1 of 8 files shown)


src/Twig/FrontendMenuExtension.php
```diff
@@ -50,4 +56,11 @@ public function renderMenu(Environment $twig, ?string $name = null, string $temp
 
         return $twig->render($template, $context);
     }
+
+    public function isCurrent($item): bool
+    {
+        $uri = $item['uri'] ?? '';
+
+        return $uri === $this->request->getUri();
```
ymarkus, line 64: I think it should say `return $uri === $this->request->getRequestUri();` here, as `getUri()` returns the absolute URI and `getRequestUri()` returns the absolute path.

I-Valchev: Yes, you're absolutely right! Thanks for spotting this! :-)

ymarkus:
Tested with the current master branch and it works fine! Thanks @I-Valchev!

PR #1668, opened 2020-07-31
Add image.extension value

### LINKED ISSUE #1664: Thumbnailing SVG returns broken link
image

The listing template tries to thumbnail an SVG, which of course doesn't really work.

But, imho, if it renders the image at the requested URL, even if it is the original one, it will already avoid a lot of issues like that.

What do you think @bobdenotter ?

### LINKED ISSUE #1665: File fields should contain extension option
For example:

`{{ record|image.extension }}` => '.png`, or '.svg'

This would make it a lot easier to decide how to display certain files.

I-Valchev:
Fixes #1665 
Fixes #1664

### DIFF
--- a/src/Controller/ImageController.php
+++ b/src/Controller/ImageController.php
@@ -157,9 +157,9 @@ class ImageController
 
     private function isSvg(string $filename): bool
     {
-        $pathinfo = pathinfo($filename);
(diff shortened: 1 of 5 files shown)


bobdenotter:
Just in time for the weekend!

PR #1669, opened 2020-07-31
Make `getFieldClassname` public so extensions can check if a Field exists

bobdenotter:


bobdenotter:
Goes with https://github.com/bobdenotter/configuration-notices/pull/5

PR #1673, opened 2020-08-03
Child fields can render html by default, without |raw filter

### LINKED ISSUE #1558: HTML field not parsed when inside set/collection
The HTML field is not parsed when inside a set or collection.

Also, if the name of the field is `content`, then `{{ setfield.content }}` returns the Content object, not the field.

I-Valchev:
Fixes #1558

### DIFF
--- a/src/Entity/Field.php
+++ b/src/Entity/Field.php
@@ -90,8 +90,9 @@ class Field implements FieldInterface, TranslatableInterface
         $value = $this->getTwigValue();
 
         if (is_array($value) && array_key_exists($key, $value)) {
+            // If value is field, return getTwigValue so that {{ value }}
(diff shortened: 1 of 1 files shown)


src/Entity/Field.php
```diff
@@ -90,7 +90,9 @@ public function __call(string $key = '', array $arguments = [])
         $value = $this->getTwigValue();
 
         if (is_array($value) && array_key_exists($key, $value)) {
-            return $value[$key];
+            $value = $value[$key];
+
+            return $value instanceof self ? $value->getTwigValue() : $value;
```
bobdenotter, line 95: This line could do with a comment, i think. I see what it does, but _why_ does it do this?

PR #1694, opened 2020-08-07
Feature/orderby taxonomies

### LINKED ISSUE #1693: [RFC] Allow order by taxonomy
I don't know how difficult this might be to implement, but it would be nice to have.

When using `{% setcontent %}` it would be nice if you could order the result of the query by a Taxonomy.

To create a list of records ordered by Taxonomy you will need to do something like this:

1.    Get taxonomy options
2.    Iterate over Taxonomy options array
3.    Get all Records for the Taxonomy option in the loop
4.    Print Records as you iterate.

For example:

```
{% set regions = config.get('taxonomies/regions/options') %}
{% for region in regions %}
    <h2>{{ region }}</h2>
    {% setcontent locations = 'locations' where { regions: region } %}
    {% for location in locations  %}
        <h3>{{ location.title }}</h3>
    {% endfor %}
{% endfor %}

```

This would cost a lot of extra Database queries when fetching the records per taxonomy value. If something like this could be triggered from the Backend and just iterate over Records array it would be nicer.

I-Valchev:
Fixes #1693

### DIFF
--- a/public/theme/skeleton/custom/setcontent_1.twig
+++ b/public/theme/skeleton/custom/setcontent_1.twig
@@ -161,4 +161,17 @@
         </ul>
     </section>
 
+    <section id="eleven">
+        <h1>Eleven</h1>
+        {% setcontent entries = 'entries' order 'categories' %}
+        <ul>
+            {% for entry in entries %}
+                <li>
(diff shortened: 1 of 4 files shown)


nestordedios:
Niiiiiiice!! 🎉 👍

PR #1704, opened 2020-08-10
Allow for `variant: inline` for most fields

bobdenotter:
Well, that was easier than expected.


bobdenotter:
Schermafbeelding 2020-08-10 om 15 52 04

PR #1710, opened 2020-08-12
Enforce unique slugs

I-Valchev:


src/Event/Listener/ContentFillListener.php
```diff
@@ -72,4 +85,74 @@ private function guesstimateAuthor(): User
...
+            $slugField->setValue($newSlug);
+            $this->guaranteeUniqueSLug($content);
+        }
+    }
+
+    private function getSafeSlug(string $slug): string
+    {
+        $separator = '-';
+        // If it already ends with -{number}, increase it!
+        $dashNumber = '/' . $separator . "(\d)$/";
```
bobdenotter, line 147: I think this matches only one number.. suggestion: 

```suggestion
        $dashNumber = '/' . $separator . "(\d+)$/";
```

I-Valchev: Yess!

PR #1720, opened 2020-08-14
Use translated bootbox 

nestordedios:
This PR translates the bootbox modals.

'en' locale is used by default and if there is another locale set then that is the locale bootbox will use to show translated text in the modals.

More info here: http://bootboxjs.com/documentation.html#bb-function-setLocale

### DIFF
--- a/templates/_base/layout.html.twig
+++ b/templates/_base/layout.html.twig
@@ -1,7 +1,7 @@
 {# Check if aside has value and set hasAside variable #}
 {% set hasAside = block('aside')|trim is not empty ? true : false %}
 <!DOCTYPE html>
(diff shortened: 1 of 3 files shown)


templates/_base/layout.html.twig
```diff
@@ -1,7 +1,7 @@
 {# Check if aside has value and set hasAside variable #}
 {% set hasAside = block('aside')|trim is not empty ? true : false %}
 <!DOCTYPE html>
-<html {% if block('html_id') %}id="{% block html_id %}{% endblock %}"{% endif %}>
+<html lang={{ app.user.locale }} {% if block('html_id') %}id="{% block html_id %}{% endblock %}"{% endif %}>
```
bobdenotter, line 4:
```suggestion
<html lang="{{ app.user.locale }}" {% if block('html_id') %}id="{% block html_id %}{% endblock %}"{% endif %}>
```

PR #1732, opened 2020-08-19
Add `url_decode` Twig Filter

### LINKED ISSUE #1729: [RFC] Add url_decode filter
At this moment Twig allows to use the url_encode filter https://twig.symfony.com/doc/3.x/filters/url_encode.html,
but there is no way of decoding the encode url string.

Implementation of a Twig filter for this purpose could be handy.

bobdenotter:
Fixes #1729


I-Valchev (approved):
👍 green!

PR #1738, opened 2020-08-20
New `page` directive for setcontent to override the current page setting

I-Valchev:
```
{% setcontent records = 'entries' page 4 %}
```


bobdenotter (approved):
Very cool that this is working now! 👍

PR #1739, opened 2020-08-20
modification to SelectQuery.php related to #1619

JTNMW:
innerJoin replaced with leftJoin. "anyField" search/filter was only being performed on records which have Taxonomies assigned. LeftJoin resolves this problem,  searching/filtering records with "anyField" on all fields and taxonomies even if the record does not have any Taxonomies assigned.


I-Valchev (approved):
Ah, nice spot! Thanks!

Just waiting for the tests to pass before merging this right in. 👍

Issue #1757, opened 2020-08-23
"Script php bin/console extensions:configure handling the post-update-cmd event returned with error code 1"


bobdenotter:
Sometimes, when running `composer update`, I get the following: (text for searching below)

![Screenshot 2020-08-23 at 16 45 33](https://user-images.githubusercontent.com/1833361/90981263-3bbbb580-e560-11ea-8daf-ffb28af4a550.png)

It usually goes away the first time I re-run `composer update`

```bash
$ php composer.phar update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Restricting packages listed in "symfony/symfony" to "^5.1"

Prefetching 16 packages 🎶 💨
  - Downloading (100%)

Package operations: 0 installs, 16 updates, 0 removals
  - Updating composer/package-versions-deprecated (1.10.99 => 1.10.99.1): Loading from cache
  - Updating composer/ca-bundle (1.2.7 => 1.2.8): Loading from cache
  - Updating tightenco/collect (v7.19.1 => v7.25.0): Loading from cache
  - Updating nikic/php-parser (v4.8.0 => v4.9.0): Loading from cache
  - Updating doctrine/annotations (1.10.3 => 1.10.4): Loading from cache
  - Updating league/flysystem (1.1.0 => 1.1.3): Loading from cache
  - Updating knplabs/knp-menu (v3.1.1 => v3.1.2): Loading from cache
  - Updating symplify/package-builder (v8.2.1 => v8.2.11): Loading from cache
  - Updating ramsey/collection (1.0.1 => 1.1.0): Loading from cache
  - Updating brick/math (0.8.15 => 0.9.1): Loading from cache
  - Updating ramsey/uuid (4.1.0 => 4.1.1): Loading from cache
  - Updating seld/jsonlint (1.8.0 => 1.8.1): Loading from cache
  - Updating composer/xdebug-handler (1.4.2 => 1.4.3): Loading from cache
  - Updating bolt/core (4.0.0-rc.34 => 4.0.0-rc.35): Loading from cache
  - Updating bolt/assets (4.0.0-rc.34 => 4.0.0-rc.36): Loading from cache
  - Updating bobdenotter/configuration-notices (1.1.1 => 1.1.2): Loading from cache
Package zendframework/zendxml is abandoned, you should avoid using it. Use laminas/laminas-xml instead.
Writing lock file
Generating autoload files
drupol/composer-packages: Regenerating classes...
drupol/composer-packages: Done.
[...]

PR #1758, opened 2020-08-23
[WIP] Fail silently when a Field can't be constructed for some reason

bobdenotter:
This is a "Fix" for #1757 , but I don't like it. 

Putting it out there, so we don't forget and can keep an eye on it.


bobdenotter:
Haven't bunped into this for a while. Let's get rid of it!

PR #1759, opened 2020-08-23
Prevent breakage when trying to excerpt a Field

bobdenotter:
For example, when doing this: 

```twig
<p>{{ record.content|excerpt(200, false, search|default('')) }}</p>
```

Bolt would break with this error: 

Screenshot 2020-08-23 at 17 32 51


I-Valchev (approved):
That's nice!! 👍

PR #1760, opened 2020-08-24
Bugfix/support db without json

### LINKED ISSUE #1753: Setcontent with `slug` doesn't work.
Neither of these two examples work as intended: 

```
    {% setcontent dummy = "dummies/about" printquery %}
    {{ dump(dummy) }}

    {% setcontent dummy = "dummies" where {'slug': 'about'} printquery %}
    {{ dump(dummy) }}
```

Screenshot 2020-08-22 at 15 29 22

(obviously, i've verified `about` exists and is set to be published)

@I-Valchev Could it be something broke in the latest refactoring?

I-Valchev:
@bobdenotter wow! such a good example of fixing "bugs" that actually also did good stuff! 😳

Fixes #1753


bobdenotter (approved):
NOICE

PR #1765, opened 2020-08-24
Fall back to "dumb" numeric sorting for older SQLite versions

bobdenotter:


I-Valchev (approved):
Let's roll with this!

PR #1770, opened 2020-08-25
For now, only use setcontent with standalone fields

### LINKED ISSUE #1769: `setcontent` %voluptat% does not work. Searches sub-fields
The "search like" option does not seem to work. It returns unexpected results, that contain values of sub-fields.

image

I-Valchev:
Fixes #1769


bobdenotter (approved):
👍

PR #1774, opened 2020-08-26
Fix orderby non-localizable field in localized contenttype

### LINKED ISSUE #1772: setcontent order does not work if field is not localized but content is
If the contenttype has defined locales and we have a field, e.g. `date` which is **not** localized, we cannot order by it (because it returns 0 results).

Bolt should be smart enough to not put the condition that the locale be the same, if the field is not localized.

I-Valchev:
Fixes #1772


bobdenotter (approved):
Whoop whoop!

PR #1773, opened 2020-08-26
Test/db supports json

I-Valchev:

PR #1776, opened 2020-08-26
Better check for `INSTR` and `CAST`

bobdenotter:
Additional fix for #1765


bobdenotter:
Same deal as with #1773

PR #1781, opened 2020-08-26
Get curl_options as an array instead of collection

bobdenotter:


I-Valchev (approved):
🙈  

Awesome, I'll pretend I didn't write that!

PR #1786, opened 2020-08-27
Avoid caching slug uses definition for new records

### LINKED ISSUE #1785: Slug sometimes not slugging
Sometimes (I've seen it twice, and also reported by @peterboorsma), the slug doesn't generate. 

For the same site, my local setup works: 

Screenshot_2020-08-27_at_08_18_02

But on production it doesn't work: 

Screenshot_2020-08-27_at_08_18_18

The contenttype is pretty straightforward: 

```yaml
news:
    name: News
    singular_name: Newsitem
    recordsperpage: 5
    fields:
	title:
            type: text
            class: large
        image:
            type: image
        content:
            type: redactor
       	slug:
            type: slug
            uses: title
```

Could this be related to issue #1784, not detecting that "title" is indeed a valid field to generate the Slug from? 🤔

I-Valchev:
Fixes #1785 

Issue is with cache (hence why it only happens _sometimes_ in production?). The method call is, for some reason, cached. Could be that on new objects it's cached the wrong way.

Calling the `definition.uses` in Twig directly resolves it.

I'm not super comfortable with the caching problem itself, however...

### DIFF
--- a/src/Entity/Field/SlugField.php
+++ b/src/Entity/Field/SlugField.php
@@ -33,9 +33,4 @@ class SlugField extends Field implements FieldInterface, ScalarCastable
 
         return $this;
     }
-
-    public function getSlugUseFields(): array
-    {
-        return Collection::wrap($this->getDefinition()->get('uses'))->toArray();
-    }
 }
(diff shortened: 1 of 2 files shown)


bobdenotter (approved):
Looks good! :+1:

Let’s keep an eye out for other instances where this might pop up.

PR #1796, opened 2020-09-01
Remove max-width

eduardomart:
max-width was keeping pre and postfix from using 100% of the width of its container


eduardomart:
As evertalbers mentioned on Slack, maybe this is something to consider before merging...

> From a legibility perspective there is something to be said for a max width of plm. 60-70 chars which may be around 46 em.

bobdenotter (approved):

bobdenotter:
@eduardomart Too late, marged already ;-)

PR #1799, opened 2020-09-01
shouldreturnsingle handles id filtering based on operator

I-Valchev:
e.g. `{% setcontent pages = 'pages' where { 'id': '>50' } %}` should not return a single record, even if it's filtering by ID.


bobdenotter (approved):
Cool! 👍

PR #1803, opened 2020-09-01
WIP: Czech localization

VentyCZ:
Adds Czech locale


bobdenotter (approved):
Thanks, @VentyCZ !

After we've released 4.0, let's see if we can get the whole thing moved to Crowdin or somewhere else! :-)

PR #1811, opened 2020-09-02
Fix `Notice: Array to string conversion` in `Field.php` by ensuring  the  array is one level deep.

bobdenotter:
This happens for fields that don't implement their own `__toString()` method, like `type: data`, or potentially Fields added by extensions. 

Schermafbeelding 2020-09-02 om 11 24 47


src/Entity/Field.php
```diff
@@ -79,7 +80,7 @@ public function __toString(): string
         $value = $this->getTwigValue();
 
         if (is_array($value)) {
-            $value = implode('', $value);
+            $value = implode('', Arr::flatten($value, INF));
```
bobdenotter, line 83:
```suggestion
            $value = implode('', Arr::flatten($value, PHP_INT_MAX));
```

I-Valchev (approved):
Nice catch! 👍

Issue #1821, opened 2020-09-03
strok() expects parameter 1 to be string, bool given on viewless homepage


I-Valchev:
On `bolt/project` with just the homepage contenttype, as is.

It works if the CT is not viewless.

![image](https://user-images.githubusercontent.com/7093518/92088735-91e4f000-edcd-11ea-8225-586f0198e73e.png)

PR #1822, opened 2020-09-03
Don't break if we're trying to get content for "no ContentTypes"

bobdenotter:


bobdenotter:
Fixes #1821

I-Valchev (approved):
✅  Nice one!

PR #1827, opened 2020-09-03
Prevent recursion in `title_format`

### LINKED ISSUE #1825: Recursion in `title_format:` breaks because of recursion
If you have a `title_format` of - say - `{title}` or a non-existing field like `{pompidom}`, Bolt breaks because of recursive calls to `ContentHelper->get()`.

bobdenotter:
Now `title_format: '{title} pom {foo}'` will result in: 

Schermafbeelding 2020-09-03 om 13 03 56

Fixes #1825

### DIFF
--- a/src/Utils/ContentHelper.php
+++ b/src/Utils/ContentHelper.php
@@ -175,6 +175,13 @@ class ContentHelper
                     return $field;
                 }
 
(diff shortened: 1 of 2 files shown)


I-Valchev (approved):
Good catch!😇  👍

PR #1830, opened 2020-09-04
Collapsible collections

I-Valchev:
image

image

### DIFF
--- a/assets/js/app/editor/Components/Collection.vue
+++ b/assets/js/app/editor/Components/Collection.vue
@@ -23,8 +23,7 @@
       <details :open="state === 'expanded'">
         <summary>
           <!-- Initial title. This is replaced by dynamic title in JS below. -->
-          <div class="collection-item-title"
-          :data-label="element.label">
+          <div class="collection-item-title" :data-label="element.label">
             <i :class="[element.icon, 'fas fa-fw']" />
             {{ element.label }}
           </div>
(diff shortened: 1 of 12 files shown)


bobdenotter:
Looks good so far! Were you planning on modifying the "Add" menu to be a bit more like it was in Bolt 3? 

Here's an example I found on one of our sites (NTS): 

Schermafbeelding 2020-09-08 om 10 44 58

I-Valchev:
@bobdenotter yes I tried that but there's an issue with the multi-select tool we us. Its docs say that you can listen to the `select` event when something is clicked (hence when we add a new item), but in reality that only works like `onchange`. That doesn't work for us.

I'm creating an issue for it here https://github.com/shentao/vue-multiselect/issues/1324

bobdenotter:
Maybe it shouldn't be a Vue-select thingy, but just a simple dropdown button: 

https://getbootstrap.com/docs/4.0/components/dropdowns/

eduardomart:
May I suggest an option for showing icons only? as opposed to icon and label on each button for move/up/down/delete :-)

I-Valchev:
@eduardomart I think that's one of the things that advanced users of Bolt are going to appreciate (less crowded interface, more space), but for people starting out with the CMS there's a high chance it won't be clear what those buttons are for.

Right now, let's merge this in as is and think about improving it further still!

bobdenotter (approved):
Whoop whoop!

PR #1831, opened 2020-09-04
Use `travis_retry` command on frequently failing behat tests

I-Valchev:
According to the [docs](https://docs.travis-ci.com/user/common-build-problems/), we can re-run failing commands within the same build, by default up to 3 times!

Hopefully, that makes behat behave!


bobdenotter (approved):
Whoa, cool!

PR #1849, opened 2020-09-15
Check if extension exists as a twig test

I-Valchev:
```
{% if 'bolt/core' is extension %}
   <p>The extension is installed.<p>
{% else %}
   <p>The extension is <strong>NOT</strong> installed.</p>
{% endif %}
```


bobdenotter (approved):
Much nicer!

PR #1850, opened 2020-09-15
Update config.yaml

eduardomart:
{contenttype}/{year}/{month}/ instead of year/month/contenttype

### DIFF
--- a/config/bolt/config.yaml
+++ b/config/bolt/config.yaml
@@ -181,7 +181,7 @@ accept_media_types: [ gif, jpg, jpeg, png, svg, pdf, mp3, tiff ]
 accept_upload_size: 8M
 
 # Default location for uploading files.
-upload_location: "{year}/{month}/{contenttype}/"
+upload_location: "{contenttype}/{year}/{month}/"
 
 # Options to use with curl requests.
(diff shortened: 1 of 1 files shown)


I-Valchev:
hi @eduardomart , one thing I like about `year/month/contenttype` is that you can directly move folders between, say, multiple servers based on the date. 

With `contenttype/year/month`, of course, you can still do that but you'd have to replicate it for all CTs..

Is there a specific use case why `contenttype/year/moth` as the default will work better? 🙂

bobdenotter (approved):
@I-Valchev This is one of those things where I personally agree with you, but apparently the general consensus is otherwise. ;-)

Tests are failing, but that's because of #1851

eduardomart:
@I-Valchev I am not sure I follow, if this is the default why do you have to replicate it?
While I prefer contenttype/year/month, I have no issue if the default is reverted to year/month/contenttype

I-Valchev:
@eduardomart what I meant is that when people upload files by date, if you look at the file structure, you can copy files uploaded on certian days easier.

But, as @bobdenotter says, more people seem to prefer the contenttype/year/month default. Let's roll with it 🙂

PR #1852, opened 2020-09-15
Multiple collections on the same page work without duplicated event listeners

### LINKED ISSUE #1818: Collection/Set 'move up' and 'move down' buttons behave erratically
It seems like the Move up and Move down buttons jump 2 elements, instead of 1 by 1

Demo:
screencast 2020-09-02 14-02-34

**Bolt 4 RC 41**

I-Valchev:
Fixes #1818 with thanks to @eduardomart for spotting this and giving detailed responses so I could reproduce this!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ script:
   # run API tests
   - make behat-api-quiet
   # run JS e2e tests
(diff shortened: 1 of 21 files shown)


eduardomart:
Thank you @I-Valchev !

PR #1856, opened 2020-09-15
Allow multiple roles to be selected for users

I-Valchev:
Strictly speaking, 1 role should be more than enough given Symfony roles are hierarchical...

But at the same time, our `User` entity has multiple roles. + The `bolt/users` extension does not (yet?) use hierarchy.

In any event, it should not hurt that 1 user may have multiple roles.


bobdenotter (approved):
Nice!

PR #1861, opened 2020-09-16
Slugs in fixtures use the `uses:` attribute to make sensible slugs

### LINKED ISSUE #1860: Slugs in fixtures don't use the `uses:` attribute to make sensible slugs.

I-Valchev:
Fixes #1860


bobdenotter (approved):
Green lights.

PR #1880, opened 2020-09-21
Update hidden.html.twig

eduardomart:
- Fixes typo in pattern
- Adds prefix/postfix


bobdenotter (approved):
Thanks! I'll merge this in for now. At some point in time we should re-evaluate how this field works though.. A "pattern" for a hidden field feels weird. It's not like the Editor has a chance of fixing it, if the input is incorrect, since they (by design) can't see or interact with the Field. 

I think ideally the hidden field should be hidden with _no_ output, and we'll make another way to add descriptions or other small blurbs in the editor in a different way.

PR #1886, opened 2020-09-22
Add pedantic notice to installation on this page, reminding people to use https://docs.bolt.cm/installation instead

bobdenotter:


I-Valchev (approved):
👍

PR #1904, opened 2020-09-24
Collection updates

eduardomart:
### Removes faint white border from bottom of collection cards

**before & after:**

Screen Shot 2020-09-24 at 3 04 31 PM

### Removes unnecessary 'set >>' labels:

**before & after:**

Screen Shot 2020-09-24 at 1 31 03 PM


bobdenotter (approved):
Thanks, @eduardomart 🚀

PR #1906, opened 2020-09-25
Update _base.scss

eduardomart:
Removes underline from  Redactor and Article toolbar buttons, the underlines only appear when Redactor or Article are within a Collection card.

Before and After:
Screen Shot 2020-09-24 at 8 02 18 PM

### DIFF
--- a/assets/scss/init/_base.scss
+++ b/assets/scss/init/_base.scss
@@ -57,9 +57,6 @@ small {
     margin-bottom: 1rem;
   }
 
-  a:not(.btn) {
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Cool, thanks! 👍

bobdenotter:
Unfortunately, this will also remove links from places where we _do_ want them: 

Screenshot 2020-09-26 at 14 07 06

I'm going to close this one, and fix it in `bolt/redactor` instead!

PR #1909, opened 2020-09-27
Try and resolve Extension classes from the DI container

rossriley:
This is useful if the base extension has a complex dependency graph.

You can configure the base extension class in `services.yaml` eg:

```yml
Bolt\Extension\SomeCompany\CoolExtension:
        public: true
```

and it will then be constructed with dependencies resolved.


bobdenotter (approved):
Thanks, @rossriley 

I've bumped Travis, and after it passes, I'll merge it in! :-)

PR #1919, opened 2020-09-28
phpstan: remove lock on container and database

TomasVotruba:
Follow up to https://github.com/bolt/core/pull/1918#discussion_r496020968

## Reasoning Behind Changes

I see some first complications - https://github.com/bolt/core/runs/1177089242

It might be caused by incorrect `composer.json` setup:

```bash
       "post-install-cmd": [
            "php bin/console extensions:configure --with-config",
            "@auto-scripts",
            "php bin/console bolt:info"
        ],
        "pre-package-uninstall": [
            "php bin/console extensions:configure --remove-services"
        ],
        "post-update-cmd": [
            "php bin/console extensions:configure",
            "@auto-scripts",
            "php bin/console bolt:info"
        ],
        "auto-scripts": {
            "cache:clear --no-warmup": "symfony-cmd",
            "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd"
        },
```

This is seems like a **project `composer.json`**, e.g [symfony/demo](https://github.com/symfony/demo/blob/e5e5a8fff048351ea8a8fe2b418f0256eabeee12/composer.json#L75-L85) or [laravel/laravel](https://github.com/laravel/laravel/blob/a6ca5778391b150102637459ac3b2a42d78d495b/composer.json#L50-L59)

I've found these lines in [bolt/project](https://github.com/bolt/project/blob/b42dc1feb5c82952e12875096586ec6d4afc4fff/composer.json#L25-L43)

Saying that, this should be part of build script or Dockerfile, not `composer.json`.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -5,9 +5,6 @@ parameters:
     paths:
         - src
 
-    symfony:
-        container_xml_path: '%rootDir%/../../../var/cache/dev/Bolt_KernelDevDebugContainer.xml'
-
     scanDirectories:
         # In order to 'recognize' Twig functions in global scope
         - %currentWorkingDirectory%/vendor/twig/twig/src/Extension
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,4 @@
 parameters:
-    tmpDir: var/cache/ecs
     level: 5
 
     paths:
--- a/composer.json
+++ b/composer.json
@@ -134,28 +134,6 @@
     },
     "minimum-stability": "stable",
     "prefer-stable": true,
-    "scripts": {
-        "post-install-cmd": [
-            "php bin/console extensions:configure --with-config",
-            "@auto-scripts",
-            "php bin/console bolt:info"
-        ],
-        "pre-package-uninstall": [
-            "php bin/console extensions:configure --remove-services"
-        ],
-        "post-update-cmd": [
-            "php bin/console extensions:configure",
-            "@auto-scripts",
-            "php bin/console bolt:info"
-        ],
-        "auto-scripts": {
-            "cache:clear --no-warmup": "symfony-cmd",
-            "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd"
-        },
-        "periodical-tasks": [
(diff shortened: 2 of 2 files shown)


phpstan.neon
```diff
@@ -1,13 +1,9 @@
 parameters:
-    tmpDir: var/cache/ecs
     level: 5
 
     paths:
         - src
 
-    symfony:
-        container_xml_path: '%rootDir%/../../../var/cache/dev/Bolt_KernelDevDebugContainer.xml'
-
```
TomasVotruba, line 10: This actually disables the idea of **static** analysis. Now the whole database connection + configuration and fully built dependency container must be working, before PHPStan can even find a missing method. 

It's only possible inside container, which is kind-off overkill for static analysis.

I personally do not use this lock, as it slowls down the feedback and couples layers that should be separated.

TomasVotruba:
This is ready to merge.

Could sb re-run CI on Travis? It might not be related to changes here

bobdenotter:
@TomasVotruba I bumped the failed tests. re-running!

TomasVotruba:
Thanks.

It seems it's the same test again, it passes on PHP 7.2 and PHP 7.4, but not on PHP 7.3 :thinking: 
Do you think it's related to `composer.json` changes somehow?

I-Valchev:
Hi,

Love all the nice work @TomasVotruba  👏

I have the feeling that by removing the scripts form `composer.json`, people who work _with_ Bolt (i.e. contributors) won't have the clone configured as it should. For example, the extension files won't be copied.

Is there a way to avoid this?

TomasVotruba:
Hi, we're duscussing it here: https://github.com/bolt/core/pull/1918#pullrequestreview-497616124
I understand it saves some work on one place, it just breaks `composer install` workflow on the other place.

Usually these build/setup/run/assets scripts are part of Dockerfile or `npm run setup` or `setup.sh`.
Together with other setup steps that are needed for first run.

TomasVotruba:
Surpassed by https://github.com/bolt/core/pull/1958

PR #1884, opened 2020-09-21
Allow numeric slugs with `allow_numeric` attribute in ContentType

bobdenotter:
Fixes #1877 

This one's for you, @eduardomart 😘

PR #1921, opened 2020-09-29
fix displaying page (singleton) by slug

### LINKED ISSUE #1920: error displaying page (singleton) by slug
|----------------- | ------
| Relevant Bolt Version |  4.0 

```
Error:
Argument 1 passed to Bolt\Repository\ContentRepository::findOneBySlug() must be of the type string, int given
```
``` php
// src: src\Controller\Frontend\DetailController.php:46

$record = $this->contentRepository->findOneBySlug($slugOrId, $contentType);

```

trying to get record by Slug (strict String), but passing Id

denis-gorin:
Fixes #1920

if we allow numeric slug
I suppose to convert numeric slug to String couse not to brake repository method `findOneBySlug`

### DIFF
--- a/src/Repository/ContentRepository.php
+++ b/src/Repository/ContentRepository.php
@@ -128,8 +128,12 @@ class ContentRepository extends ServiceEntityRepository
         return $this->find($id);
     }
 
-    public function findOneBySlug(string $slug, ?ContentType $contentType = null): ?Content
+    /*
+    * @param string|int $slug
+    */
+    public function findOneBySlug($slug, ?ContentType $contentType = null): ?Content
     {
+        $slug = (string) $slug;
         $qb = $this->getQueryBuilder();
         $connection = $qb->getEntityManager()->getConnection();
 


bobdenotter:
Hi @denis-gorin,

How are you triggering this? Are you calling the method yourself, instead of via routing? AFAIK, it should always be a string already, even if it's a numeric one.

denis-gorin:
> Hi @denis-gorin,
> 
> How are you triggering this? Are you calling the method yourself, instead of via routing? AFAIK, it should always be a string already, even if it's a numeric one.

- There is nothing in my `config\routes.yaml`
- I have several singletones 
- and corresonding to singletone's names - several templates (ex. company => company.twig)

when I call `/company` page I've got an error

```
Argument 1 passed to Bolt\Repository\ContentRepository::findOneBySlug() must be of the type string, integer given

ContentRepository->findOneBySlug(2, object(ContentType))

called in \vendor\bolt\core\src\Controller\Frontend\DetailController.php on line 46

```

**Request info**

```
Matched route "listing_locale".

{
    "route": "listing_locale",
    "route_parameters": {
        "_route": "listing_locale",
        "_controller": "Bolt\\Controller\\Frontend\\ListingController::listing",
        "_locale": "en",
        "contentTypeSlug": "company"
    },
    "request_uri": "https://127.0.0.1:8000/en/company",
    "method": "GET"
}
```

develth:
Confirm, have the same error. This PR fixes it

I-Valchev:
@denis-gorin could it be that the param converter is trying to be smart and type hints the `slug`? If that's so, maybe we can just do like so:

 `public function findOneBySlug(string $slug, ?ContentType $contentType = null): ?Content`

rather than casting it to a string?

denis-gorin:
> `public function findOneBySlug(string $slug, ?ContentType $contentType = null): ?Content`

exactly that it was
so, method `findOneBySlug` waiting for a string but got an integer 

The roots of an `integer` are in this

``` php
//src: src\Controller\Frontend\ListingController.php:61

return $this->forward($controller, ['slugOrId' => $content->getId()]);
```

... trying to "list" a singleton, we're getting a single Content **by Id**

bobdenotter:
@denis-gorin I've managed to reproduce it! Thanks for the extra info! We'll merge in, after Travis passes.

@I-Valchev It's already typehinted as `string`. I believe conversion like that is only done in Autowiring (in calling PHP code), _or_ by PHP but then only in absence of `strict_types=1`.

bobdenotter:
For those curious: I've looked it up, and this indeed got broken very recently. I introduced this bug here: #1884

I-Valchev (approved):
👍  Thanks for fixing this @denis-gorin . I think we're ready to roll with this PR once the final test passes :-)

PR #1922, opened 2020-09-29
Fix: Don't break when passing in params for "Search Query"

bobdenotter:


I-Valchev (commented):
@bobdenotter I wonder if this will break it if you'd like to seach for something that is empty?

bobdenotter:
I think we need to differentiate between two types of "breaking" here: 

1) Malformed input causing Bolt to show an error page (which it should _never_ do)
2) Malformed input causing incorrect results (which is unavoidable, because we don't control what's being input)

This PR fixes ~one~ two instances of the former, where I managed to get an exception screen. 

I did some monkey-testing on searching for non-existing Fieldnames and the like. Searching for 'empty' values treats it like that parameter isn't there, so I think that's OK.

I-Valchev (approved):
@bobdenotter right I see! Let's roll with this 👍

PR #1925, opened 2020-09-30
Empty image Field should return `null` instead of `files/`

### LINKED ISSUE #1911: After deleting an image from a contenttype, the image data is still visible
I have a contenttype (singleton, viewless), with an image contenttype:

```
        homepage_picture:
            label: Homepage afbeelding
            type: image
            alt: true
```

In the template, I try to check if the image exist with:

```
{% if homepage.homepage_picture %}do something{% endif %}
```

When I start with a new content type, and don't select an image, 'do something' will not show.  (as expected)
When I select an image (from the library), it (the text) will show. (as expected)
But when I then remove the image (and save the content type, and clear cache), it (the text) will stay visible.

Also, when I do `{{ dump(homepage.homepage_picture|media) }}`, it will keep showing the data for the image I just removed. 
When I try to display the image (after I removed it) with `{{ homepage.homepage_picture.alt }}`, just `` will be outputted.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.0
| Install type                   |  composer
| PHP version                 | 7.4.10 |
| Web server                  | Apache / docker |

bobdenotter:
Fixes #1911

### DIFF
--- a/src/Entity/Field/FileExtrasTrait.php
+++ b/src/Entity/Field/FileExtrasTrait.php
@@ -10,8 +10,12 @@ use Symfony\Component\HttpFoundation\Request;
 
 trait FileExtrasTrait
 {
-    private function getPath(): string
+    private function getPath(): ?string
     {
+        if (empty($this->get('filename'))) {
+            return null;
+        }
+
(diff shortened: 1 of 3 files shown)


I-Valchev (approved):
👍

PR #1927, opened 2020-09-30
Exclude relations from `Content::toArray()`

### LINKED ISSUE #1900: Content->toArray() always returns [] for relations.
When calling` toArray()` on an object of type `Bolt\Entity\Content` you'll get a serialized version of the content. However, the current implementation always returns the empty array for 'relations'.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.0
| PHP version                 | 7.2 / 7.3 / 7.4 |
| Web server                  | NA
| For UX/UI issues        | NA

Reproduction
------------

Take a look at the source of Content.php (https://github.com/bolt/core/blob/6d838da3e4bebfd702f4512200295d7b840d7846/src/Entity/Content.php) and locate the toArray() function to the 'relations' being set to [] https://github.com/bolt/core/blob/6d838da3e4bebfd702f4512200295d7b840d7846/src/Entity/Content.php#L714.

### Steps to reproduce

Get a reference to a content object in php that has at least one relation set and call $myContentItem->toArray();

The resulting array has a key 'relations' that will be [], even if the content does have a relation.

### Expected result

I'd expect the toArray() call to either return with the actual relations, or alternatively without the 'relations' key at all. In the last case I'd expect this to be mentioned in the phpdoc of the function.

### Actual result

The key 'relations' is always [].

I-Valchev:
Fixes #1900 

Right now there's ways to get relations both in code (using the repository) or in twig using the various filters.

I'm not sure that's the _most desirable_ way to go, but as @simongroenewolt points out it's better to _not_ have them than return something wrong.

### DIFF
--- a/src/Entity/Content.php
+++ b/src/Entity/Content.php
@@ -711,7 +711,6 @@ class Content
         $result['fields'] = $this->getFieldValues();
 
         $result['taxonomies'] = $this->getTaxonomyValues();
-        $result['relations'] = [];
 
         unset($result['contentTypeDefinition']);
         unset($result['contentExtension']);


simongroenewolt (commented):
I had a look at the code and agree that leaving them out is a better solution than trying to include the relations. Including them looks hard from this place in the code.

I-Valchev:
@simongroenewolt you are right, it is difficult because we cannot use autowiring in entities. So at best a solution will be patchy.

Good news is, we have ways to retrieve them both programatically and from the templates, so hopefully no harm done :-)

bobdenotter (approved):
Agreed, let's roll with this for now. ^_^

PR #1930, opened 2020-10-01
fix HTML DOM problems

LordSimal:
As I looked through some of the TWIG templates my IDE (PHPStorm) marked some rather obvious HTML problems.

This PR cleans them up so at least these parts have valid HTML.

I did not add any containers or changed tags, I only moved mostly ending tags to the correct position or removed duplicate ending tags.


I-Valchev (approved):
👍  Awesome, thank you for spotting and resolving this!

bobdenotter (approved):
Thanks, @LordSimal 👍

PR #1933, opened 2020-10-02
Initial PostgreSQL support

Wieter:
I've been working to get bolt compatible with PostgreSQL.
From my end everything seems to be working fine, good enough to be used for the purpose at hand (running at FreeBSD 12.1 \w PostgreSQL v11.6).
A quick check from my side on usage of MySQL and SQLite did not show serious problems after fixing a CAST specifically for MySQL (2707e71). 

There might be some minor bugs still around (hence the WIP), that need user testing. For example something strange with the template used (bypassed in c68e6d5) where a selectfield does not have an .selectedIds (integer). Somehow sqlite & mysql have no problems with that, but postgres is more strict. There may arise more of those issues.


bobdenotter:
@Wieter Awesome, thank you for your work! 👍 

I'll keep a close eye on this, and as soon as the `[WIP]` is removed, we'll merge it in!

Wieter:
@bobdenotter Sure no problem! 
I'm just trying to improve the PostgreSQL (or better: database agnostic) compatibility while not breaking other things :) 
A new (postgres-only), more tricky, issue popped up at the admin search /bolt/?filter=css 
SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: jsonb ~~ unknown
LINE 1: ...t_type = $6) AND (b0_.status <> $7 AND (b2_.value LIKE $8 AN...
seems like b2_.value needs also to be casted to TEXT for non-mysql. 
I didn't find out yet where the WHERE clause of this query gets assembled, might take some time. 
Meanwhile I'm looking into the doctrine docs if this can be done automatically through the model annotation, which would be a silver bullet to cover for any remaining issues of the field_translation.value **json-->jsonb** change.

bobdenotter:
Two quick pointers: 

 - Some of the "Json or not" stuff is handled in `wrapJsonFunction()` in `src/Doctrine/JsonHelper.php`. Perhaps the `hasJson()` method in `src/Doctrine/Version.php` needs to be adjusted, too. 

- The heavy lifting of the queries generated from `{% setcontent %}` tags is done here: `src/Storage/Handler/SelectQueryHandler.php`

.env.dist
```diff
@@ -24,7 +24,7 @@ DATABASE_URL=sqlite:///%kernel.project_dir%/var/data/bolt.sqlite
 #DATABASE_URL=mysql://db_user:"db_password"@localhost:3306/db_name
 
 # Postgres
-#DATABASE_URL=postgresql://db_user:"db_password"@localhost:5432/db_name?serverVersion=11"
+#DATABASE_URL=postgresql://db_user:"db_password"@localhost:5432/db_name?serverVersion=11&charset=utf8"
```
bobdenotter, line 27: Is `utf8` not the default? Do we need to explicitly set it here?

config/extensions/acmecorp-referenceextension.yaml
```diff
@@ -0,0 +1,3 @@
+# Reference extension configuration file
```
bobdenotter, line 1: This file doesn't have to be committed.

config/packages/doctrine.yaml
```diff
@@ -31,8 +31,11 @@ doctrine:
                 alias: App
         dql:
             string_functions:
-                JSON_EXTRACT: Bolt\Doctrine\Functions\JsonExtract
-                CAST: DoctrineExtensions\Query\Mysql\Cast
+                JSON_EXTRACT: Bolt\Doctrine\Functions\JsonExtract # Why have a duplicate of the Scienta version?
```
bobdenotter, line 34: I know that it _used_ to be different. I'll check if that's still the case, or if we can fall back to the original. I'll get back to you on that.

config/packages/doctrine.yaml
```diff
@@ -31,8 +31,11 @@ doctrine:
...
+                # JSON_EXTRACT_PATH: Scienta\DoctrineJsonFunctions\Query\AST\Functions\Postgresql\JsonExtractPath
```
bobdenotter, line 35: Let's remove the commented-out ones. :-)

src/.preload.php
```diff
@@ -0,0 +1,5 @@
+<?php
```
bobdenotter, line 1: This file is not needed. I don't know where it came from, but I occasionally see it popping up on my end as well. Regardless, we can remove it here.

src/Doctrine/JsonHelper.php
```diff
@@ -23,9 +23,19 @@ class JsonHelper
     public static function wrapJsonFunction(?string $where = null, ?string $slug = null, Connection $connection)
     {
         $version = new Version($connection);
+        //print_r($version->getPlatform()['driver_name']); # pgsql
+        //exit(print($where)); // translations_anyField.value
```
bobdenotter, line 27: Let's remove commented-out code

src/Doctrine/JsonHelper.php
```diff
@@ -23,9 +23,19 @@ class JsonHelper
...
-            $resultWhere = 'JSON_EXTRACT(' . $where . ", '$[0]')";
+            //PostgreSQL handles JSON differently than MySQL
+            if ($version->getPlatform()['driver_name'] === 'pgsql') {
+                // PostgreSQL
+                $resultWhere = 'JSON_GET_TEXT(' . $where . ', 0)';
+            } else {
+                // MySQL and SQLite
+                $resultWhere = 'JSON_EXTRACT(' . $where . ", '$[0]')";
+            }
+            //exit($resultWhere);
```
bobdenotter, line 38: Same

templates/helpers/_field_blocks.twig
```diff
@@ -43,11 +43,14 @@
 {% block extended_fields %}
 
     {# Special case for 'select' fields: if it's a multiple select, the field is an array. #}
+    {# I dont know what is going on here, but field.selectedIds for pgsql showcase/mr-jean-feeney (and others) #}
+    {# yield field values, NOT integer IDs.. Also, selectedIds does not exists (try dump(field) ) #}
+    {# TODO: FIX IT, bypassing for now by setting field.id #}
```
bobdenotter, line 48: This is weird. Might be an unrelated bug, that's coming to light here.. I'll check this one out as well.

vendor/vaimo/webdriver-binary-downloader/src/Analysers/PlatformAnalyser.php
```diff
@@ -0,0 +1,58 @@
+<?php
```
bobdenotter, line 1: This file doesn't have to be committed. (as it is in `vendor/`)

vendor/vaimo/webdriver-binary-downloader/src/Interfaces/PlatformAnalyserInterface.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
bobdenotter, line 1: This file doesn't have to be committed. (as it is in `vendor/`)

bobdenotter (changes requested):
Shaping up nicely! 👍 

I've left some comments/remarks in your changes.

config/packages/doctrine.yaml
bobdenotter: OK, looked it up: The Scienta versions (like `src/Query/AST/Functions/Mysql/JsonExtract.php`) are specific for SQLite, Mysql or Postgres. Ours extends `AbstractJsonFunctionNode` instead, and can be used on either.

src/Doctrine/JsonHelper.php
bobdenotter: Oh, sidenote: instead of `print_r`, use `dump()`. It's so much nicer. ;-) 

https://symfony.com/doc/current/components/var_dumper.html#dump-examples-and-output

templates/helpers/_field_blocks.twig
bobdenotter: Ok, took a look at this. Not sure why it was failing on your end.. `{{ field.id }}` is not a good fix, because it's an entirely different ID than the one(s) we need. 

It uses method `getSelectedIds` from `src/Entity/Field/SelectField.php`, so it really should exist, if the field is a SelectField.

One thing you might try is to explicitly set it to return an _array_ instead of one item: 

```
{% setcontent selected = field.contentType where {'id': field.selectedIds} returnmultiple %}
```

.env.dist
Wieter: I think it is better to explicitly set here, to force users to think about the charset.
Additionally, it is hardcoded in `config/packages/doctrine.yaml` as `charset: utf8mb4`, which is fine for MySQL but unsupported in PostgreSQL (for as far as I could find). Setting it in the DATABASE_URL ensures an override to UTF8.

vendor/vaimo/webdriver-binary-downloader/src/Analysers/PlatformAnalyser.php
Wieter: Sure :) Just needed this dirty quick-fix to get the test working on my system. I should not forget to remove it from the commits once everything is ready..

vendor/vaimo/webdriver-binary-downloader/src/Interfaces/PlatformAnalyserInterface.php
Wieter: Sure :) Just needed this dirty quick-fix to get the test working on my system. I should not forget to remove it from the commits once everything is ready..

.env.dist
bobdenotter: Ok, fair enough! Does Postgres still handle emoji properly, if it doesn't have `utf8mb4`?

Wieter: Emoji should work as they are UTF-8, I'll have to test it to know for sure. 
From what I understand, MySQL utf-8 is not really full utf-8, it uses 3-bytes per character instead the regular 4byte. I do not know why it was decided, but here lies the origin: [https://github.com/mysql/mysql-server/commit/43a506c0ced0e6ea101d3ab8b4b423ce3fa327d0](url)
That's probably why for MySQL utf8mb4 is required. For postgres. utf8 is full utf8, so 4 byte.   
something to keep in mind in case of problems -> [https://stackoverflow.com/questions/36198847/what-type-should-i-use-to-store-emoji-in-postgresql](url)

templates/helpers/_field_blocks.twig
Wieter: Should be fixed now. selectedIds function returned full (string) values, instead of (integer)IDs. So, why not select on field.value: .. instead of field.id: .. :) 
Works nicely in postgres, I still have to run a test for the others.

src/Doctrine/Version.php
```diff
@@ -17,15 +17,19 @@ class Version
...
      * - 3.28.0 - Not OK (Bob's PHP 7.2, installed with Brew)
      * - 3.29.0 - OK (MacOS Mojave)
      * - 3.30.1 - OK (MacOS Catalina)
      */
-    public const SQLITE_WITH_JSON = '3.27.0';
-    public const PHP_WITH_SQLITE = '7.3.0';
+    // JSON supported since SQLite version 3.9.0
+    public const SQLITE_WITH_JSON = '3.9.0';
+    // PHP supports SQLite since version 5.3.0
+    public const PHP_WITH_SQLITE = '5.3.0';
```
bobdenotter, line 32: This is not correct. If we merge this in, Bolt will absolutely assume it has JSON support on installations where it doesn't have it. 

Maybe we should change the name of the constant, because it's more "PHP_WITH_SQLITE_THAT_HAS_JSON_SUPPORT_BUNDLED". I've definitely seen this break on PHP 7.2 setups. (Ubuntu 2018.04, if i recall correctly)

Ideally, we could properly test if we have it, but I have not found a proper way to do that.. If you have a good suggestion, I'm all for it!

Wieter: I've written a small test and implemented in a new version (e827a799). Meanwhile I'll leave the (restored) old code commented out until we've fully tested everything. Cleanup is the last step.

Wieter:
@bobdenotter somehow the test fails, while if I manually perform the failed action in the test myself (by going to ../bolt/edit/44 and embedding the video) the correct output is given ("Matched embed" --> Silversun Pickups - Nightlight (Official Video), and Silversun Pickups) 

For now, I'll leave it like this and work on something differently. 

sidenote: it would be nice if a test can be run for database-specific implementation, i.e. postgres vs mysql.

bobdenotter:
@Wieter Travis failing is a known issue. It's a bit annoying, but i've restarted the build. 

We have an issue in the works, where we'll improve it: #1918

Wieter:
It seems that for PostgreSQL on my side everything is working nicely, but other users should help test the full functionality as well, as bugs might be hidden. 
Once the Travis tests pass I think we can remove [WIP] and implement the merge. Perhaps mark as "experimental PostgreSQL support"?

bobdenotter:
It looks like some of the tests are failing, because of the JSON check now. See: https://travis-ci.com/github/bolt/core/jobs/396066538

Schermafbeelding 2020-10-07 om 12 35 54

Schermafbeelding 2020-10-07 om 12 36 00

I'll see if I can reproduce that on my end!

bobdenotter:
@Wieter Tests are passing now! 

> Once the Travis tests pass I think we can remove [WIP] and implement the merge. Perhaps mark as "experimental PostgreSQL support"?

That sounds like the prudent thing to do! Let's roll with that. 

Once again, thank you for your contribution on this feature!

bobdenotter (approved):

Wieter:
@bobdenotter You're welcome!
I see that while we are struggling with a new Ziggo modem (working now all day from a mobile hotspot..), you managed to fix the last bugs! Great!

src/Entity/FieldTranslation.php
```diff
@@ -21,7 +21,7 @@ class FieldTranslation implements TranslationInterface
      */
     private $id;
 
-    /** @ORM\Column(type="json") */
+    /** @ORM\Column(type="json", options={"jsonb": true}) */
```
Spomsoree, line 24: @Wieter do you remember if this is part of the fix, or a improvement made on the fly?

Because this is causing trouble for me with mysql: https://github.com/bolt/core/pull/3281

PR #1935, opened 2020-10-02
Better feedback when `|related` and their ilk fail

bobdenotter:
Before, when using `{% set foo = bar|related %}` was used and `bar` wasn't a Content, it'd fail like this: 

Schermafbeelding 2020-10-02 om 16 19 38

Now it shows a more helpful note: 

Schermafbeelding 2020-10-02 om 16 19 16


I-Valchev (approved):
👍 Nice!

PR #1936, opened 2020-10-02
Dont create folders if the image can't be created

### LINKED ISSUE #1928: Buggy thumbnail url generation using "path" and "location" parameters
When generating thumbnails with the `thumbnail` twig helper using `path` and `location` parameters, generates a image URL that the `ImageController` doesn't support.

Details
-------

The ImageController only supports and parses `width`, `height` and `fit` parameters
https://github.com/bolt/core/blob/4.0.1/src/Controller/ImageController.php#L136

But the thumbnail generator can use additional parameters (`location`, `path`) wich the ImageController doesn't support.
https://github.com/bolt/core/blob/4.0.1/src/Utils/ThumbnailHelper.php#L39

So, when generating thumbnails using the not suported parameters like in:
https://github.com/bolt/core/blob/4.0.1/templates/media/edit.html.twig#L19

It generates urls like:
http://127.0.0.1:8000/thumbs/900×500×path=stock/image_12174.jpg

And this confuses the ImageController router:
https://github.com/bolt/core/blob/4.0.1/src/Controller/ImageController.php#L41

And doesn't understands the extra parameters, returning `Error 500` with a `FileNotFoundException`.

|----------------- | ------
| Relevant Bolt Version | 4.0.1
| BC Break                     | -

Reproduction
------------

- Fresh install: `composer create-project bolt/project myprojectname` with dummy content.
- Start built-in webserver: `bin/console server:run`
- Navigate to: http://127.0.0.1:8000/bolt/media/edit/16
- The media images does not load correctly.

UtechtDustin:
I tried to fix #1928, in the meantime i found out that a folder was created in the `thumbs` directory.
This folder will also be created if the image could not be created.

### DIFF
--- a/src/Controller/ImageController.php
+++ b/src/Controller/ImageController.php
@@ -88,8 +88,9 @@ class ImageController
         $fileMode = $this->config->get('general/filepermissions/files', 0664);
 
         try {
+            $imageBlob = $this->buildImage($filename);
             $filesystem->mkdir(dirname($filePath), $folderMode);
-            $filesystem->dumpFile($filePath, $this->buildImage($filename));
+            $filesystem->dumpFile($filePath, $imageBlob);
             $filesystem->chmod($filePath, $fileMode);
         } catch (\Throwable $e) {
             // Fail silently, output user-friendly exception elsewhere.


bobdenotter (approved):
Thanks, @UtechtDustin 👍

PR #1937, opened 2020-10-03
reset password command is in the documentation (https://github.com/bo…

simongroenewolt:
…lt/docs/blob/4.0/docs/howto/resetting-a-password.md), but not in the code. This is a simple version. Differences to the documentation: you have to specify the username (e-mail not possible) and you get the option of specifying a password yourself.

### DIFF
--- a/src/Command/ResetPasswordCommand.php
+++ b/src/Command/ResetPasswordCommand.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace App\Command;
+
+use Bolt\Common\Str;
+use Bolt\Entity\User;
+use Bolt\Repository\UserRepository;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+/**
+ * Resets password for the user with the specified username.
+ *
+ * @package App\Command
+ */
+class ResetPasswordCommand extends Command
+{
+    protected static $defaultName = 'bolt:reset-password';
+
+    /** @var EntityManagerInterface */
+    private $entityManager;
+
+    /** @var UserPasswordEncoderInterface */
+    private $passwordEncoder;
+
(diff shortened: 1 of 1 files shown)


simongroenewolt:
Reason to not implement username or email, but only username was to prevent having to specify what happens in edge-cases like someone having a username that is the same as someone else's email.

src/Command/ResetPasswordCommand.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
```
JKetelaar, line 2: Missing
```
declare(strict_types=1);
```

src/Command/ResetPasswordCommand.php
```diff
@@ -0,0 +1,102 @@
...
+        $this->passwordEncoder = $encoder;
+        $this->validator = $validator;
+        $this->userRepository = $userRepository;
+    }
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Reset password for a user.')
+            ->addArgument('username', InputArgument::REQUIRED, 'User name')
```
JKetelaar, line 56: I'd say change 'User name' to 'Username of the user'

src/Command/ResetPasswordCommand.php
```diff
@@ -0,0 +1,102 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $username = $input->getArgument('username');
+
+        /** @var User $user */
+        $user = $this->userRepository->findOneBy(['username' => $username]);
+
+        if (!$user) {
```
JKetelaar, line 68:
```
        if (!$user) {
```
to
```
        if ($user === null) {
```

JKetelaar (commented):
Just my 2¢

src/Command/ResetPasswordCommand.php
simongroenewolt: yep, done.

src/Command/ResetPasswordCommand.php
simongroenewolt: Yes, that's more clear.

src/Command/ResetPasswordCommand.php
simongroenewolt: I don't really see what difference this would make for this code. Is this a code-style thing, or is it actually handling a specific situation better? I've changed it to your suggestion.

JKetelaar (approved):

simongroenewolt:
Ok, I've figured out how to run the coding style check locally - should have done that earlier.

src/Command/ResetPasswordCommand.php
bobdenotter: The `===` is more strict. In the former, also `""`, `false` and `0` would've passed. Sometimes that's what you want ("be strict in what you output, be liberal in what you accept"), but in this case `$user` is determined right before, so strict checking is fine.

bobdenotter (approved):
Looking good! Thanks, @simongroenewolt 👍

PR #1938, opened 2020-10-03
Fix Buggy thumbnail url generation #1928

luistar15:
- Do not split parameters using "x", could cause an other bug when the `location` or `fit` contains an "x".
- Do not use "parameter=value" syntax anymore.
- Assumes always sending both "WIDTH×HEIGHT" parameters.

### DIFF
--- a/src/Controller/ImageController.php
+++ b/src/Controller/ImageController.php
@@ -139,23 +139,23 @@ class ImageController
         $this->parameters = [
             'w' => is_numeric($raw[0]) ? (int) $raw[0] : 400,
             'h' => is_numeric($raw[1]) ? (int) $raw[1] : 300,
-			'fit' => 'default',
-			'location' => 'files',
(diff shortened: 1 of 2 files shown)


luistar15:
How do I proceed?
Do I need to send another pull request fixing the Coding Standard issues? or can it be fixed automatically?

bobdenotter:
> Do not split parameters using "x", could cause an other bug when the location or fit contains an "x".

Right, the current `str_replace` was too lazy, so it needs to be removed. From experience i know that a lot of people have no clue how to input an `×`, so I think we should breing that back in a separate PR. A simple regex `[0-9]x[0-9]` should already do the trick. 

I've nudged Travis. If it passes, i'll merge this in! 👍

bobdenotter (approved):

bobdenotter:
Thanks again, @luistar15

PR #1943, opened 2020-10-04
Better way to allow `x` in thumbnail URLs alongside `×`

bobdenotter:


luistar15:
1. I don't think it's necesary this change, because the thumbnail urls (`/thumbs/100×72×crop×themes/base-2021/image/hero.png`) are generated automatically. The user doesn't have to write thumbnail urls manually.

2. And if we allow alternative syntax for the same thumbnail imagen, this will prevent the server to serve the fisically cached file.
For example for the thumbnail:
/thumbs/100×72×crop×themes/base-2021/image/hero.png

If the fiscally file doesn't exists, the system will create the file and stored in:
/thumbs/100×72×crop×themes/base-2021/image/hero.png

So, the next time someone access this url, the server will server the previos created file.

If we allow an alternative syntax for accessing the same resource:
/thumbs/100x72xcropxthemes/base-2021/image/hero.png

The server will not find this file and will delegate the process to php an generate the thumbnail file again.

3. Another caveat will be to change the parameter separator (`×`) not only for `width` and `height` parameters, but for `fit` and `location` parameters also for consistency, for example:

`100×72×crop×themes` should be the same as `100x72xcropxthemes`

But this will break `fit` and `location` parameters.

------

If it is necessary this change, I will suggest a differente syntax:

`{WIDTH}x{HEIGHT}---{FIT}---{LOCATION}`

This syntax need to be unique so we prevent the 2. caveat I mentioned.

bobdenotter:
I don't fully agree on point 1: It's true that Bolt only generates thumbs using `×`, but we can't control what users input manually. Also, this _used_ to work, but we removed it, so i'd view that as a regression, which we need to mend. 

About your second and third points: I agree. But, that significantly broadens the scope of this issue. These should be addressed in a separate PR. :-)

I-Valchev (approved):
I think people will try to use normal `x` as opposed to `×`, and they'd be surprised that it won't work. I agree it's definitely not optimal to be regenerating this every time we use the `x`, though. If there's a way to fix it, I'm all for it! In the meantime, afaik this is good to go :-)

PR #1945, opened 2020-10-04
[WIP] Feature/server side validation support

simongroenewolt:
NOTE: this PR is meant for discussion, see https://github.com/bolt/core/issues/1931 -- please don't merge it just yet.

These changes show a possible change to bolt to make it _support_ server side validation. This change is not meant to add validation itself to bolt (at the moment), just to make it possible for other developers without touching the core.

At the moment a class Bolt\Demo\DemoContentValidator is included to test the changes. Without using the suggested support structure it's hard to see if it is good enough.

Currently missing: field-level feedback in the editor.

### DIFF
--- a/src/Controller/Backend/ContentEditController.php
+++ b/src/Controller/Backend/ContentEditController.php
@@ -24,6 +24,7 @@ use Bolt\Repository\MediaRepository;
 use Bolt\Repository\RelationRepository;
 use Bolt\Repository\TaxonomyRepository;
 use Bolt\Utils\TranslationsManager;
+use Bolt\Validator\ContentValidatorInterface;
 use Carbon\Carbon;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\ORMInvalidArgumentException;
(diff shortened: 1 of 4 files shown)


bobdenotter:
Hi, @simongroenewolt,

I finally has some time to poke around at this. I really like it so far.. One thing I noticed is that we autowire `ContentValidatorInterface`. So, that means currently we could only use the "built in" one, or somebody would override it, but lose the built-in one. is that right? 

I tweaked it to look like this: 

Screenshot 2020-10-06 at 21 21 51

```twig
    {% if errors is defined %}
        <div class="alert alert-warning">
            <p>The Record has one or more validation errors. Please correct them, and try again.</p>
            {% dump(errors) %}
            <ul>
            {% for error in errors %}
                <li>
                    {% set propertyPath = error.propertyPath|split(']') %}
                    {% set link = '#' ~ propertyPath[0]|slice(1) ~ propertyPath[1] ~ ']' %}
                    {{ error.propertyPath }}: <a href="{{ link }}">{{ error.message }}</a>
                </li>
            {% endfor %}
            </ul>
        </div>
    {% endif %}
```

Making it link to the location of the error is almost working. Also, I can add repeat the error with the field it applies to. Not hard to add, thankfully. 

Finally, it looks like it really shouldn't be hard to expand this to support all Symfony's validators, like this: 

```yaml
        ip:
            type: text
            variant: inline
            validation:
                ip: ~
```

Cool stuff! 👍

simongroenewolt:
> I finally has some time to poke around at this. I really like it so far.. One thing I noticed is that we autowire ContentValidatorInterface. So, that means currently we could only use the "built in" one, or somebody would override it, but lose the built-in one. is that right?

Correct, I thought that would be the 'Symfony way' to build it. A developer can then use https://symfony.com/doc/current/service_container/service_decoration.html to extend the provided validator, or implement a fully custom one by replacing.

Of course It will be totally ok (and working in the current setup) to create a standard bolt validator that allows limited customization based on additional configuration, or by providing an additional interface that the 'end user' can implement to augment the validation.

simongroenewolt:
> it looks like it really shouldn't be hard to expand this to support all Symfony's validators

I was aiming for that, as it would instantly give bolt validation for lots of fields. I didn't find out yet if more complex validation of something silly like restricting the total character count of all terms added as a 'set' field, or requiring  could be created using a custom constraint/validator.

For me the system would be a success if we make this possible:
- dafault symfony validator constraints on fields (including set and collection)
- required/min/max validation on number of relations and taxonomy (using Symfony\Component\Validator\Constraint\Count constraint)

bobdenotter:
@simongroenewolt Sounds good, on both comments! This will need some work, before we put it out to "the public", so let's merge this into a feature-branch on `bolt/core`. That way both us and you can make additional pull-requests to get it into tip-top shape! Does that sound like it could work for you as well?

simongroenewolt:
@bobdenotter sounds good to me. I've got some outstanding changes that I'd like to commit/push before creating a branch on bolt. (mainly getting the relation via the content class instead of providing it separately)

But if you create a branch I can just change this pull request to target that branch, correct?

bobdenotter:
> But if you create a branch I can just change this pull request to target that branch, correct?

Almost, yes! This PR will be closed, but you can keep pushing to _your_ local branch, and create a *new* follow-up PR to the branch.

simongroenewolt:
@bobdenotter ok, let's do that. I'll create a new PR when you've got a branch ready.

bobdenotter:
@simongroenewolt I've merged it into `bolt:feature/server-side-validation`. Now we can collaborate further on this branch! :-D

PR #1952, opened 2020-10-06
Allow default values for Fields in new Content

bobdenotter:


config/bolt/contenttypes.yaml
```diff
@@ -45,6 +45,22 @@ homepage:
     icon_many: "fa:home"
     icon_one: "fa:home"
 
+dummies:
```
I-Valchev, line 48: 🤪Looks like someone kept the dummies

I-Valchev (changes requested):

config/bolt/contenttypes.yaml
bobdenotter: 🤦

I-Valchev (approved):

I-Valchev:
That's pretty cool! 👍

Issue #1953, opened 2020-10-06
Error when using api/contents endpoint while supplying two or more id's 


sirTachyon:
Hi,

I've came across an error when using the api/contents endpoint (backend -> Maintenance -> API -> /api/contents Retrieves the collection of Content resources).
You can enter an id as integer in the first field (which works just fine) but also an array of id's. When you enter one (existing) id the result all the data of contenttype with the giving id (this also works). 
But when you supply two or more id's you get this error:

`"An exception occurred while executing 'SELECT b0_.id AS id_0, b0_.content_type AS content_type_1, b0_.status AS status_2, b0_.created_at AS created_at_3, b0_.modified_at AS modified_at_4, b0_.published_at AS published_at_5, b0_.depublished_at AS depublished_at_6, b0_.author_id AS author_id_7 FROM bolt_content b0_ WHERE b0_.status = ? AND b0_.content_type NOT IN (?) AND b0_.id IN (?) ORDER BY b0_.id ASC LIMIT 30' with params [\"published\", {\"blocks\":\"blocks\"}, [\"104\",\"116\"]]:\n\nNotice: Undefined offset: 2`

The curl code is:
`curl -X GET "http://127.0.0.1:8000/api/contents?id%5B%5D=104&id%5B%5D=116" -H  "accept: application/ld+json"`

The request URL:
`http://127.0.0.1:8000/api/contents?id%5B%5D=104&id%5B%5D=116`

Things I've checked:
- Both id's exists (quadrupel check this)
- Both id's are of the same contenttype (in my "blogposts")
- When the first id doesn't exists (for example 8080) the error isn't thrown. Just an empty result
- When the second id is set the error is thrown. If it doesn't exists, the error is thrown
- Made an combination of all of the above and a contenttype (in my case "blogposts") and the results are the same

Could you take a look at this? Thanks in advance.

Also, when I've done something stupid, please correct. I'm always to learn.

Tom Schutte
The Netherlands 

![2020-10-06_23h25_28](https://user-images.githubusercontent.com/32093848/95261784-476eed00-082b-11eb-90a1-e45070c85d5a.png)

LordSimal:
Well the error seems to be caused by the following:
![image](https://user-images.githubusercontent.com/9105243/95377610-194ce400-08e3-11eb-9fde-1da590e30637.png)

This is with the API URL `/api/contents?id%5B%5D=2&id%5B%5D=3`

[...]

PR #1960, opened 2020-10-07
$this->viewlessContentTypes should be a normal array. toArray() would…

simongroenewolt:
I think this fixes https://github.com/bolt/core/issues/1953

### DIFF
--- a/src/Api/Extensions/ContentExtension.php
+++ b/src/Api/Extensions/ContentExtension.php
@@ -30,7 +30,7 @@ final class ContentExtension implements QueryCollectionExtensionInterface, Query
             return $ct->get('viewless', false);
         })->map(function (Collection $ct) {
             return $ct->get('slug');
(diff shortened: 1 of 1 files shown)


LordSimal:
I can confirm, that this fixes the issue from #1953 
I tried it on my test instance.

Simon already explained the problem in the issue nicely.

bobdenotter (approved):
Thank you, @simongroenewolt! Great catch! 👍

simongroenewolt:
FYI: A workaround for this issue is to change your configuration to not have any 'viewless' content types, so change every `viewless: true` to false in contenttypes.yaml. Disclaimer: I think this will only change them to be 'publicly visible' to people knowing the url, and linkable in twig by a developer, but it might have other side-effects I don't know of.

sirTachyon:
Thanks for the bug fix! Much appreciated.

PR #1964, opened 2020-10-08
Use `id` in `{% setcontent %}` for Select in Blocks

bobdenotter:


bobdenotter:
Pinging @Wieter,

I had to revert this one bit, because it wasn't working correctly. In fact i'm quite certain it _can't_ work, because `{% setcontent … where { 'value': … } %}` means we're selecting on the field named `value`, which isn't present, normally.

For now i've reverted to the old situation, but let's look into what's causing troubles on your end. 

For me, `{{ dump(field.selectedIds) }}` returns something like `"34"` or `"34 || 32 || 36 || 35"`. Are you certain you get string values, when you put that dump in the code?

Wieter:
@bobdenotter , now, after updating to the latest version, IDs are returned and no longer strings (verified by the dump).
when this started to work I'm not sure. I'll try later to run some tests using earlier commits to find out why strings were returned earlier.  
short conclusion: 46c34c9 works on my end as well (now).

bobdenotter:
@Wieter Awesome! We'll leave it like this for now then. Be sure to let me know if anything breaks. :-)

Wieter:
Ok.. it just broke. 
After pulling the latest master in, and re-filling my .env (overwritten by the pull), I reinitialized the database scheme again and filled the fixtures. now `{{ dump(field.selectedIds) }} ` gives "Numquam tempora aliquam et repudiandae." (for /showcase/gudrun-koepp) instead of an integer. 
I think this could have something to do with loading of the fixtures, but I'm not sure as I currently don't have time for a thorough investigation of this.

PR #1966, opened 2020-10-08
[CI] Move npm jobs from Travis to Github Actions

TomasVotruba:


assets/js/app/common.js
```diff
@@ -2,135 +2,135 @@ import $ from 'jquery';
 import { DateTime } from 'luxon';
 
 $(document).ready(function() {
-  // add a js class to indicate we have JS enabled. Might need a change to either modernizr or somethng comparable
-  $('html').addClass('js');
```
TomasVotruba, line 6: This is just space shift due to `.editorconfig` update. 
#til It's nice that npm style lint adapts to it :)

.editorconfig
```diff
@@ -4,7 +4,5 @@ root = true
 ; Unix-style newlines
 [*]
 end_of_line = LF
-
-[*.php]
 indent_style = space
 indent_size = 4
```
TomasVotruba, line 8: It's seem this file haven't been serving its purpose before.
This allows to have standard 4-space format for each file.

TomasVotruba:
2 less jobs for Travis CI :+1: 

Ready to merge :heavy_check_mark:

TomasVotruba:
CI fail is unrelated, random 1 fail of 3 behat jobs

TomasVotruba:
The space changes were moved to new pull-request, to make it easier to review
https://github.com/bolt/core/pull/1967

bobdenotter (approved):
Cool stuff! Thanks yet again, @TomasVotruba 👍

TomasVotruba:
:+1:

TomasVotruba:
The smalles possible jobs are now split on Github Actions. I guess there is ~3 minutes improvement now and the Travis only handles Behat content.

To analyse how to decouple the last bit, I'd need more time and learn about test structure.

I need to focus somewhere else (on Rector monorepo and phpdoc). So I suggest following, let's give it a month trial and then you ping me and give me a feedback, what needs to be improved.

If we feel in a good mood, we can try to move Behat as well and cut these 10 minutes to ~3. 
What do you think:

bobdenotter:
I am all in favour of this gradual approach we've taken here. Improving one step at a time is a good way to work on Open Source projects. If we would change too sudden, there's always more chance of introducing unwanted side-effects. Taking this approach allows us to keep a better eye on it! 😇 

> If we feel in a good mood, we can try to move Behat as well and cut these 10 minutes to ~3.

Yes, eventually I think we should do that. If only because we can then get rid of Travis entirely. Simple math tells us the change of something breaking is lower when there's less cogs in the machine. ;-)

PR #1967, opened 2020-10-08
[CI] Add stylelint 4 space indent

TomasVotruba:
Follow up to https://github.com/bolt/core/pull/1966 - merge it first 

Just space shift due to .editorconfig update.
#til It's nice that npm style lint adapts to it :)


.editorconfig
```diff
@@ -4,7 +4,5 @@ root = true
 ; Unix-style newlines
 [*]
 end_of_line = LF
-
-[*.php]
```
TomasVotruba, line 8: It's seem this file haven't been serving its purpose before.
This allows to have standard 4-space format for each file.

bobdenotter (approved):
It's quite common in projects that are mainly JS to use 2-space indents, but I think it makes sense here to use 4-space for everything. We already do so for PHP and Twig, anyways! 

Let's roll with this! 👍

bobdenotter (approved):

PR #1972, opened 2020-10-09
Log.php had an incorrect repositoryClass specified in the Entity anno…

simongroenewolt:
Tiny change - Log.php had an incorrect path to the repository class in the annotation.


bobdenotter (approved):
Whoa, that's an old one! ^_^

PR #1977, opened 2020-10-09
Do not cast the order by expression for MySQL drivers as casting a TE…

### LINKED ISSUE #1971: Setcontent with order by causes SQL exception in 4.1
Since upgrading to 4.1, a "setcontent" clause with an order by is causing a SQL exception.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.1.0
| BC Break                     | yes
| For UX/UI issues        | Firefox on Mac OS

Reproduction
------------

### Bug summary

Including an orderby clause in setcontent is throwing a SQL exception in Bolt 4.1.0 (with latest dependencies installed using composer update.)

```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'TEXT)) ASC) dctrn_result) dctrn_table' at line 1").
```

### Specifics

On my home page template, I am using "setcontent" to render the last 6 items of content type "reviews" in descending order based on the "date" field.

My content type:

```
reviews:
    name: Reviews
    singular_name: Review
    fields:
        name:
            type: text
            label: Person's name who left the review
        body:
            type: redactor
            label: Content of the review
        date:
            type: date
            mode: datetime
            label: Date/time the review was posted
        link:
            type: text
            label: Link to see the review (optional)
    icon_many: "fa:users"
    icon_one: "fa:users"
```

The twig tag:

```
{% setcontent reviews = 'reviews' limit 6 orderby '-date' %}
```

This worked correctly in 4.0.1, but is throwing the following exception in 4.1.0:

```
An exception has been thrown during the rendering of a template ("An exception occurred while executing 'SELECT COUNT(*) AS dctrn_count FROM (SELECT DISTINCT id_0 FROM (SELECT b0_.id AS id_0, b0_.content_type AS content_type_1, b0_.status AS status_2, b0_.created_at AS created_at_3, b0_.modified_at AS modified_at_4, b0_.published_at AS published_at_5, b0_.depublished_at AS depublished_at_6 FROM bolt_content b0_ LEFT JOIN bolt_field b1_ ON b0_.id = b1_.content_id AND b1_.type IN ('generic', 'email', 'number', 'collection', 'date', 'filelist', 'textarea', 'set', 'data', 'text', 'image', 'html', 'file', 'hidden', 'checkbox', 'select', 'embed', 'templateselect', 'markdown', 'block', 'slug', 'imagelist', 'article', 'color', 'redactor') LEFT JOIN bolt_field_translation b2_ ON b1_.id = b2_.translatable_id WHERE b0_.content_type = ? AND b0_.status = ? AND b1_.name = ? ORDER BY LOWER(CAST(b2_.value AS TEXT)) ASC) dctrn_result) dctrn_table' with params ["reviews", "published", "date"]:

SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'TEXT)) ASC) dctrn_result) dctrn_table' at line 1").
```

The database is hosted on MariaDB 10.5.5, PHP 7.4.10.

### Steps to reproduce

* Create a content type as above
* Create multiple items of that content type, making sure to specify the date
* Put the twig tag into a template (e.g. the home page template)
* View the page using the twig template (e.g. the home page)

### Expected result

The tag would work, and something like `{{ reviews|length }}` would show the number of review items added.

### Actual result

Exception as described above.

Switching the order by tag to ascending results in the same exception.

Removing the order by tag works but obviously the content is not ordered.

andysh-uk:
Fixes #1971 - the CAST to TEXT causes an error on MySQL.

This follows similar logic implemented in src/Doctrine/Query/Cast.php.

Getting the backend driver name (e.g. mysql) is a long chain of method calls but I cannot see any other way.  It also uses `getDatabasePlatform()->getName()` instead of the `getDriver()->getName()` call in `Cast.php` as the docblock on that method suggests it is deprecated (but no alternative suggested.)

### DIFF
--- a/src/Storage/Directive/OrderDirective.php
+++ b/src/Storage/Directive/OrderDirective.php
@@ -115,10 +115,17 @@ class OrderDirective
             } else {
                 // Note the `lower()` in the `addOrderBy()`. It is essential to sorting the
                 // results correctly. See also https://github.com/bolt/core/issues/1190
-                // again: lower breaks postgresql jsonb compatibility, first cast as txt
+                // again: lower breaks postgresql jsonb compatibility, first cast as txt (but not for MySQL/MariaDB)
+                $backend_driver = $query->getQueryBuilder()->getEntityManager()->getConnection()->getDatabasePlatform()->getName();
+                $lowerExpression = $translationsAlias . '.value';
+
+                if (mb_strpos($backend_driver, 'mysql') === false) {
+                    $lowerExpression = 'CAST(' . $lowerExpression . ' as TEXT)';
+                }
+
                 $query
                     ->getQueryBuilder()
-                    ->addOrderBy('lower(CAST(' . $translationsAlias . '.value as TEXT))', $direction);
+                    ->addOrderBy('lower(' . $lowerExpression . ')', $direction);
             }
             $query->incrementIndex();
         } else {


bobdenotter:
This looks like a correct fix, but in the last two weeks this has been changed back and forth.. 

@wieter , do you know of a way to make it work on Postgres, without breaking MySQL?

Wieter:
@bobdenotter this fix seems incorrect to me, as the queries here are built against Doctrine DRM and not SQL. 
Therefore, this CAST will not be 'native SQL', but should fetched by Doctrine in `src/Doctrine/Query/Cast.php`. 
There, it is either translated into an SQL cast, or omitted (by the driver check). 
I'll install mariaDB and run some tests of my own to see what the $backend_driver yields.

andysh-uk:
My findings were that the line in src/Doctrine/Query/Cast.php was never hit. I had a var_dump just after $backend_driver (before the if condition) followed by an exit(), and I was still getting the exception.

Therefore it felt like the CAST in OrderDirective.php was getting passed as-is to MariaDB, and not being filtered out by the condition in Cast.php.

My $backend_driver does indeed yield “mysql”.

Wieter:
My .env was just erased by ffff76b , it will take a bit more time to get all connections back as I didn't bother to save the dev passwords elsewhere.. :)

Wieter:
@andysh-uk when I insert in `src/Doctrine/Query/Cast.php` on line 24: `exit(print($backend_driver));` I don't get an error but I do get, in my case, 'pdo_pgsql' returned and the page stops rendering. So the src/Doctrine/Query/Cast.php was hit. 
I still have to look into how to test mariadb without breaking mysql dependencies.. I guess this has to be yet another VM :)

andysh-uk:
@Wieter could there be something in Doctrine that it knows how to render the SQL for CAST against MySQL, that it doesn’t with Postgres, so it calls your function for Postgres but not MySQL?

If you can provide me with your IP address I’m more than happy to give you access to my MariaDB dev server, if that helps? If you don’t want to disclose it here, you can send it through [my website](https://andysh.uk/contact)

Wieter:
@andysh-uk thanks for the offer, but I managed. My ip address is temporarily super-dynamic as we still have troubles with the ISP ('unlimited' 4G to the rescue..). 

I was not able to reproduce the error.
Steps taken (roughly):
* install MariaDB:  `Server version: 10.3.22-MariaDB-1ubuntu1 Ubuntu 20.04`
* install php 7.4 with necessary extensions, `php-curl php-xml php-pgsql php-sqlite3 php-mysql`
* clone https://github.com/Wieter/core/tree/development-postgres and checkout development-postgres
* edit .env -> `DATABASE_URL=mysql://bolt_dev:"..."@localhost:3306/bolt_dev`
* the usual composer/npm install procedure
* add your content-type 1:1 as described in #1971 
* add some dummy reviews (first 4, later >10), some without name or content or date. 
* implemented `{% setcontent reviews = 'reviews' limit 6 orderby '-date' %}` within `public/theme/skeleton/listing.twig` with corresponding dump, loop, and dump-per-review.
all works flawlessly
The file `src/Doctrine/Query/Cast.php` is called correctly, and the $backend_driver = pdo_mysql.

I wonder.. what happens with a composer-only fetch and install, through the official release channels. I'll try that later.

Wieter:
I managed to get the error now and will look into it. #7878 contains the exact steps to reproduce.

Wieter:
@bobdenotter I think I found the culprit.

using composer install creates an `config/packages/doctrine.yaml` within the users project folder.
In there, Doctrine>Orm>dql: `CAST: DoctrineExtensions\Query\Mysql\Cast`
This apparently loads `./vendor/beberlei/doctrineextensions/src/Query/Mysql/Cast.php`
and this in turn collides with our Cast.php.

I see more is missing from the doctrine config, it should be:
                JSON_EXTRACT: Bolt\Doctrine\Functions\JsonExtract
                JSON_GET_TEXT: Scienta\DoctrineJsonFunctions\Query\AST\Functions\Postgresql\JsonGetText
                CAST: Bolt\Doctrine\Query\Cast

andysh-uk:
Thanks @Wieter. Is that why I couldn’t get the line to be hit in src/Doctrine/Query/Cast.php?

That would make better sense than my proposed fix in this PR, knowing now that the query I changed is DQL, not pure SQL.

Wieter:
yes @andysh-uk 
Could you try modifying (commenting out the old values) your `config/packages/doctrine.yaml` to the definitions I gave in my previous comment, and see if that fixes/breaks things for your MariaDB installation?

andysh-uk:
@Wieter that works perfectly, many thanks 😄 

I rolled my change back to make sure I was still getting the exception (I was) and then applied your changes to `config/packages/doctrine.yaml` and refreshed the page - works great.

I did double-check that `src/Doctrine/Query/Cast.php` is called and my driver name is "pdo_mysql" - which is a yes to both.

Just FYI, PhpStorm flags the `getName()` call on the Doctrine `Driver` object as deprecated - and [this PR](https://github.com/doctrine/dbal/pull/3558) has removed it.

A better option might be to use `getDatabasePlatform()->getName()` instead of `getDriver()->getName()` for when this method is removed.  This method returns `mysql` instead of `pdo_mysql`.

Wieter:
Good to hear it works. 
Also; this means that PostgreSQL people will have troubles installing Bolt 4.1 as well using composer install.. 

Good finding, indeed `getDriver()` should be changed to `getDatabasePlatform()` should be changed. Feel free to implement and run some tests ;)

andysh-uk:
@Wieter thanks, I'll implement and raise a separate PR.

Issue #1978, opened 2020-10-10
ImageFixtures preview doesn't show correctly when editing in the admin (Windows)


luistar15:
ImageFixtures uses backward slash (`\`) as directory separator for relative paths in Windows.
Example: `stock\image_40672.jpg`

This causes problems when [using in Vue as a css background-image property](https://github.com/bolt/core/blob/4.1.0/assets/js/app/editor/Components/Image.vue#L119) (probably parsed as an escape sequence):
`stockimage_40672.jpg`

Since PHP works the same with paths using `/` as a separator and since the `/` is the path separator in URLs, I think we should use it in this case.

To fix this issue I tried to change the separator from the beginning in the origin.
`Symfony/Finder` uses [`RecursiveDirectoryIterator` wich suports `UNIX_PATHS`](https://github.com/symfony/finder/blob/5.x/Iterator/RecursiveDirectoryIterator.php#L51), but [`Symfony/Finder` doesn't allow such configuration](https://github.com/symfony/finder/blob/5.x/Finder.php#L724).

The next best thing I think is to replace the separator [before storing it in the database](https://github.com/bolt/core/blob/4.1.0/src/DataFixtures/ContentFixtures.php#L278).

Details
-------

| Question                      | Answer
|----------------- | ------
| Operating System | Windows
| Relevant Bolt Version | 4.1
| Install type                   |  Composer install
| BC Break                     | no
| PHP version                 | 7.4
| Web server                  | Built-in
| For UX/UI issues        | Browser name and version

Reproduction
------------

### Steps to reproduce

- Install bolt `composer create-project bolt/project myproject`
- Setup bolt with fixtures (dummy content)
- Start built-in server `bin\console server:run`
- Edit any people entry

### Expected result

The preview of the image field should be displayed.

### Actual result

The preview of the image field doesn't show.

bobdenotter:
Fixed by #1979

PR #1979, opened 2020-10-10
Force path separator to / in ImageFixtures relative paths

luistar15:
Fix bug with ImageFixtures relative paths in windows #1978


bobdenotter (approved):
Thanks, @luistar15

PR #1980, opened 2020-10-11
Feat: add .gitattributes file for release

toofff:
[When creating an archive](https://git-scm.com/docs/gitattributes#_creating_an_archive) we may not take certain files or folders into account.

Here is an example of what we could gain in weight in the vendors of the project:
- `bolt-core-4.1.0.tar.gz` => 3,1 Mo
- `bolt-core-my-test.tar.gz` => 334 Ko

By removing all the files that are not necessary for the core to work in a project vendor, we reduce the package by 90% of its weight.
This is not negligible for a production release and to avoid overloading server resources for nothing.


bobdenotter:
@toofff Cool stuff. 

I'm getting questions occasionally from people who still have trouble using Composer, saying they'd prefer a simple `.zip` distro. Is the technique in this PR something that might be applicable to `bolt/project` as well, so we could potentially make a lighter `.zip` distro for these people?

.gitattributes
```diff
@@ -0,0 +1,32 @@
...
+/package-lock.json export-ignore
+/phpstan.neon export-ignore
+/phpunit.bootstrap.php export-ignore
+/phpunit.xml.dist export-ignore
+/postcss.config.js export-ignore
+/run_behat_tests.sh export-ignore
+/symfony.lock export-ignore
+/webpack.config.js export-ignore
+/assets export-ignore
+/bin export-ignore
```
bobdenotter, line 27: If we exclude this, won't we exclude `bin/console` as well? Or is that no problem because `bolt/project` has its own copy?

toofff: For me "bolt/core" should be taken as a Symfony component, we just need to have the Bolt business code for everything to work properly.

"bolt/project" is a symfony project, so for any symfony project the "bin/console" file must be versioned.

I think this is the best solution to make the deposit lighter while waiting to clean up.

bobdenotter: That makes sense! Let's roll with this! 👍

bobdenotter (approved):

toofff:
@bobdenotter in my PR on "bolt/project" I didn't put it because I didn't need it. But I'll still check and get back to you this week.

PR #1984, opened 2020-10-12
Use DatabasePlatform instead of Driver to check for MySQL/MariaDB

andysh-uk:
Replace getDriver()->getName() call with getDatabasePlatform()->getName() to avoid [a deprecation in Doctrine](https://github.com/doctrine/dbal/pull/3558).

This returns `mysql` instead of `pdo_mysql` on MariaDB, at least, so the `mb_strpos` check should work just fine as-is.

### DIFF
--- a/src/Doctrine/Query/Cast.php
+++ b/src/Doctrine/Query/Cast.php
@@ -20,7 +20,7 @@ class Cast extends FunctionNode
 
     public function getSql(SqlWalker $sqlWalker): string
     {
-        $backend_driver = $sqlWalker->getConnection()->getDriver()->getName();
+        $backend_driver = $sqlWalker->getConnection()->getDatabasePlatform()->getName();
 
         // test if we are using MySQL
         if (mb_strpos($backend_driver, 'mysql') !== false) {


bobdenotter (approved):
Thank you, @andysh-uk for this fix! Also thanks to you @wieter, for helping to get to the root of the problem! 

I love fixes like this! It seems like such an insignificant change, since it literally only changes one method call. The previous fix was a patch, fixing the symptoms. This little one-word change took several hours of work from multiple people, however it is a *proper fix* for an actual bug, and reduces workarounds further down the line! 🤗

Wieter:
@bobdenotter you're welcome, this fix prevents future breakage due to depreciation, however it does not resolve issues #1971 and #1985 (outdated config file as used by composer install)

andysh-uk:
I love your sentiments @bobdenotter  (thank you) however the credit goes to @Wieter for finding the actual issue (with `doctrine.yaml`) - as you said, my previous fix was incorrect as it only worked around a problem, not resolved the root of the problem.

This PR is just something I noticed whilst working on this issue, that the Doctrine project have deprecated this method, believing it only to be used in their tests.

PR #2003, opened 2020-10-15
Don't break if slug is `null` as opposed to "empty string"

bobdenotter:


I-Valchev (approved):
For context: this happens if there's no localization (only 1 langauge) and the default locale is changed after records already exist.

👍 Nice, let's merge this in!

PR #2006, opened 2020-10-15
Prepare release 4.1.1

bobdenotter:


I-Valchev:
@bobdenotter the simplified Chinese translation looks like a _feature_, as Russian and Bulgarian were in 4.1 🤔

bobdenotter:
Yeah, i realised that as well, yesterday.. But, it’s already in master, so taking it out wouldn’t make much sense, i think.

I’d suggest we leave this is as-is, and make a note that we still mention it as a feature, when we do 4.2 with a blogposting.

I-Valchev:
@bobdenotter yes, I'm with you on that one.

PR #2008, opened 2020-10-16
Filter by any field

### LINKED ISSUE #1965: [RFC] Search filter based on specific field
The search filter on ContentType Listing pages in the Editor allows Editors to search based on any fields from a Content record.

image

Sometimes, especially when there are many records in a ContentType or the whole website, it is useful to limit that filter to a specific field, optionally.

This RFC is about creating an optional dropdown in proximity to the filter input, so that it can be limited to one field only.

How can you help?
-------------------------------

I can implement it :-)

I-Valchev:
Fixes #1965 

It also works for things which are not fields, e.g .status=published, but not through the UI.

It would be nice, in a future PR, to allow editors to filter by date, status, author... All things which are possible with this, _once the query params are set correctly_.


bobdenotter (commented):
Noice! 👍

bobdenotter (approved):

PR #2016, opened 2020-10-17
Show actual error message from upload error response instead of [object Object]

simongroenewolt:
When you upload a file that is too large you currently get an error showing [object Object], this change checks if there is a error message in that object, and displays that instead.

This fixes the [object Object] issue mentioned in https://github.com/bolt/core/issues/2000


bobdenotter (approved):
I have no idea who wrote that! 🙈 

Thanks, @simongroenewolt

PR #2026, opened 2020-10-20
Show set as block

### LINKED ISSUE #2012: Missing label for Content-Type Set
Hello,

as mentioned in [my comment](https://github.com/bolt/core/commit/53873f578bf7fa68e68eb9077b6e858a7a7b9567#r43305900), why turned off the Label at this Point?

https://github.com/bolt/core/blob/b436e24e41912c175571fce9749f3e74a76d327e/templates/_partials/fields/set.html.twig#L3

Previously i had a `Set` Label like "More Button" above that:
Bildschirmfoto 2020-10-16 um 16 40 13

Maybe make it configurable that only no Label will be displayed when it is within a Collection?

I-Valchev:
Fixes #2012 

Only affects sets that _are **not**_ in a collection.

Before:
image

After:

image

### DIFF
--- a/assets/scss/modules/editor/fields/_set.scss
+++ b/assets/scss/modules/editor/fields/_set.scss
@@ -1,0 +1,13 @@
+.form-set:not(.in-compound) {
+  @extend .card;
+
+  background: var(--background);
+
+  & > label {
+    @extend .card-header;
+  }
+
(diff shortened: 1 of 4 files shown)


bobdenotter (approved):
Ah, that looks nice! 👍

PR #2029, opened 2020-10-20
Changed name of generic taxonomy config file identifier

bfkolle:
I wasn't able to set the listing template for a taxonomy in taxonomy.yml. Did some snooping in the code and found that this line was referring to the taxonomy config by the wrong identifier. Just changed it to the new identifier and I could now set the listing template in my taxonomy.yml.

Hope I'm putting the pr in the right place, followed the instructions from the Contributing documentation online. There was no issue so I just named it after what I did.

### DIFF
--- a/src/TemplateChooser.php
+++ b/src/TemplateChooser.php
@@ -125,7 +125,7 @@ class TemplateChooser
         $templates = new Collection();
 
         // First candidate: defined specifically in the taxonomy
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Hi @bfkolle, thanks for the PR! 

Good catch!

I-Valchev:
hi @bfkolle , nice catch! Thanks for the PR, merging it in now 🤗

PR #2031, opened 2020-10-22
Don't break on unknown 'fit' parameter

bobdenotter:
If we have a seemingly reasonable URL like `thumbs/580×405×resize/foo/bar.jpg` Bolt shouldn't throw a `500`, but just show the image instead.

### DIFF
--- a/src/Controller/ImageController.php
+++ b/src/Controller/ImageController.php
@@ -196,9 +196,8 @@ class ImageController
                 return 'stretch';
             case 'c':
             case 'crop':
-                return 'crop';
             default:
-                return $fit;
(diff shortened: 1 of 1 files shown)


I-Valchev:
@bobdenotter That is nice, however if we do this without a bit more change to that switch, we're throwing away a few other options that the Glide `fit` parameter supports, e.g.: `crop-left`, `crop-25-75`, `crop-25-75-2`.

https://glide.thephpleague.com/1.0/api/crop/

bobdenotter:
Let's just close this one, and re-visit properly if it's warranted.

PR #2035, opened 2020-10-22
Project specific extension bug in ExtensionExtension.php

c2webstudio:
When a project specific extension is created an exception is throwed in front end because there is no composer package name.


I-Valchev (approved):
👍. Nice catch, thank you for the PR @c2webstudio .

bobdenotter:
Hi @c2webstudio. Nice catch! 👍 

We've merged this into the `4.1` branch. https://github.com/bolt/core/tree/4.1

PR #2038, opened 2020-10-23
Use remember me field in login Symfony Form

I-Valchev:
Follow up from #2007, using remember me as part of the form.


bobdenotter (approved):
✅ Finally!

PR #2044, opened 2020-10-23
Taxonomy singular slug fix

### LINKED ISSUE #2036: Taxonomy singular_slug is being ignored when generating URL
I'm migrating [my existing WordPress website](https://andysh.uk) to be built in Bolt, so I'd like to maintain the URLs where possible.

One particular example is a category link - it should be `/category/software-development`, for example, but Bolt is generating it as `/categories/software-development`.

The taxonomy option `singular_slug` appears to be ignored - the `slug` is used when generating the URL using `link`.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.1.2
| BC Break                     | not sure
| PHP version                 | 7.4 |
| For UX/UI issues        | Firefox 81.0.2

Reproduction
------------

### Bug summary

Using the default Bolt taxonomies set up with a new Bolt 4.1.2 site, categories and tags, the `singular_slug` option appears to be ignored when the URL is generated using the `link` Twig filter.

### Specifics

Using the default Bolt taxonomy for "categories":

```
categories:
    name: Categories
    slug: categories
    singular_name: Category
    singular_slug: category
    behaves_like: categories
    multiple: true
    options:
        - "Something"
        - "Something else"
    required: false
```

You can see the `singular_slug` is set to `category`, while the `slug` is set to `categories`.

However the link generated is /categories/something, instead of /category/something.

Twig template:

```
{% setcontent blog_posts = 'blog' orderby '-publishedAt' %}
{% for blog_post in blog_posts %}
    {% if blog_post|taxonomies['categories'] is defined %}
        {% for category in blog_post|taxonomies['categories'] %}
            <a href="{{ category|link }}">{{ category.name }}</a>{% if not loop.last %}, {% endif %}
        {% endfor %}
    {% endif %}
{% endfor %}
```

This renders as:

```
<a href="/categories/miscellaneous">Miscellaneous</a>
```

My `blogs` content type is defined as follows:

```
blog_posts:
    name: Blog Posts
    singular_name: Blog Post
    slug: blog
    singular_slug: blog
    record_route: blog_post
    fields:
        title:
            type: text
            class: large
            group: content
            label: Title
        content:
            type: article
        slug:
            type: slug
            uses: title
            group: Meta
    taxonomy: [ categories, tags ]
    icon_many: "fa:newspaper"
    icon_one: "fa:newspaper"
```

### Steps to reproduce

* Create a default Bolt install using Composer
* Set up the `blog_posts` content type, as above
* Create a blog post in the `blog_posts` content type and select a category for it
* Add the Twig code to the home page template to render out the categories for each blog post
* Observe the link generated doesn't seem to use `singular_slug`

### Expected result

The `singular_slug` value of `category` is created, like:

```
<a href="/category/miscellaneous">Miscellaneous</a>
```

### Actual result

The `singular_slug` option is ignored, and `slug` is used instead:

```
<a href="/categories/miscellaneous">Miscellaneous</a>
```

You should be able to observe this on [my development site](https://my.andysh.dev/).

andysh-uk:
Fixes #2036.

Taxonomies now use the singular_slug when configured to do so, otherwise they fall back to the slug.

I've replicated the relationship between `Content` -> `ContentFillListener` -> content type config -> `ContentType` with taxonomies, so we have:

`Taxonomy` -> `TaxonomyFillListener` -> taxonomies config -> `TaxonomyType`.

The Twig `getLink` filter now uses the TaxonomyType `singular_slug`/`slug` value (Bolt seems to automatically substitute `slug` if `singular_slug` is not configured, so the method call looks like it only uses singular_slug - but this is the same as Content.)

There are a couple of issues I have:

- I feel this should have a test, as a behaviour change in future could break links. I'm happy to code such a test but I have no idea where to start, so any guidance would be appreciated 😄 
- This is a potentially high-impact change.  All sites created using Bolt will have singular_slug configured on the default categories and tags taxonomies, but this isn't being used currently. I feel the behaviour is now correct based on what is configured, but it is going to change the URLs generated with `|link` on a taxonomy. This should perhaps be either strongly called out in the release notes, or go into a bigger update - e.g. 4.2.


I-Valchev:
Hi @andysh-uk , good catch, I know this has been bugging a colleague of mine, @nestordedios , for some time now 😅

**For tests**
I agree, it'd be great to have more tests! It is best to look at `tests/e2e/display_record.feature` to add some more tests there. There is a bit of [documentation here](https://github.com/bolt/core/tree/master/tests/e2e), albeit not too much. You can take a look at the existing steps. For me, the IDE also auto-completes what is available?

If you bump into issues with it, reach me or @bobdenotter on Slack.

I-Valchev:
About the big change, the other issue is that right now the addition to `services.yaml` won't be automatically included in people's existing projects. Prior to 4.0.0, we sort of sneaked those changes in because it was not a stable release. Now, we should not be introducing breaking changes in .x and .x.x releases. Let's see how best to approach this as well, as I'm sure it'll pop up in future too.

andysh-uk:
Thanks @I-Valchev, I did wonder about the services.yaml change following the previous issue with the Doctrine change in 4.1 😄 

Let me see how this works when that listener is not configured - I feel it would be acceptable to go into a smaller release if the code works the way it does prior to this change if services.yaml is not updated.

That way, you could highlight it in the release notes that to opt-in to this new behaviour for existing projects, you have to make the change manually.

I'm assuming newly-created projects would get the updated services.yaml, and so would start with the desired behaviour from the get-go?

andysh-uk:
@bobdenotter @I-Valchev I'm struggling to understand the PHPstan failure on this PR. I can't see that it's anything to do with what I've changed but of course I could be wrong.

Are you able to offer any pointers please?

bobdenotter:
Hi @andysh-uk , 

The failure is probably unrelated to your changes, see https://github.com/bolt/core/pull/2043 

Sometimes it happens that a minor update in ECS or PHPStan suddenly detects a (valid) issue that it didn't spot previously. :-)

I'm going to rebase this PR, and then it'll probably pass.

bobdenotter (approved):
Merging in to `4.1` as well as `master`! Thanks, @andysh-uk

bobdenotter:
(Oh, about that breakage, there's also this one: #2048)

andysh-uk:
Thanks for sorting Phpstan @bobdenotter. I've just created another small PR #2051 , which prevents an exception being thrown if `services.yml` does not have the new listener configure - it falls back to existing behaviour.

PR #2046, opened 2020-10-23
fix duplicate html element ids in editor when specifying 2 or more collections in one content type

simongroenewolt:
Fixes https://github.com/bolt/core/issues/2045


bobdenotter (approved):
Thanks, @simongroenewolt 👍 

Don't mind the failing tests: we need to clean up the mess caused by Composer 2, PHPStan and ECS breaking at pretty much the same time. :-)

Merging into 4.1 as well as master!

PR #2051, opened 2020-10-24
Taxonomy singular slug fix part 2

andysh-uk:
Following on from #2044, this prevents an exception being thrown when the `TaxonomyFillListener` has not been added to `services.yml`.  Any existing links will still generate as '/categories/something', for example, until the new listener is added.

@bobdenotter please review this before the next 4.1 release or it will break existing sites if `services.yml` is not updated.

### DIFF
--- a/src/Entity/Taxonomy.php
+++ b/src/Entity/Taxonomy.php
@@ -107,7 +107,7 @@ class Taxonomy
     public function getTaxonomyTypeSlug(): string
     {
         if ($this->getDefinition() === null) {
-            throw new \RuntimeException('Taxonomy not fully initialized');
+            return $this->getType();
         }
 
         return $this->getDefinition()->get('slug');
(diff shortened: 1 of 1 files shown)


bobdenotter:
@andysh-uk Nice! 

Don't mind the failing tests.. I'll need to clean up the mess caused by Composer 2, PHPStan and ECS breaking at pretty much the same time. :-)

andysh-uk:
Thanks @bobdenotter, I was a bit alarmed at the amount of failures following such a tiny change, but I know you'd rebased and I did see a lot activity in master around Composer!

I'm happy with this now it's backwards compatible so I'll delete my own branch and update my own repo.

bobdenotter:
Merging into 4.1 as well as master!

bobdenotter (approved):

PR #2059, opened 2020-10-25
Fix parsing of theme.yml

artggd:
Hello 👋 

When using custom `public_dir` folder, the config fails to parse the `theme.yaml` because the `PathResolver` is constructed with the default `$publicFolder` : "public"

I propose a fix by passing `$publicFolder` to `PathResolver`'s constructor .

### DIFF
--- a/src/Configuration/Config.php
+++ b/src/Configuration/Config.php
@@ -29,6 +29,9 @@ class Config
     /** @var Stopwatch */
     private $stopwatch;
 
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Nice catch. Thank you, @artggd 👍

bobdenotter:
Merging into `4.1` as well as `master`!

PR #2060, opened 2020-10-25
update .travis.yml according to info on travis-ci.com

LordSimal:
I just checked the travis-ci.com build of the core because it was failing and I noticed some `build config validation` warnings.
image

This PR basically does exactly what is being explained in there.
We'll see if something breaks or not 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
+os: linux
 dist: xenial
-sudo: false
 php:
     - 7.2
     - 7.3
@@ -13,7 +13,7 @@ cache:
     directories:
         - $HOME/.composer/cache/files
 
-matrix:
+jobs:
     fast_finish: true
 
 before_install:


LordSimal:
Again that `Scenario: As a user I want to see the results of Setcontent # tests/e2e/setcontent.feature:4`...
I think we gotta have to look into that one why this still keeps popping up in some (but not all) tests

simongroenewolt:
> Again that `Scenario: As a user I want to see the results of Setcontent # tests/e2e/setcontent.feature:4`...
> I think we gotta have to look into that one why this still keeps popping up in some (but not all) tests

I took a look, and maybe I'm missing something, but it looks like the tests are run on a database created using Faker, and if you don't set a seed on Faker it will randomize itself. I don't see any code calling seed(). Now this test checks for the word 'voluptat' in the database. My guess would be that on some runs faker simply will not have generated that word for any content type that is being queried. The only reason that it mostly succeeds would then be the big chance of this word being in the generated title at random.

bobdenotter:
> I think we gotta have to look into that one why this still keeps popping up in some (but not all) tests

It happens sometimes, but as we're ironing out other kinks in the CI, this'll start popping up relatively more often. 

>  I don't see any code calling seed(). 

😮

LordSimal:
I can confirm that!

Basically in `src/DataFixtures/ContentFixtures.php` Line 446 the custom Template `public/theme/skeleton/custom/setcontent_1.twig` is being set for the `Setcontent test page`.

image

In that template a `yes` is being printed inside the <span> IF `entries` has some results.

The problem here is - as @simongroenewolt already mentioned - that the query only checks for blocks, entries and showcases which have `voluptat` somewhere in their title.

image

The title of an entry is `text` and therefore is being randomized via this function
src/DataFixtures/ContentFixtures.php Line 310

image

Therefore either refactor that part of the code to always set the title of content with a preset randomzied set of words or call `seed()` for the faker instance 😄

LordSimal:
I will make a separat issue where we can discuss this

bobdenotter:
Some of those fixtures have set values, so we can use them as `404` or similar. This is OK, as far as i'm concerned. 

I just never realised there being a `seed()` method for Faker. If we can update our fixtures to take a seed param, like: 

```
bin/console doctrine:fixtures:load --group=without-images -n --seed=98765
```

Then the tests will become instantly more predictable!

bobdenotter (approved):
Looking good, merging in!

PR #2062, opened 2020-10-26
Fix faker data for e2e tests

### LINKED ISSUE #2061: E2E tests fail sometimes on tests/e2e/setcontent.feature:4
Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.1.4
| BC Break                     | yes/no - What is that? 😅 
| Web server                  |  Built-in
| For UX/UI issues        | Google Chrome

Reproduction
------------

### Bug summary

As we already discussed in #2060 the automatic E2E tests fail due to inconsistencies with the test-data.

### Explanation of the problem

Basically in `src/DataFixtures/ContentFixtures.php` Line 446 the custom Template `public/theme/skeleton/custom/setcontent_1.twig` is being set for the `Setcontent test page`.

image

In that template a `yes` is being printed inside the <span> IF `entries` has some results.

The problem here is - as @simongroenewolt already mentioned - that the query only checks for blocks, entries and showcases which have `voluptat` somewhere in their title.

image

The title of an entry is `text` and therefore is being randomized via this function
src/DataFixtures/ContentFixtures.php Line 310

image

Therefore either refactor that part of the code to always set the title of content with a preset randomzied set of words or call `seed()` for the faker instance 😄

LordSimal:
Closes #2061 

This should fix the problem with e2e Tests not being consistent all the time.

Couldn't test it locally because my webdriver doesn't work locally (yet)


bobdenotter (approved):
Cool! This should be another small step to making Travis less fickle! 👍

PR #2066, opened 2020-10-28
Feature/server side validation

simongroenewolt:
I wanted to be able to express constraints on bolt content types on the server. 

This implementation of server side validation introduces the possibility to specify a validator that will be called when saving data. On top of this I've created a 'default' validator that allows you to add constraints on fields using the 'standard' symfony/validator notation, and adds the possibility of a count (min,max) constraint for relations and collections.

To allow for customization by developers using bolt the validation is injected as an interface into the `ContentEditController` with the implementing class currently specified in `config/services.yaml`

`ContentEditController` has been changed to show the entry with errors if there were validation errors.

To make it possible for the validator to validate the content before saving and the `ContentEditController` to show the edited-but-not-saved version of the content again I had to make a couple of changes that involve the entities.

`Content.php` and `Relation.php` now have functionality to create relations (outgoing and incoming) and use them before saving. Before this change a modification to the relations was one-sided and would only be accessible after saving to the db and reloading via Doctrine.
The data for making these relations in the db was already there, it's just that they are accessible from both sides now. No db migration should be necessary.

To show the content-in-edit in the editor I've added a new twig function/filter `related_values_from_content`, because the original `related_values` would query the db and therefore retrieve the saved state instead of the 'work in progress'.

To show the errors in the editor a small addition was done to the twig template.

After this preparation the actual validation can be done. I've tried to stay as close as possible to `symfony/validator` and have created a default validator that will take its configuration from `contenttypes.yaml`. See the `validationdemos` type as an example.

I don't know if it is necessary, but I've added a global enable switch to `config/bolt/config.yaml` so a release can ship with validation 'off'.

To decide:
- I'm not fond of the names relationsFromThisContent and relationsToThisContent in Content.php -> maybe rename to outgoingRelations and incomingRelations?
- The `related_values_from_content` twig functionality is weirdly named, maybe it would be better to create an option on `related_values` instead?
- Enable validator by default or not?

Also:
- Not really to decide, but for a follow up: Make it more clear if the edit was saved or not. This ties in with stuff like https://github.com/bolt/core/issues/1372 I think.


bobdenotter (approved):
Hi, @simongroenewolt 

I'm really liking this! I see it has a few small errors in testing, but nothing that should be too hard to fix. Is this something you can tackle? 

Otherwise i can probably get them to pass over the weekend, and we can merge this in!

simongroenewolt:
> I see it has a few small errors in testing, but nothing that should be too hard to fix. Is this something you can tackle?

I think those failures are all because I've added a contenttype to demo the validation.
To fix this there are a couple of options, I'll start by moving the constraint demo contenttype to the end of `contenttypes.yaml` I think that will change 'fix' a number of failing tests because they assume a specific order of contenttypes in the admin interface.

bobdenotter:
(closing and re-opening to trigger a stuck Travis build)

bobdenotter:
Two out of three on Travis. Good enough for me! 

Screenshot 2020-10-31 at 18 12 46

bobdenotter:
About your specific questions: 


> I'm not fond of the names relationsFromThisContent and relationsToThisContent in Content.php -> maybe rename to outgoingRelations and incomingRelations?

I can live with either, to be honest :-)

> The related_values_from_content twig functionality is weirdly named, maybe it would be better to create an option on related_values instead?

I think i prefer it as `related_values_from_content`. It's quirky, but it's explicit, and not used in a ton of places. As an option to `related_values`, it'd become more "hidden". 

> Enable validator by default or not?

Yes. If people don't want to use this functionality, they will most likely simply not add validators to the content-type.

simongroenewolt:
> About your specific questions:
> 
> > I'm not fond of the names relationsFromThisContent and relationsToThisContent in Content.php -> maybe rename to outgoingRelations and incomingRelations?
> 
> I can live with either, to be honest :-)
> 
> > The related_values_from_content twig functionality is weirdly named, maybe it would be better to create an option on related_values instead?
> 
> I think i prefer it as `related_values_from_content`. It's quirky, but it's explicit, and not used in a ton of places. As an option to `related_values`, it'd become more "hidden".
> 
> > Enable validator by default or not?
> 
> Yes. If people don't want to use this functionality, they will most likely simply not add validators to the content-type.

ok, in that case I'll leave evertything as it currently is.

bobdenotter:
@simongroenewolt That took a bit of git-juggling, but it's merged! 🤹

simongroenewolt:
> That took a bit of git-juggling

Is there something I should/shoudn't do in the future to prevent that?

bobdenotter:
@simongroenewolt Nope, not that I can think of. It's just how git works, and that we're keeping a `4.1` and `master` branch in sync, mainly. 

I expect when we've worked with this git-workflow for a bit longer, we'll become more proficient with it. :-)

PR #2068, opened 2020-10-28
Make "files" selector use more levels

bobdenotter:


I-Valchev (approved):
Better!

maybe this should be configurable? @bobdenotter

bobdenotter:
> maybe this should be configurable? @bobdenotter

Nah, can't really think of a reason why one would _not_ want to see files on deeper levels.. We'll consider that, when people start complaining.

I-Valchev:
On second thought, yes, I fully agree.

PR #2072, opened 2020-10-29
Remove extraneous whitespace in excerpts

bobdenotter:
Before, excerpts could look like `This is an excerpt.            With two lines of text.                  Or even mo…`

In HTML this isn't a big deal, but it is when generating - say - SEO / OG tags. 

Edit: Github refuses to render the extra whitespace 🤦


I-Valchev:
Let's merge!

PR #2075, opened 2020-10-29
Fallback to default locale when not explicitly given in URL

I-Valchev:
Hitting `/pages` for example will use the default locale, not the last used one.


bobdenotter (approved):
Jagshemesh! 🏴‍☠️

PR #2082, opened 2020-10-31
Fix: Support standard Symfony `.env` options

### LINKED ISSUE #1973: Support 'standard' Symfony .env options
At the moment Bolt 4 only supports one file for environment variables configuration: .env

Symfony supports overriding/extending .env based on the environment (.env.prod / env.test), and allows additional .local suffix for local changes not meant to be checked in to version control. This is really convenient.

See:

https://symfony.com/doc/current/configuration.html#configuring-environment-variables-in-env-files

https://symfony.com/doc/current/configuration.html#overriding-environment-values-via-env-local

I'd think supporting this wouldn't make anyone's existing config to break, but it will enable lots of developers with Symfony experience to use env files like they are used to.

How can you help?
-------------------------------

I can create a pull request, It looks like this could even be a 1 line change.

bobdenotter:
Fixes #1973 

Pinging @simongroenewolt : This is what you meant, right? 

Changes in here are analogous to what's been done in https://github.com/bolt/project/pull/31


simongroenewolt (commented):
> 
> Pinging @simongroenewolt : This is what you meant, right?
> 

Yes, this is what I meant. 
The changes to .env and .editorconfig are not related are they? Specifically: is .editorconfig in line with the current coding standards in bolt/core?

.env
```diff
@@ -1,42 +1,63 @@
...
+# https://symfony.com/doc/current/best_practices.html#use-environment-variables-for-infrastructure-configuration
+
 ###> symfony/framework-bundle ###
 APP_ENV=dev
 APP_DEBUG=1
-APP_SECRET=73fd814e2c9fbf61dc5f87a62967d829
-#TRUSTED_PROXIES=127.0.0.1,127.0.0.2
-#TRUSTED_HOSTS='^localhost|example\.com$'
+APP_SECRET=!ChangeMe!
+TRUSTED_PROXIES=127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
```
toofff, line 20: This is linked to the docker configuration, is it necessary to take it back to the `bolt/core` repository?`10.0.0.0/8,172.16.0.0/12,192.168.0.0/16`

.env
```diff
@@ -1,42 +1,63 @@
...
+#TRUSTED_HOSTS='^(localhost|nginx)$'
```
toofff, line 21: This is linked to the docker configuration, is it necessary to take it back to the `bolt/core` repository? => `nginx`

public/index.php
```diff
@@ -3,38 +3,30 @@
...
+if ($_SERVER['APP_DEBUG']) {
     umask(0000);
 
     Debug::enable();
 }
 
-if (! empty($_SERVER['TRUSTED_PROXIES'])) {
-    Request::setTrustedProxies(explode(',', $_SERVER['TRUSTED_PROXIES']), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
+if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? false) {
+    Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
```
toofff, line 22: Warning, with Docker if we want everything to work correctly, it has to be like this

`Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL);`

.editorconfig
```diff
@@ -1,8 +1,92 @@
...
+
+[*.php]
+indent_style = space
+indent_size = 4
+
+[*.sh]
+indent_style = tab
+indent_size = 4
+
+[*.vcl]
```
toofff, line 41: remove, because no varnish configuration in the `bolt/core` repository.

.env
bobdenotter: I copied this from `bolt/project`, figuring it'd make sense to keep them the same. What do you think this line should say?

.env
bobdenotter: Same here. What do you suggest? Why not keep it the same?

public/index.php
bobdenotter: This is the same as in https://github.com/bolt/project/blob/master/public/index.php 

Does Docker break on the `^ Request::HEADER_X_FORWARDED_HOST` part?

.editorconfig
bobdenotter: Ok, will do.

.env
toofff: I just corrected that in this [pull request](https://github.com/bolt/project/pull/47/files)

.env
toofff: I just corrected that in this [pull request](https://github.com/bolt/project/pull/47/files)

.editorconfig
toofff: I just corrected that in this [pull request](https://github.com/bolt/project/pull/47/files)

public/index.php
toofff: I just corrected that in this [pull request](https://github.com/bolt/project/pull/47/files)

Because by default the project works with Composer and the SQLite database.
With hindsight I think I removed it not exactly for docker but to add a varnish configuration to the project and that's when I had to remove this part.

Since I haven't added Varnish yet, there is obviously no point in removing it.

PR #2093, opened 2020-11-03
Correctly instantiate fields in prod

### LINKED ISSUE #2030: Bolt's FieldRepository should not assume all classes are loaded
Since #1734 has a large discussion that seems to cloud the actual problem, I decided to open a new issue. But it is effectively a duplicate.

We had the same issue... Diving very deep in the Bolt-internals, I found the cause.

It is related to caching, but not the cache specifically.

https://github.com/bolt/core/blob/b2aba74de9ca50d392e38ac77f3881f83e319e23/src/Repository/FieldRepository.php#L134

With that `get_declared_classes`-call it is assumed all Field-classes are loaded at this time. This happens to be true when the cache is not yet defined, probably due having to create the entity/field-configuration. But is an incorrect assumption when the cache is used. 

In our case, the amount of declared classes at that time drops from 2270 to 1140.
The only Field-class during that call to get_declared_classes is the generic Field-class, none of its subclasses is available. 

As such, this has incorrect behavior for all fields, not just the TemplateselectField. The TemplateselectField's Twig-template just happens to use helper methods with typehints. And those cause fatal errors.

A quick-fix for Bolt-specific fields would be to first test wheter `'Bolt\\Entity\\' . $classname` exists (and implements that interface). But that won't work for fields from outside that namespace.

I-Valchev:
Fixes #2030 

But it runs on every request.

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -8,6 +8,7 @@ use Bolt\Configuration\Parser\ContentTypesParser;
 use Bolt\Configuration\Parser\TaxonomyParser;
 use Bolt\Extension\ExtensionCompilerPass;
 use Bolt\Extension\ExtensionInterface;
+use Bolt\Repository\FieldRepository;
 use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
 use Symfony\Component\Config\FileLocator;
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
Let's roll with this.. Merging into `4.1` as well as `master`

PR #2094, opened 2020-11-03
Fix reference to 404 html template

I-Valchev:


bobdenotter:
Why is this needed? It should work as-is, right? And if it's needed, we should also do the same for `maintenance:`, `forbidden:` and `internal_server_error:` as well.

I-Valchev:
@bobdenotter you are right, it is not needed. I was confused that it did not work on project with a custom template, but after discussing it with you and verifying, I saw it was my mistake.

Let's close this!

Issue #2079, opened 2020-10-30
While editing page/entry/..<ContentType> with no image set, 404 response in the background


Wieter:
Not breaking, but some room for a small improvement.

While setting up a new website, and editing pages/entries/(doesn't really matter what), I noticed a 404 response in the console:
`GET https://.../bolt/edit/null 404 (Not Found)`
This I traced back to an empty image field (there was no 'Highlighted image' for the ContentType which is being edited, which is correct)
The preview image field is templated to become the following:
`<a class="editor__image--preview-image" style="background-image: url('null');"></a>`
Probably a check need to be made if 'null' is returned, and then just omit setting the background-image. 

I haven't dissected yet where the template is rendered, will do so in due time :)

Details
-------

| Question                      | Answer
|----------------- | ------
| Relevant Bolt Version | 4.1 
| Install type                   |  Composer install
| BC Break                     | no
| PHP version                 | 7.2
| Web server                  | Apache, version 2.4.46
| For UX/UI issues        | N/A

Reproduction
------------

### Bug summary

404 for pages/entries/.. without highlighted image in editor

### Steps to reproduce

* Install BOLT the usual way, using PostgreSQL as backend
* open developer console of the browser, and watch the traffic flow
* edit page/entry/whatever... that does not have set a highlight image 

### Expected result

No unnecessary 'chitchat' with the server when no image is present

### Actual result

404 in the background, trying to resolve 'null'

Wieter:
Some findings:
https://github.com/bolt/core/blob/6bf1db6f7994193b6c90622fe5d127388582969a/assets/js/app/editor/Components/Image.vue#L126 previewImage is checked for empty string, but further down `thumbnailImage` is used.
https://github.com/bolt/core/blob/6bf1db6f7994193b6c90622fe5d127388582969a/assets/js/app/editor/Components/Embed.vue#L87 no check is embedded to test if this field might be empty, then if empty `:style="`background-image: url('${previewImage}')`"` could cause this 404 while trying to retrieve null.

I-Valchev:
hi @Wieter , nice little catch! Would you like to create a small PR to fix this, so that you also show up as a bolt/core contributor? :-)

Wieter:
Sure, I have to work on some other fixes as well (postgres related).

PR #2097, opened 2020-11-03
Fix 404 emptyimage

Wieter:
Fix for #2079


Wieter:
WIP because there is another minor fix, but I'm a bit stuck there: 
while accessing http://../bolt/edit/54#media:
https://github.com/bolt/core/blob/50e6f1baf283df112a820d97fbaea6d536079848/assets/js/app/editor/Components/Select.vue#L30 `empty: props.option.value == ''` --> TypeError: Cannot read property 'value' of undefined

And another which I've not yet been able to trace back (0.js:formatted:46090), seems to be all related.

bobdenotter (approved):
Nice catch! Thanks, @Wieter

Wieter:
You're welcome!

https://github.com/bolt/core/blob/50e6f1baf283df112a820d97fbaea6d536079848/assets/js/app/editor/Components/Select.vue needs some additional attention, I leave that for a new issue/pull.

I-Valchev (approved):
Again, nice catch @Wieter .

Let's merge it in! Also merged into 4.1

PR #2108, opened 2020-11-05
Configure extension config on install and update + more output when something fails

### LINKED ISSUE #2002: Installing an extension through composer no longer copies default config
Before, when we ran `composer require bolt/forms` for example, we'd get the default config the first time _only_.

But right now, it does not do it anymore. Could it be the recent changes to `composer.json` involved with the CI Improvements? @bobdenotter  @TomasVotruba

I-Valchev:
Fixes #2002 + it provides output if a command actually fails.


bobdenotter (approved):
Neat! 👍

PR #2112, opened 2020-11-07
Rename guesstimatePublicFolder() to getPublicFolder() 

simongroenewolt:
Rename guesstimatePublicFolder() to getPublicFolder() and make the function protected. This way subclasses of Kernel can provide their own implementation, for example returning a hard-coded path.


bobdenotter (approved):
Looking good! Thanks, @simongroenewolt 👍

PR #2114, opened 2020-11-08
Refactor add & edit user to symfony forms

LordSimal:
Hello my dear friends 😄  This gonna take a bit.

As already decided in #1976 here is another part of the great "Symfony Forms Refactoring"

## What did I do

This time I will just go from file to file and try to explain why there were changes.

~~`assets/js/app/editor/Components/Select.vue` had changes because in the current state the Vue Select Element always wraps the `value` inside an Array even though it is definitely not need on "not multiple selects".
But e.g. the `locale` field only allows to be a single value and Symfony Forms wouldn't recognise that value in an array.~~

I hope this doesn't break fields which use this as a base as well, but we will see. (UPDATE: yep it did, see bellow)

`assets/scss/init/_base.scss` is just styling for the new "inline errors" which would look like that.

image

## src/Controller/Backend/UserEditController.php
well here we go...

### Add a separate function and route for adding a user

In the current state BoltCMS handles adding users via the route `/user-edit/0` and the same function as editing a user.
Now there is an extra route `/user-edit/add` which handles that. With this the logic for adding and editing is now split (besides the function `_handleValidFormSubmit` which is used by both add and edit)

And now there is also a separate Event for `UserEvent::ON_ADD`

### Validation and outputting of Form is handled via Symfony Form

There are basically 3 main parts for how the form is being brought into TWIG

```
$form_data = [
    'suggested_password' => $suggestedPassword,
    'roles' => $roles,
    'require_username' => true,
    'require_password' => true
];
$form = $this->createForm(UserEditType::class, $user, $form_data);
```

The `$form_data` contains all the necessary data which is used to build the data in the Symfony Form.
`UserEditType::class` is just the Class which defines which fields should be used for the form
Due to the fact that I connected the User Entity Class with the Form we have to pass on the entity `$user` (which in the `edit` form is the selected user or in the `add` form the generated user from `UserRepository::factory();`

And nearly at the end of the function I pass that `$form` into the view with
```
return $this->render('@bolt/users/add.html.twig', [
    'userForm' => $form->createView()
]);
```

The whole validation process is now being started via `$form->isSubmitted() && $form->isValid()`. How these validations are being set can be seen in the Annotations set in `src/Entity/User.php`

## src/Entity/User.php

Since there is no "easy" way to disable validation for specific fields in Symfony Forms I had to add `Validation Groups`
These groups are being used to basically say `when should this Assertion be applied`

See later in the `UserEditType.php` how this is being used
Based on https://symfony.com/doc/current/validation/groups.html

### Special array structure required by Symfony forms

In the `buildForm` function you first see 2 foreach loops which create custom array structures for Roles and Locales.
The reason behind that is the fact, that Symfony Forms require the array structure to be defined like

```
'choices'  => [
    'Maybe' => null,
    'Yes' => true,
    'No' => false,
],
```

So the `key` is actually the *Label* and the `value` is actually what is used to save the value... which is exactly the opposite of what I am used to and what the template `@bolt/_partials/fields/select.html.twig` uses. You can see that change later in the `templates/users/_form.html.twig` as well.

See https://symfony.com/doc/current/reference/forms/types/choice.html#example-usage

### Set suggested password as attribute

Setting the suggested password was only possible as `attribute`

```
'attr' => [
    'suggested_password' => $options['suggested_password']
],
```

 which is then visible in Twig via `userForm.plainPassword.vars.attr.suggested_password`

### More Symfony Form Configuration

Via the `configureOptions` function there are many things you can set for this form.

```
$resolver->setDefaults([
    'data_class' => User::class,
    'suggested_password' => '',
    'roles' => '',
    'require_username' => '',
    'require_password' => '',
    'validation_groups' => function (FormInterface $form) {
        /** @var User $entity */
        $entity = $form->getData();

        $validation_group = "edit_user";
        if($entity->isNewUser()){
            $validation_group = "add_user";
        } else if($entity->getPlainPassword() === '') {
            $validation_group = "edit_user_without_pw";
        }

        return $validation_group;
    },
]);
```

`'data_class' => User::class,` connects this form with the User Entity and therefore fields that have exactly the same name as fields in the User Entity Class will be saved automatically in the user entity and have the same assertions (as already seen above with the `Validation Groups`)

`suggested_password`, `roles`, `require_username` and `require_password` do need to be defined here, otherwise you can't pass data from the controller in the form

`validation_groups` now this callback defines which `Validation Groups` to use for the current instance of the form.
So in there I check wether or not we are on a
-  "Add User" form
- "Edit User" form with a filled out password
- "Edit User" form without a filled out password

The last 2 are different because we wan't to allow users to change their user data without re-entering their password everytime.

## templates/users/_form.html.twig

This is the whole Twig representation of the form which is used by the `add.html.twig` and `edit.html.twig` template

## templates/users/add.html.twig 

Is basically a copy of the already present `edit.html.twig` template with just some adjustments to show `Add User` at the top etc.

## To summarize

- ~~Had to adjust `assets/js/app/editor/Components/Select.vue` because of single select values not being "correct"~~
- Added `src/Form/UserEditType.php` to handle fields for the add and edit form
- Adjusted `src/Controller/Backend/UserEditController.php` to use the new Symfony Form
- Added a separate `/user-edit/add` route to handle adding users beside the already present `/user-edit/edit` route
- Adjusted assertions in `src/Entity/User.php` to allow editing of user without password entry
- Refactored Twig template to have separate `add.html.twig` and `edit.html.twig` but a common `_form.html.twig`
- Adjusted `templates/users/listing.html.twig` to use new `bolt_user_add` route

I hope I didn't forget anything but could be possible. If so just ask away 😄


LordSimal:
As I just saw my Vue change for the assets/js/app/editor/Components/Select.vue broke other things.
So I have to find a way to do what I need to do without adjusting that

LordSimal:
For some reason the id for the `_form.html.twig` needs to be `editcontent` but other then that the last comments were mainly adjusting the tests to the new symfony inline error messages which replaced the notification bubble in the bottom left.

I am pretty happy that I got that working now 😄

assets/js/app/editor/Components/Select.vue
```diff
@@ -7,7 +7,7 @@
             :limit="1000"
             :multiple="multiple"
             :options="options"
-            :searchable="searchable || taggable"
+            :searchable="taggable"
```
I-Valchev, line 10: Hi @LordSimal , first of all thanks for the great detailed PRs. It's a pleasure to read through the explanations and code! Keep up the awesome work!

This is something that was changed recently, to make Relations searchable. I assume we need to rebase? Or, alternatively, is it GitHub not updating the file changes after a rebase?

src/Controller/Backend/UserEditController.php
```diff
@@ -43,44 +43,117 @@ class UserEditController extends TwigAwareController implements BackendZoneInter
...
-        if (! $user instanceof User) {
-            $user = UserRepository::factory();
-            $suggestedPassword = Str::generatePassword();
-        } else {
-            $suggestedPassword = '';
+        $user = UserRepository::factory();
+        $submitted_data = $request->get('user_edit');
+
+        $event = new UserEvent($user);
+        $this->dispatcher->dispatch($event, UserEvent::ON_ADD);
```
I-Valchev, line 73: Great idea to add more user events!

Wonder if, for the sake of consistency with `ContentEvent`, we should have `PRE_SAVE` and `POST_SAVE` somewhere? Just a thought, maybe it's out-of-scope for this PR.

src/Entity/User.php
```diff
@@ -114,7 +114,7 @@ public function getId(): int
         return $this->id;
     }
 
-    public function setDisplayName(string $displayName): void
+    public function setDisplayName(?string $displayName): void
```
I-Valchev, line 117: Should we call `setDisplayName` with null at all?

src/Entity/User.php
```diff
@@ -134,7 +134,7 @@ public function getUsername(): string
         return $this->username;
     }
 
-    public function setUsername(string $username): void
+    public function setUsername(?string $username): void
```
I-Valchev, line 137: Likewise, should we call `setUsername` with null?

src/Entity/User.php
```diff
@@ -314,4 +314,9 @@ public function toArray(): array
     {
         return get_object_vars($this);
     }
+
+    public function isNewUser(): bool
+    {
+        return $this->id === null;
```
I-Valchev, line 320: In the Field, we check if a field is new like so: https://github.com/bolt/core/blob/577f5b9fff286c7a6ffd81553bfb903e8d911e58/src/Entity/Field.php#L206

Do we have to update that here, or there? Given the strict comparison operator (`===`), I wonder if both cases work?

src/Form/UserEditType.php
```diff
@@ -0,0 +1,116 @@
...
+        $this->localeHelper = $localeHelper;
+        $this->twig = $twig;
+    }
+
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        // Create custom role options array
+        $roleOptions = [];
+        $custom_roles = $options['roles'];
+        foreach ($custom_roles as $roleName => $roleHierarchy):
```
I-Valchev, line 39: Ha! It's been some time since I've seen this version of the `foreach` loop. I'm surprised it even passes the tests 🙈

assets/js/app/editor/Components/Select.vue
LordSimal: Just refreshed the branch with current master but had to add that manually since some of my commits had changes in there.

src/Entity/User.php
LordSimal: I had problems with a test where the displayName was " " so just a whitespace.
Due to the trim functionality of Symfony Forms the " " will become a null and that produced a type error because string !== null

src/Entity/User.php
LordSimal: same as displayName

src/Form/UserEditType.php
LordSimal: why shouldn't such a foreach pass the test?
I write all my loops like that because I don't like that JS like .map() style or whatever.

src/Entity/User.php
I-Valchev: Ah right! Thanks for explaining this to me 🤗

src/Controller/Backend/UserEditController.php
LordSimal: Love that Idea, will add that right now.
Since add and edit both use the private function `_handleValidFormSubmit` I will put them in there.

src/Entity/User.php
LordSimal: well in my case I can say, that creating a new user produces a $user->id of null, not 0
Thats why I used that here.
Could be that the Field Entity is handeld differently so a new field has an ID of 0 before it is persisted

src/Controller/Backend/UserEditController.php
I-Valchev: Awesome 😎

src/Entity/User.php
I-Valchev: Ah, yes you are correct.

https://github.com/bolt/core/blob/2d0a6017ad385c4842d0431b235a78ca36933c23/src/Entity/Field.php#L51

By the looks of it, we explicitly set it to 0 for the field 🤷‍♂️

All is good then. 👍

src/Form/UserEditType.php
I-Valchev: I meant this foreach:
```
foreach:
  ...
  endforeach;
```

As opposed to:

```
foreach ($foo => $bar) {

}
```

assets/js/app/editor/Components/Select.vue
I-Valchev: Nice, thanks!

src/Form/UserEditType.php
LordSimal: Ah, I see.
I like to use types of "endings" because when you nest if, foreach etc. and you have a bunch of closing `}` at the end next to each other you don't know by a quick glance where you can add e.g. an else statement.

LordSimal: I know it is more to write but I like more sourcecode to be more structured and readable than "compact"
Thats basically why I do that

bobdenotter: I'd prefer these to be in this format: 

```php
foreach (…) { 

}
```

Not because I think one is better than the other, but because of consistency. Also, PSR-12 says so: https://www.php-fig.org/psr/psr-12/#55-foreach

I'm surprised it passes tests, though. 🤔

LordSimal: alright, then I will adjust that

LordSimal: actually I just saw how the ecs changed that foreach.
I didn't see that till now that only the first line of code was indented.
Now with the correct { and } it looks good.

LordSimal:
As just discussed in Slack "FredXD" will fork my branch and add the possibility to add a profile picture to the new `UserEditType` form.

Fredxd:
Thanks @LordSimal and @bobdenotter to let me contribute on Bolt Core.
We have add on this PR 

- Possibility to upload avatar in user (add/edit)
- Possibility to change configuration (upload path, default avatar, extensions allowed)

Feel free to comment, review or suggest changes

LordSimal:
I will add the refactor for the `edit profile` functionality here as well so that `/bolt/profile-edit` will use the `UserEditController` as well and doesn't have its own logic.
Will try to do that tomorrow so that we can probably merge that in till sunday and add that to the incoming 4.2 release

LordSimal:
as can be seen in the last commit I did some more adjustments. These contain:

- Removing the `ProfileController.php` and merging that functionality in the already present `UserEditController.php`
- Removing not used `src/Form/UserType.php` and renaming the `src/Form/UserEditType.php` to `src/Form/UserType.php` so the Form Type has the same name as the entity it is using/adjusting
- Therefore had to adjust all all occurrences of `user_edit` to `user` in the tests
- Removed the fields `roles` and `status` from the `profile-edit` route so future implementations of `NON_ADMIN` profile editing can use this. Currently only admin users can edit profiles therefore this can't be tested yet.

LordSimal:
And of course thanks to @Fredxd who added the `avatar` field to the symfony form 👍

bobdenotter:
Hey @LordSimal and @Fredxd! Loving these additions! Let me know when it's ready, so we can merge it in!

LordSimal:
Basically I am done with my adjustments and as far as I know Fredxd as well.

bobdenotter (approved):
Ok, looks good at first glance, and tests are passing as well! 

Let's merge in, and iron out any kinks (if any) as we move towards 4.2 stable!! 🎉 🎊

bobdenotter:
Quick note: when trying this out, be sure to update the DB: 

```
bin/console doctrine:schema:update --force
```

LordSimal:
I always did `make db-reset` to be extra sure

PR #2118, opened 2020-11-10
Bugfix incorrect `|current` filter for localized menu items

I-Valchev:
Fixes issue with localized menu URI's not recognised as current. Credits to @nestordedios for finding and debugging this!

Also fixes issue with using `record.slug`. It used to return the _wrong_ locale.


bobdenotter (approved):
Thanks, @I-Valchev and @nestordedios

bobdenotter:
I have cherry-picked this into `4.1`!

Issue #2109, opened 2020-11-06
viewless setting doesn't disable list view


simongroenewolt:
When you create a content type and set `viewless` to `true` I expect both the 'single' view and the 'listing' view to be unavailable. At the moment the listing view is still registered when you set `viewless` to `true`.

Details
-------

| Question                                              | Answer
|------------------------------------------------------ | ------
| Relevant Bolt Version                                 |  4.0 / 4.1 / master
| Install type                                          |  Composer install
| <abbr title="Backwards Compatibility">BC</abbr> Break | no
| PHP version                                           | 7.2 / 7.3 / 7.4 |
| Web server                                            | Apache / Nginx / Built-in, version [ version ]

Reproduction
------------

Create a content type with viewless true

```
promos:
    name: Promos
    singular_name: Promo
    fields:
        heading:
            type: text
        subheading:
            type: textarea
        body:
            type: textarea
    viewless: true
```

If you now visit http://localhost:8088/promos you'll see a list view.

simongroenewolt:
`DetailController.php` uses `TwigAwareController->renderSingle()` which does a check on viewless. `ListingController.php` uses `TwigAwareController->render()` and that call cannot check for viewless because it doesn't 'know' anything about the data that will be shown.

I don't really understand the separation of the logic for the single view between TwigAwareController and DetailController. For the ListingController it looks like the only solution would be to put the check in in that controller itself.

I-Valchev:
That _does_ look like a bug.

From what I can see, it can be fixed by checking for a `viewless` contentType once we fetch the definition in `ListingController.php` here:

https://github.com/bolt/core/blob/577f5b9fff286c7a6ffd81553bfb903e8d911e58/src/Controller/Frontend/ListingController.php#L47

Would you like to create a PR about it? :-)

simongroenewolt:
@I-Valchev I can make a PR, no problem. I do have a slightly uneasy feeling about one check for it being in the parent class, (TwigAwareController) and the other in one of the children (ListingController). But, as the check for the listing should definitely be done in the ListingController it is ok to start with only that change.

I-Valchev:
@simongroenewolt yes, I see what you mean there. Like you say, we have `renderSingle` in TwigAwareController, but there's no `RenderListing` (yet?). For now, that will do. In future, if we refactor, definitely I agree we can revisit it.

Thanks again!

PR #2122, opened 2020-11-11
return 404 for listing route of contenttype that has viewless: true

simongroenewolt:
Fixes https://github.com/bolt/core/issues/2109


I-Valchev (approved):
👍  Awesome, thank you for the PR!

Fixes #2109

bobvandevijver:
@I-Valchev I would have expected this to also be merged into 4.1, as this is also a bug in that version. Is that something that can still be done?

I-Valchev:
Hi @bobvandevijver, yes I think it's possible to backport it to 4.1 :-) I'll do that tomorrow and ping you when it's merged in.

PR #2123, opened 2020-11-11
Fix/collection ordering

simongroenewolt:
Fixes https://github.com/bolt/core/issues/2115 - only checked in firefox at the moment, so marked draft.


bobdenotter (approved):
I've checked it out, and tested it on Firefox, Chrome and Safari. Seems to work just fine! 👍

PR #2126, opened 2020-11-14
Login: redirect to dashboard, if authenticated

BlueM:
When requesting the login page (e.g. via a bookmark), even an already authenticated user is presented the login dialog.
This commit improves that behavior by sending authenticated users to the dashboard.


BlueM:
Sorry, my local code was not up to date for some reason. Just noticed that this behavior is already implemented.

bobdenotter:
Thanks, regardless, @BlueM ;-)

PR #2128, opened 2020-11-14
add default value for $extensionMenus so bolt won't break when configuration hasn't been updated yet.

simongroenewolt:
The new support for extensions to insert menu items requires a change to the services.yaml - with a default set for the injected $extensionMenus the menu items will not appear when the configuration change hasn't been made, but it will allow the the site to keep working instead of throwing a big error.

### DIFF
--- a/src/Menu/BackendMenuBuilder.php
+++ b/src/Menu/BackendMenuBuilder.php
@@ -42,7 +42,7 @@ final class BackendMenuBuilder implements BackendMenuBuilderInterface
 
     public function __construct(
         FactoryInterface $menuFactory,
-        iterable $extensionMenus,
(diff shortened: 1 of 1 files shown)


simongroenewolt:
PR is ready for review, I've thought of ways to signal the user that something needs to change in the configuration but I couldn't come up with a method that would make sense. I've considered: 

1. logging a deprecation -> but there are so many deprecations it won't be visible, and https://github.com/symfony/deprecation-contracts is not in use anywhere in the bolt/core code.

or

2. logging a warning using the Logger interface -> also not very common in bolt/core code.

So for now there won't be a message.

bobdenotter (approved):
Thanks, @simongroenewolt 👍

PR #2131, opened 2020-11-16
Feature: Reset password from the login screen.

bobdenotter:
This was in a separate branch. Let's get it merged, so it can go into 4.2!


config/bolt/config.yaml
```diff
@@ -199,3 +199,11 @@ user_avatar:
     upload_path: avatars
     extensions_allowed: ['png', 'jpeg', 'jpg', 'gif']
     default_avatar: '' # Put path of file locate in files directory
+
+# Settings for the reset password logic
+reset_password_settings:
+    show_already_requested_password_notice: true
+    mail_from: "do-not-reply@dummy.com"
```
simongroenewolt, line 206: I think you shouldn't use a 'fake' domain that exists and isn't explicitly documented as usable in these kind of configurations (e.g. https://en.wikipedia.org/wiki/Example.com), in case someone doesn't change this mail bounces could actually be delivered to this address and might include sensitive data. You could use example.com/example.org but that is more intended for documentation, maybe do-not-reply@localhost or do-not-reply@invalid is a better option)

assets/scss/layout/_reset-password.scss
```diff
@@ -0,0 +1,48 @@
+.page--reset-password,
```
simongroenewolt, line 1: A question for the other reviewers, existing css in bolt seems inspired by http://getbem.com/naming/ or a similar scheme, I haven't seen it documented explicitly anywhere. Is it a requirement? (I think that might make sense, but I'm not certain I've followed it myself...)

src/Form/ChangePasswordFormType.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('plainPassword', RepeatedType::class, [
+                'type' => PasswordType::class,
+                'first_options' => [
+                    'constraints' => [
+                        new NotBlank([
+                            'message' => 'Please enter a password',
```
simongroenewolt, line 25: Are these messages translate/translatable? They probably should be.

src/Form/ResetPasswordRequestFormType.php
```diff
@@ -0,0 +1,32 @@
...
+
+class ResetPasswordRequestFormType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('email', EmailType::class, [
+                'constraints' => [
+                    new NotBlank([
+                        'message' => 'Please enter your email',
```
simongroenewolt, line 21: Same here for the translation.

symfony.lock
```diff
@@ -420,7 +420,7 @@
         "version": "2.0.1"
     },
     "php": {
-        "version": "7.2"
+        "version": "7.3"
```
simongroenewolt, line 423: Should this change be part of this PR?

assets/scss/layout/_reset-password.scss
LordSimal: i basically copied that from https://github.com/bolt/core/blob/master/assets/scss/layout/_login.scss
But I can confirm that the whole SCSS naming scheme can be improved. Would rather handle that in a separate PR though

config/bolt/config.yaml
LordSimal: didn't know that, good to know. bob, can you handle that please since you created that PR from the internal branch?

LordSimal: otherwise you can leave it as is and I will create a PR after this has been merged

src/Form/ChangePasswordFormType.php
LordSimal: Seems like I overlooked that. I will fix these messages after this PR has gone through.

src/Form/ResetPasswordRequestFormType.php
LordSimal: Same answer as above

symfony.lock
LordSimal: I guess this happened because my local environment was using PHP 7.3 while I developed that

templates/security/login.html.twig
```diff
@@ -22,13 +22,17 @@
         <div class="login__logo">
             <img class="logo" alt="Bolt CMS logo"
                 {% if not config.get('general/omit_backgrounds') %}
-                 src="{{ asset('assets/images/bolt_logo_loginpage--bg.svg', 'public') }}">
+                 src="{{ asset('assets/images/bolt_logo_loginpage--bg.svg', 'bolt') }}">
```
I-Valchev, line 25: @bobdenotter is this correct?

templates/security/login.html.twig
```diff
@@ -22,13 +22,17 @@
...
                 {% else %}
-                 src="{{ asset('assets/images/bolt_logo_loginpage.svg', 'public') }}">
+                 src="{{ asset('assets/images/bolt_logo_loginpage.svg', 'bolt') }}">
```
I-Valchev, line 27: @bobdenotter same as the one above?

I-Valchev:
Very, _very_ nice feature! 

@LordSimal great work. Wonder if you can create a PR against `feature/reset-password`, so that when we merge this one into master, all the changes are present? The point being, when bigger PRs like this one are merged and we need to go back and check how it was done, it is easier to track and follow.

symfony.lock
bobdenotter: Fixed this

templates/security/login.html.twig
bobdenotter: Yes, this doesn't hurt. Currently `public` and `bolt` are the same, but a project _could_ overwrite `public`, potentially.

LordSimal: As I remember I had to do that locally so the logo was showing up correctly.

config/bolt/config.yaml
bobdenotter: I've changed it to `do-not-reply@example.org`

bobdenotter:
I removed the `[WIP]` and lights are green. Let's merge in! 🚂 🚋 🚋

Issue #2132, opened 2020-11-17
New Feature: add keys setting for select field type


denis-gorin:
As it was in Bot3 (updated)

> If you wish to store another field or value from the original ContentType in your database, use the `mode: format` and set the `format` settings . If you do this, it will not store the 'id', but the formatted value of the field(s) you specify. For example:

``` yaml
vendor:
    type: select
    values: vendors/title
    mode: format
    format: '{title} - {sub_title}'
```

This is good in combination with `title_format` in ContentTypes settings for concretize listing elements in dashboard

``` yaml
title_format: '{vendor}: {title}'
```

![image](https://user-images.githubusercontent.com/3963911/99341368-5336e000-28bc-11eb-95bb-12fa9caaef9f.png)

I-Valchev:
hi @denis-gorin , just to make sure I understand this: instead of saving an ID which we can link to a record, this stores a _field value_ from the selected record, right? Would that mean that it can't be used to retrieve the record afterwards? Not that it is a bad thing, I want to make sure that I understand how it does. We may have to adjust the `|selected` filter to make sure it doesn't break.

PR #2134, opened 2020-11-17
add keys setting for select field type

denis-gorin:
As it was in Bolt3...

Implements #2132

### DIFF
--- a/src/Twig/ContentExtension.php
+++ b/src/Twig/ContentExtension.php
@@ -618,14 +618,13 @@ class ContentExtension extends AbstractExtension
 
         /** @var Content[] $records */
         $records = iterator_to_array($this->query->getContent($contentTypeSlug, $params)->getCurrentPageResults());
-        $keyName = $field->getDefinition()->get('keys');
 
         foreach ($records as $record) {
-            if ($keyName && $record->hasField($keyName)) {
-                $keyValue = $this->contentHelper->get($record, "{{$keyName}}");
+            if ($field->getDefinition()->get('mode') === 'format') {
+                $formattedKey = $this->contentHelper->get($record, $field->getDefinition()->get('format'));
             }
             $options[] = [
-                'key' => $keyValue ?? $record->getId(),
+                'key' => $formattedKey ?? $record->getId(),
                 'value' => $this->contentHelper->get($record, $format),
             ];
         }
--- a/src/Twig/ContentExtension.php
+++ b/src/Twig/ContentExtension.php
@@ -618,10 +618,14 @@ class ContentExtension extends AbstractExtension
 
         /** @var Content[] $records */
         $records = iterator_to_array($this->query->getContent($contentTypeSlug, $params)->getCurrentPageResults());
+        $keyName = $field->getDefinition()->get('keys');
 
         foreach ($records as $record) {
+            if ($keyName && $record->hasField($keyName)) {
+                $keyValue = $this->contentHelper->get($record, "{{$keyName}}");
+            }
             $options[] = [
-                'key' => $record->getId(),
+                'key' => $keyValue ?? $record->getId(),
                 'value' => $this->contentHelper->get($record, $format),
             ];
         }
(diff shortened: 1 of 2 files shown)


I-Valchev:
hi @denis-gorin , just to make sure I understand this: instead of saving an ID which we can link to a record, this stores a field value from the selected record, right? Would that mean that it can't be used to retrieve the record afterwards? Not that it is a bad thing, I want to make sure that I understand how it does. We may have to adjust the |selected filter to make sure it doesn't break.

denis-gorin:
> instead of saving an ID which we can link to a record, this stores a field value from the selected record, right? 

right, this works exactly as in previous Bolt3 version

> Would that mean that it can't be used to retrieve the record afterwards?

Sure, this case has some limitations (conventions) why user will be using it and how... 
For my case I have bundle of records with unique titles (or slugs, or any other field for ex.) and I know I can get corresponding record by this key (ex. title) using where condition. 

But again in my case I do not need to get the record afterwards in my templates, I just need only string (title) - and I got it with keys setting.

And on the other hand having keys setting helps me to better visualize records in a dashboard (showing me records relations in listings)

> We may have to adjust the |selected filter to make sure it doesn't break

Sorry, I don't have an answer for now. I need to have some research about |selected filter. So, I went to look for where it is.

I-Valchev:
Hi @denis-gorin , thanks for the explanation!

Here's the `|selected` filter
https://github.com/bolt/core/blob/24b00053473a05e317a67942da95a46ecf57fd47/src/Twig/FieldExtension.php#L44

Then, the actual function is a bit lower:
https://github.com/bolt/core/blob/24b00053473a05e317a67942da95a46ecf57fd47/src/Twig/FieldExtension.php#L95

It would be nice to differentiate between this use case for the SelectField and a _standard_ SelectField using the `mode`. To give you an example, we use the `mode` config to differentiate between Float and Integer fields. 

Then, the definition can look like this:

```
booktitle:
    type: select
    mode: format
    format: '{title} - {subtitle}'
```

Let me explain what changes I suggest:

1. Use `mode` to differentiate between this type of select and the other.
2. Instead of `keys`, use `format` for consistency with things like `title_format` and `fixture_format`
3. The `format` option accepts the entire format, not just the field name. So, for example, the call to `ContentHelper` would look like this:

```
$this->contentHelper->get($record, $field->getDefinition()->get('format'));
```

Does that make sense to you as well?

bobdenotter:
I agree with @I-Valchev here: I'm also missing this specific piece of functionality from Bolt 3, but I also know we've made a conscious decision to _only_ store the ids in the DB.. 

What Ivo suggests would be a nice interim fix, though!

denis-gorin:
> Hi @denis-gorin , thanks for the explanation!
> 
> Here's the `|selected` filter
> 
> https://github.com/bolt/core/blob/24b00053473a05e317a67942da95a46ecf57fd47/src/Twig/FieldExtension.php#L44
> 
> Then, the actual function is a bit lower:
> 
> https://github.com/bolt/core/blob/24b00053473a05e317a67942da95a46ecf57fd47/src/Twig/FieldExtension.php#L95
> 
> It would be nice to differentiate between this use case for the SelectField and a _standard_ SelectField using the `mode`. To give you an example, we use the `mode` config to differentiate between Float and Integer fields.
> 
> Then, the definition can look like this:
> 
> ```
> booktitle:
>     type: select
>     mode: format
>     format: '{title} - {subtitle}'
> ```
> 
> Let me explain what changes I suggest:
> 
> 1. Use `mode` to differentiate between this type of select and the other.
> 2. Instead of `keys`, use `format` for consistency with things like `title_format` and `fixture_format`
> 3. The `format` option accepts the entire format, not just the field name. So, for example, the call to `ContentHelper` would look like this:
> 
> ```
> $this->
> Does that make sense to you as well?

denis-gorin:
Hi I,ve just been added `mode: format` in select field in second commit, hope it is good now

bobdenotter (approved):
Looks good, @denis-gorin! Thanks! 👍

denis-gorin:
I've lost this changes in Patch release 4.1.7.1
ehqt should I do? re-PR?

bobdenotter:
@denis-gorin Since this is a new feature, it'll be included in the upcoming `4.2`, but not in `4.1.x` releases that are mostly bugfixes. It's still in `master` branch` -> 720cf7c9b770a133

PR #2136, opened 2020-11-17
Lookup slugs from record IDs in Twig templates

benr77:
This extension makes it easy to create links in Twig templates that use slugs, but based on the record ID.

Slugs may change over time, so hard-coding them in templates is undesirable. Much better to use ID values which are a much more stable identifier.

`{{ path('record', { 'contentTypeSlug' : 'pages', 'slugOrId' : 'some-great-document' }) }}`

Can now be written as 

`{{ path('record', { 'contentTypeSlug' : 'pages', 'slugOrId' : idToSlug(6) }) }}`

### DIFF
--- a/src/Twig/ContentExtension.php
+++ b/src/Twig/ContentExtension.php
@@ -137,6 +137,7 @@ class ContentExtension extends AbstractExtension
             new TwigFilter('status_options', [$this, 'statusOptions']),
             new TwigFilter('feature', [$this, 'getSpecialFeature']),
             new TwigFilter('sanitise', [$this, 'sanitise']),
+            new TwigFilter('record', [$this, 'record']),
         ];
     }
 
@@ -788,5 +789,10 @@ class ContentExtension extends AbstractExtension
     public function sanitise(string $html)
     {
         return $this->sanitiser->clean($html);
+    }
+
+    public function record(int $id)
+    {
+        return $this->contentRepository->findOneBy(['id' => $id]);
     }
 }


I-Valchev:
Hi @benr77 , nice catch and idea!

I wonder if we can abstract this a little bit, like so:

1. Define a `|record` filter that tries to fetch a record based on the given ID
2. Use `.slug` to get the slug.

So it would work like so

{{ path('record', { 'contentTypeSlug' : 'pages', 'slugOrId' : 6|record.slug }) }}

If so, I suggest the following changes:

1. Instead of a function, create a filter named `|record` which takes the same parameters.
2. Move the filter definition from `IdToSlugExtension.php` to `ContentExtension.php` (this is where all filters related to Content are defined)
3. Instead of returning the slug, return the `$record` object itself.

This would also allow us, in future _if it is needed_, to add more parameters, e.g. fetch the record based on the slug or another field value.

benr77:
Whilst I see where you are coming from, I don't really agree with what you are saying. 

I think the syntax, which is already a bit complex for beginners, gets too cryptic with the changes you are suggesting.

This started with a conversation with @bobdenotter on Slack earlier this morning about me expecting the path() to automatically decode the ID into whatever the current slug was, as you don't want to hardcode slugs in to templates, but would hardly ever want the ID in the URL either.

Would it not be better to abstract things away in the other direction, and create Twig functions for each defined content type, to simplify the syntax - e.g.

```
{{ pagesPathById(6) }}
{{ pagesPathBySlug('some-slug') }}
{{ productsPathById(6) }}
{{ productsPathBySlug('some-slug') }}
```

It would be easy enough to generate the Twig filters "magically" by just looking up the available content types.

Any thoughts?

bobdenotter:
If we make a '|record' filter, we can simplify this: 

```
{{ path('record', { 'contentTypeSlug' : 'pages', 'slugOrId' : 6|record.slug }) }}
```

as 

``` 
{{ 6|record.link() }}
```

I-Valchev:
Ah, yes, good point @bobdenotter !

benr77:
Ok that's much better. Keeping template syntax as simple as possible is a major win.

Shall I go ahead and convert the function to a filter etc?

bobdenotter:
@benr77 Yes, sounds good! Let's convert it to `|record`, and then we can also add it to the docs as such! :-)

bobdenotter (changes requested):
Looking at the diff, it looks like you've reverted to an older version of this file before you applied your changes. :-)

benr77:
Ok all sorted now

bobdenotter (approved):
Excellent, thanks @benr77 👍

PR #2137, opened 2020-11-17
`|selected` filter orders by the value of the select

### LINKED ISSUE #2120: |selected filter ignores order of IDs
The `|selected` right now does not respect the order in which the IDs are provided.

If I specify IDs `1, 5, 2, 3`, then the selected content should also come in that order, irrespective of how it is saved in the database.

I-Valchev:
Fixes #2120

### DIFF
--- a/src/Twig/FieldExtension.php
+++ b/src/Twig/FieldExtension.php
@@ -105,7 +105,13 @@ class FieldExtension extends AbstractExtension
 
         $ids = $field->getValue();
         // Find records by their respective ids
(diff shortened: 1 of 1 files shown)


bobdenotter:
This one should also go into 4.1, if tests passes, right?

I-Valchev:
@bobdenotter yeah, I'd think so.

bobdenotter (approved):
Merging into `4.1` and `master`!

PR #2138, opened 2020-11-17
Set inside set displays properly

I-Valchev:
Don't ask me why a set inside a set works 🙈

image


bobdenotter (approved):
in**SET**ption!! 👍

PR #2140, opened 2020-11-17
Show field label inside collection

### LINKED ISSUE #2124: Set name is not visible anywhere
Right now, the set name label is not visible anywhere in the backend.

This is problematic if you have different sets that _look_ similar but have subtle differences. The Editor should be able to tell what set they added inside a colleciton, _not only from the icon_

I-Valchev:
Fixes #2124 

image

### DIFF
--- a/assets/js/app/editor/Components/Collection.vue
+++ b/assets/js/app/editor/Components/Collection.vue
@@ -19,14 +19,14 @@
             v-for="element in elements"
             :key="element.hash"
             class="collection-item"
-            v-bind:class="{ collapsed: variant !== 'expanded' }"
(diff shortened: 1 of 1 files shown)


bobdenotter:
Is that...  𝐓𝐢𝐦𝐞𝐬 𝐍𝐞𝐰 𝐑𝐨𝐦𝐚𝐧?

I-Valchev:
We never override it so.. no? It's just how the built-in font works on that size + bold, I guess.

bobdenotter (approved):
Merging in, let's see how the small font behaves on other's screens.

PR #2142, opened 2020-11-17
Fix reset password not being translatable

LordSimal:
As can be seen in #2131 I forgot to transform all english text to translatable text.

This PR fixes that.


bobdenotter (approved):
Cool thanks! 👍

PR #2144, opened 2020-11-18
Make 'create user' a bit more robust

bobdenotter:


bobdenotter:
Pinging @I-Valchev.. I'm merging this in, because i want to prep the release, but tomorrow i'd like to hear your opinion on the weirdness that's going on here.. I don't see how this could ever have worked correctly, and why it chose yesterday to start breaking on this.

tests/e2e/users.feature
```diff
@@ -151,6 +152,7 @@ Feature: Users & Permissions
     And I should see "Suggested secure password"
 
   @javascript
+      @foo
```
I-Valchev, line 155: @bobdenotter Oops... think we have to remove that! 🤪

I-Valchev:
@bobdenotter I don't know why it was breaking? Or what was breaking? 

It's strange that we need the default locale there, because the form is submitted and there the select already has a prepopulated option to begin with.

tests/e2e/users.feature
bobdenotter: Removed it! 🙈

PR #2146, opened 2020-11-18
Fix: Set a default status for the `default_status`

bobdenotter:
This is a fix for 'creating content from a 404 page', and the contenttype has no default_status set.

### DIFF
--- a/src/Entity/Content.php
+++ b/src/Entity/Content.php
@@ -186,7 +186,7 @@ class Content
         }
 
         // Set default status and default values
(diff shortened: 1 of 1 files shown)


bobdenotter:
yolo self merge

I-Valchev:
@bobdenotter I am curious why this happened, because we already set a default one, _in case it's not set_ in the parser:

https://github.com/bolt/core/blob/9448fc9be42ee4dcb62f6c8699f90a7e0904ff68/src/Configuration/Parser/ContentTypesParser.php#L109

So if it wasn't there.... is it a problem with a missing/faked config?

PR #2149, opened 2020-11-20
Add comment explaining how the orderByNumericField() internals work, …

simongroenewolt:
…so Ivo won't have to explain it in person the next time.


I-Valchev (approved):
@simongroenewolt 🙏. Thanks! Always better to comment complex things, especially those inside Storage.

PR #2172, opened 2020-11-25
Checkbox value can be unset

### LINKED ISSUE #2171: `type: checkbox` shouldn't be permanent. ;-)
If we have a contenttype like this: 

```yaml
dummies:
    name: Dummies
    singular_name: Dummy
    fields:
        title:
            type: text
        slug:
            type: slug
            uses: title
        clicky2:
            type: checkbox
    icon_many: “fa:viruses”
    icon_one: “fa:virus”
```

We can then select the checkbox.. save it.. unselect it, and save it again.. The checkbox shouldn't be checked anymore, but it is. 🤦

I-Valchev:
Fixes #2171 

It also changes some behaviour of the field value. Instead of being something _unexpected_ ("checked", "on", etc.) it will not be `true` (checked) or `false` (unchecked).

### DIFF
--- a/assets/js/app/editor/Components/Checkbox.vue
+++ b/assets/js/app/editor/Components/Checkbox.vue
@@ -1,0 +1,55 @@
+<template>
+    <div>
+        <div :class="['custom-control', 'custom-' + mode]">
+            <input
+                :id="id"
+                class="custom-control-input"
+                :name="name"
+                :checked="value"
+                type="checkbox"
+                :readonly="readonly"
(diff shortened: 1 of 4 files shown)


bobdenotter:
Cool. Let's merge this into `4.1` as well, and get out a new `4.1.x` release out this week! 🙄

bobdenotter (approved):

bobdenotter:
Merged into 4.1 and master!

PR #2173, opened 2020-11-25
Allow '/' in setcontent where query

I-Valchev:
So, for example `{% setcontent any = 'page' where { 'slug' : '/' } %}


bobdenotter (approved):
👍

bobdenotter:
Merged into `4.1` and `master`

PR #2183, opened 2020-11-26
delete controllers that only rendered the 'placeholder' template 

simongroenewolt:
These weren't used anywhere at the moment. Probably were added as 'reminders' to build something, but since they aren't even linked somewhere they can safely go.


I-Valchev (approved):
👍  Nice spot, always goot to keep core clean!

Thanks @simongroenewolt .

bobdenotter (commented):
👍

PR #2184, opened 2020-11-26
Update .travis.yml for PHP 8 compatibility

bobdenotter:
As @LordSimal said: 

> so there should be no problem to at least start a Travis CI Test started with the needed changes.

Here goes. 🎲 🎲

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
     - 7.2
     - 7.3
     - 7.4
+    - 8.0
 
 addons:
     chrome: stable
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -28,6 +28,7 @@ class Kernel extends BaseKernel
     public function registerBundles(): iterable
     {
         $contents = require $this->getProjectDir() . '/config/bundles.php';
+
         foreach ($contents as $class => $envs) {
             if ($envs[$this->environment] ?? $envs['all'] ?? false) {
                 yield new $class();


LordSimal:
I think without a PHP change travis wont start ^^

bobdenotter:
Srsly, is php going to be clever now, all of a sudden?! ;-)

LordSimal:
I wonder how the whole `psr/container` situation is being handled now.
Because as I said in my issue there is a difference between the PHP Module `php-psr` and a composer installed vendor module `psr/container`

They have both the same functionality, its just the way how it integrates with the app I think.
Because I had problems on my server where I had the `php-psr` PHP module installed without this in the composer.json

```
"replace": {
    "psr/container": "1.*"
}
```

LordSimal:
It could also be that Travis CI needs some time to add PHP 8.0 support since the final version just got released today

bobdenotter:
@LordSimal I'm guessing you're right! 

What we can do is, merge this in, and we'll find out when it does support it. 

Although, doing that might bite us in the ass at some unfortunate time. What do you think, @I-Valchev ?

LordSimal:
https://travis-ci.community/t/php-8-0-missing/10132 🧐

I-Valchev:
This PR seems to have broken Travis, because I don't see the Travis tests running?

I'm a fan of supporting PHP 8.0 as soon as we humanly can, but let's merge this in _once_ we know it actually all works.

Otherwise, if we merge it in, 1) there's no Travis at all? and 2) when Travis runs, it's quite possible tests are breaking for PHP 8.0, so we'd either have to merge PRs that fail the tests, or add PHP 8 compatibility _in a hurry_. I don't like either to be honest 😕

LordSimal:
I aggree with I-Valchev that PHP 8.0 support should be merged in 1 go and not as problems occur.
The fact that travis doesn't link this is a bit strange, but we will need to investigate more.

bobdenotter:
For future reference: Travis isn't running, because we're out of our OSS credits: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

Closing this now, moving on with #2187

LordSimal:
I think we can actually close this PR

I-Valchev:
I agree, let's close this not least because we've substituted Travis with GitHub Actions by now.

PR #2189, opened 2020-11-27
Fix for Argument 1 passed to Bolt\Repository\FieldRepository::findAllBySlug() must be of the type string, null given

### LINKED ISSUE #2188: Exception or wrong saving when changing locale
Changing locale in services.yaml causes SF exception or empty fixture (undefined name, title, etc)

Details
-------

Bolt 4.1.4 and 4.1.7 (same problem before and after udpate)
Composer install 4.1.4, / composer update 4.1.7
PHP 7.4
built-in server (bin/console server:run)

Reproduction
------------
install Bolt via composer
setup bolt
login to backend, create page, enter title, save => works fine
modify service.yaml to change main locale (tried with fr and nl, same result)
clear cache, login to backend, create page, enter title, save => SF exception
error 500
Argument 1 passed to Bolt\Repository\FieldRepository::findAllBySlug() must be of the type string, null given, called in D:\_PROJETS\Regards des lieux\www\vendor\bolt\core\src\Event\Listener\ContentFillListener.php on line 95

same scenario, but with a db reset after modifying the locale in service.yaml causes no SF exception but the fixture is saved "empty" with undefined name, undefined slug, and the url of the backend adds ?edit_locale=fr to the url.
Second attempt reach the exception

I-Valchev:
Fixes #2188

### DIFF
--- a/src/DataFixtures/ContentFixtures.php
+++ b/src/DataFixtures/ContentFixtures.php
@@ -185,6 +185,9 @@ class ContentFixtures extends BaseFixture implements DependentFixtureInterface, 
 
         $field = FieldRepository::factory($fieldType, $name);
 
+        // Make sure to create the fixture in the correct default locale.
+        $field->setLocale($this->defaultLocale);
+
         if (isset($preset[$name])) {
             $field->setValue($preset[$name]);
         } else {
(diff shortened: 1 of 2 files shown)


bobdenotter (approved):
:+1:

PR #2191, opened 2020-11-29
Allow bolt/common to be 3.0 for PHP 8.0

LordSimal:
Now that `bolt/common` allows PHP 8.0 `bolt/core` should at least allow the newer version to be installed.


composer.json
```diff
@@ -11,7 +11,7 @@
         "api-platform/core": "^2.5",
         "babdev/pagerfanta-bundle": "^2.5",
         "beberlei/doctrineextensions": "^1.2",
-        "bolt/common": "^2.1.10",
+        "bolt/common": "^2.1.10 || ^3.0",
```
bobdenotter, line 14: Let's make it 'just' 3.0, though. :-)

```suggestion
        "bolt/common": "^3.0",
```

bobdenotter (changes requested):

composer.json
LordSimal: well, already adjusted it now directly in code 😄

bobdenotter (approved):

PR #2198, opened 2020-11-30
Allow `id` in sanitiser (purifier). Don't sanitise fields with `sanitise: false`

I-Valchev:


bobdenotter (approved):
Let's roll with this, until we decide to rip the entire thing out!

PR #2200, opened 2020-11-30
Feat: remove DoctrineMigrationBundle that does not seem to be in use

toofff:


bobdenotter (approved):
Seem like this isn't used indeed! Thanks, @toofff 👍

toofff:
@bobdenotter on the other hand I do not see why the CI is in failure knowing that it works perfectly with PHP 7.2.

By the way, what do you use the symfony.lock file for in the "Bolt/Core" repository?
I would tend to delete it completely in this repository.

I-Valchev:
@toofff this is a quick with behat, most of the time it works but sometimes it fails randomly...

After restarting it, though, they pass! 👍 

Let's merge it in, thanks!

toofff:
@I-Valchev & @bobdenotter 

For me it is a very good thing to use it, if you are OK I will revert my commit and I will do 2 doctrine migrations for you.

A migration for version 4.0 and a migration for 4.1.

I also plan to remove all the commands to directly update the DB in the tests by the command to start the doctrine migrations. Like that the doctrine migrations will be tested in passing. And change the command to launch the fixtures or the documentation so that we can go through the doctrine migrations for the creation of the database.

Then documentation to update the upgrade files of the different versions of Bolt 4 with an explanation for not having to launch some migration otherwise it could break everything in production.

And surely a pull request to update the `Bolt/Project` repository.

Would that be okay with you?

bobdenotter:
> For me it is a very good thing to use it, if you are OK I will revert my commit and I will do 2 doctrine migrations for you.

Well, if you can make it so it works on MySQL, _and_ SQLite, _and_ Postgres, that'd be wonderful. I haven't figured out how we can do a migration that's not tied to a specific database platform. :-)

toofff:
I'll look in that direction in this case.

But if we have to do migrations according to the DBMS used, then we will have to do 3 migrations for each feature in my opinion.

Afterwards, if we use the same DBMS functionalities, there shouldn't be too many problems.

I'll leave this Pull Request active until then and open a new Pull Request for whatever I plan to do. If I can't, maybe it would be better to delete the Bundle.

If we want to correctly test the migrations, we will have to run the tests that use it on the 3 desired DBMS.

toofff:
I think I have a solution to the problem of multiple DBMSs but for it to be clean, easy to use and without configuration problem, I will pass my pull request on adding the Boltbundle before if you don't mind?  At the moment we can remove the dependency as it will be reinstalled with the latest version, if you agree?

bobdenotter:
> I will pass my pull request on adding the Boltbundle before if you don't mind? 

Yes, certainly. Let's tackle one thing at a time! :-)

> At the moment we can remove the dependency as it will be reinstalled with the latest version, if you agree?

Yup, sounds good. Let's roll with this!

PR #2203, opened 2020-12-01
Don't break with `TaxonomyRepository::factory() must be of the type string, int given`

### LINKED ISSUE #2201: Taxonomy with Number and + converted to int and cannot be added to a content type
I have a taxonmy option called '66+', when adding it at the Backend to the Content-Type, i get an Error:

> Argument 2 passed to Bolt\Repository\TaxonomyRepository::factory() must be of the type string, int given, called in /var/www/bolt/vendor/bolt/core/src/Controller/Backend/ContentEditController.php on line 482

When debugging in Frontend, i recognized that the value is already set as int by vue.

> value='["50-65",66]'

Details
-------

| Relevant Bolt Version                                 | 4.1.8
| <abbr title="Backwards Compatibility">BC</abbr> Break | no
| Web server                                            |  Nginx
| For UX/UI issues                                      | Firefox Developer Edition 84.0b4

Reproduction
------------
Add Taxonmy like 

```
offerAges:
  name: Age-Groups
  singular_name: Age-Group
  slug: ages
  singular_slug: age
  behaves_like: categories
  multiple: true
  options: [ "8-10", "10-14", "14-16", "16-18", "18-20", "20-29", "30-49", "50-65", "66+" ]
  required: true
```

### Steps to reproduce

Add Taxonmy to Content-Type

### Expected result

Should save  '66+', not 66.

### Actual result

Error, because converted to integer

I-Valchev:
Fixes #2201 

PHP Arrays are **_THE WORST_**

`A key may be either an integer or a string. If a key is the standard representation of an integer, it will be interpreted as such (i.e. "8" will be interpreted as 8, while "08" will be interpreted as "08").` 🤦‍♂️

### DIFF
--- a/src/Controller/Backend/ContentEditController.php
+++ b/src/Controller/Backend/ContentEditController.php
@@ -485,7 +485,7 @@ class ContentEditController extends TwigAwareController implements BackendZoneIn
             ]);
 
             if ($taxonomy === null) {
-                $taxonomy = $this->taxonomyRepository->factory($key, $slug);
+                $taxonomy = $this->taxonomyRepository->factory($key, (string) $slug);
             }
 
             $content->addTaxonomy($taxonomy);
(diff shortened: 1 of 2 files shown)


bobdenotter:
Haha, Arrays _are_ terrible indeed! :-D

bobdenotter:

PR #2206, opened 2020-12-02
[WIP] Re-enable multiple db table prefixes

simongroenewolt:
Force initialization of ObjectManager for correct spl_object_hash() behavior. 
Needed to get the correct spl_object_hash in case it is a proxy

### DIFF
--- a/src/Doctrine/TablePrefixTrait.php
+++ b/src/Doctrine/TablePrefixTrait.php
@@ -23,6 +23,10 @@ trait TablePrefixTrait
 
     protected function setTablePrefix(ObjectManager $manager, string $prefix)
     {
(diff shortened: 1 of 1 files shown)


simongroenewolt:
@bobdenotter I just realized that without talking about it this PR will just be forgotten :-)

Are you interested in getting this merged? I've still marked it WIP as the change itself doesn't _do_ anything, it merely makes it possible to revert your old workaround that made it impossible to have multiple table prefixes. I'm a bit wary to include a revert of those changes in this PR, but maybe that would be the correct thing to do?

bobdenotter:
> Are you interested in getting this merged?

Yes, certainly. What I've done there was undeniably a hack. If we can revert it, I'm all for it. :-)

PR #2207, opened 2020-12-02
Fix homepage link generation, generate fallbacks as relative unless canonical

SahAssar:
This fixes an issue I had where the current locale was not picked up by homepage link generation.

Along the way I noticed that the fallback link generation ("Just use the current URL") always generated canonical links which sometimes led to links like "https://domain.tld/https://domain.tld" because it is used as a argument to setPath in getLink.


I-Valchev (approved):
Looks good! Thanks @SahAssar  👍

I-Valchev:
And merged into 4.1 too.

PR #2211, opened 2020-12-04
Ensure `selected` filter returns record if set to `multiple: false`

I-Valchev:


bobdenotter:

bobdenotter (approved):

PR #2214, opened 2020-12-04
Update bolt/common version

milosa:
This is blocking the use of PHP8.


bobdenotter:
This is already done in `master`: https://github.com/bolt/core/blob/master/composer.json#L14

milosa:
Hmm, I missed that because the default branch is 4.1, not master.

I'll close this.

Issue #2215, opened 2020-12-05
Searchable select fieldtype


denis-gorin:
Hi, please tell, are there any plans to add searchable option to select fieldtype?

bobdenotter:
Plans yes, but i'm afraid no timeframe.

denis-gorin:
> Plans yes, but i'm afraid no timeframe.

Ok, I’ll try ta handle with vue-multiselect

PR #2217, opened 2020-12-05
Issue 2215 searchable select

denis-gorin:
Implements #2215

Just added forwarding for searchable prop in twig template

### DIFF
--- a/templates/_partials/fields/select.html.twig
+++ b/templates/_partials/fields/select.html.twig
@@ -15,6 +15,10 @@
     {% set multiple = field.definition.get('multiple')|default ? 'true' : 'false' %}
 {% endif %}
 
+{% if searchable is not defined %}
+    {% set searchable = field.definition.get('searchable')|default ? 'true' : 'false' %}
+{% endif %}
+
 {% block field %}
     <editor-select
         :classname="{{ "wide-options"|json_encode }}"
@@ -24,6 +28,7 @@
         :options="{{ options|json_encode }}"
         :form='{{ form|json_encode }}'
         :multiple="{{ multiple }}"
+        :searchable="{{ searchable }}"
         :readonly="{{ readonly|json_encode }}"
         :errormessage='{{ errormessage|json_encode }}'
         :allowempty="{{ required ? 'false' : 'true' }}"


denis-gorin:
oops, something went wrong... where I've made a mistake in branches...

bobdenotter:
I've changed the base of the PR to `master`. It looks like it's better now. 

I'm not certain your fix is totally correct though. I recall we did something similat for "Relations", but we couldn't do it the same for "select". Maybe @I-Valchev can chime in?

denis-gorin:
Please tell, what is confusing you and I'll try to fix or test.
The most of case have been already done in `assets\js\app\editor\Components\Select.vue`  I mean `searchable` in props and `multiselect` component, so, only that I had to do was to forward `searchable` option from `contenttype.yaml` to `templates\_partials\fields\select.html.twig` and do testing that all is working good

I guess I need to test it with Relations to not to break.

I-Valchev:
Hi @denis-gorin , I remember an issue when using https://github.com/shentao/vue-multiselect with `taggable` and `searchable` options at the same time. Here's the details:

Taggable is used in Bolt when `multiple: true` is set for the select. It allows editors to re-arrange the order of the selected values. However, when `searchable: true` is also set, the two did not seem to work together.

This is the reason that, when @nestordedios and I worked on the relations, we did not extend the searchable to `select` fields. It could be that this is no longer an issue, are you able to verify this @denis-gorin ?

I-Valchev:
As a side note, I'm really confused why the Github Acitons checks are not running on this PR?

denis-gorin:
> This is the reason that, when @nestordedios and I worked on the relations, we did not extend the searchable to `select` fields. It could be that this is no longer an issue, are you able to verify this @denis-gorin ?

You are right, I've got an issue with ordering. Everything was ok till I try to re-arrange elements... I will investigate this.


image

image

denis-gorin:
Hi, I have not good news for now. 
I've found that when multiselect is searchable and we can not drag tags to re-arrange, the root is deep in `vue-multiselect` component. 
Look at this [fiddle](https://jsfiddle.net/mkhz84Ld/). Try emulate `drag` - just mark text as selected in selected option (I mean in the tag) both in not searchable case and in searchable case.  So, In searchable case you couldn't. It looks like vue-multiselect re-rendered every time on focus.

I-Valchev:
@denis-gorin thanks, even though we can't implement this _just yet_, it is good to confirm where the issue is. I wonder if it makes sense to raise an issue with the `vue-miltiselect` developers directly about it? I am pretty sure we can easily incorporate your PR if the problem is fixed upstream.

denis-gorin:
Yes, on the one hand we have 'upstream approach', on the other - local workaround - a little bit ugly, but working - adding mouseenter (mouseleave) events on tags to toggle searchable option on the fly, so if I want to re-arrange - I can, just starting drag, else - use search

I-Valchev:
Hi @denis-gorin , if there's an option that keeps both functionalities working, and a clean-ish code, I'm all for bringing it in. 👍

denis-gorin:
We need to meditate on this... May be some icons to switch options to swich searchable and draggable modes in UI

Now it looks a little [buggy (video)](https://drive.google.com/file/d/1CezZPTRYw7Cw0RwkfrsbrSMHAzuAqnCt/view)

I-Valchev:
It's good progress! But I agree UI-wise we can try to improve it.

bobdenotter:
What is the status of this PR? Is this "good enough" as in "Not perfect, but better than it was before"? If so, let's merge and worry about further improvements later! :-)

bobdenotter (approved):
Let's get the ball rolling on this one.. Additional tweaks can be done in follow-up PR's. 👍

PR #2223, opened 2020-12-08
Don't break if a set's definition no longer contains a field

I-Valchev:


bobdenotter:
Closed in favor of #2240

PR #2227, opened 2020-12-09
Run E2E tests on lowest supported PHP version only [new features]

I-Valchev:
Same as #2228 but for the features (next minor version) branch


bobdenotter (approved):
Cool, let's roll with this! :-)

PR #2228, opened 2020-12-09
Run E2E tests on lowest supported PHP version only [bugfixes]

I-Valchev:
Same as #2227 but for the bugfix branch


bobdenotter (approved):
Likewise!

PR #2229, opened 2020-12-09
Feat: remove SymfonyMakerBundle that does not seem to be in use

toofff:


I-Valchev:
hi @toofff , you are right I don't think we use this in `bolt/core` directly. However, I have seen colleagues use things like `bin/console make:entity` when they are creating custom entities within a Bolt project.

I'd prefer to keep this functionality. Would you agree @toofff ?

toofff:
@I-Valchev 

For me it would be no but I will explain my choice.

It is important to distinguish between developing a Symfony component for Bolt and using the Bolt component in a Symfony project.

**For the case of the development of a Symfony component**
If you add it as a `require` dependency in the component it will be downloaded in vendors in all its situations:
- component installation normally
- installation of the component with the development packages

If you add it as a `require-dev` dependency in the component it will be downloaded in vendors in all its situations:
- installation of the component with the development packages

So this package will be downloaded and not used in production and in the CI, which for me is a very bad practice.

**For the development of a Symfony project**
We will add this dependency in `require-dev` for the symfony project and the Bolt component will be `require` dependency.

So when we put the project into production, we will install only the `require` dependencies with composer and we will not have the MakerBundle in production.

**This dependence is only useful for development.** So on the developer's development machine

If I'm doing this Pull Request it's to lighten the vendors which can be quite heavy for a docker image or for normal server disk space. And everything that is used in development of a symfony project should absolutely not end up in production.

I-Valchev:
Hi @toofff , thanks for the detailed answer and explanations. Appreciate it!

I agree with you that we should keep `bolt/core` as clean as practical. Similarly, `bolt/project` should also be clean, even cleaner! That's one of the nice things right now for the `composer.json` that is in `bolt/project`.

My understanding is that if the dependency is moved in `require-dev` here, then it will not be installed when people build a bolt project with `composer create-project bolt/project [projectname]`. Which, in turn, means that all the `bin/console make:` commands will not work. Is that right? @toofff 

If so, my 2 cents will go to keeping the bundle as-is, for the sake of development experience. I've not yet personally faced a situation that there's not enough space on the server to install dependencies. Maybe others have, though.

toofff:
@I-Valchev 

If you place the dependency in `require-dev`, it will not be in production (because in production we only install the dependencies of the` require` part), on the other hand it will be installed but not used in the CI (and a CI is very energy and time consuming, so the less there is in it the better for the planet).

So we only need this Bundle for development, so in a Symfony project.
The Symfony community is familiar with this tool and if they need it and use it, they install it in their Symfony project.

So for our specific case, I recommend removing it from the Bolt / Core repository and if you want to keep it to highlight it at this time you have to add it in the Bolt / Project repository in `require-dev`.

Personally I find this to be a bad idea, I will just talk about it in the Bolt documentation in a development section by linking to the MakerBundle documentation and I won't add it in any repository as the developers are not likely to use it.

For me MakerBundle is a tool like any other, if you need it you install it, if it is not vital to the project it has nothing to do in it.

bobdenotter (approved):
Thanks, @toofff 👍

PR #2231, opened 2020-12-10
Show correct timezone for Content dates (createdAt, modifiedAt, publishedAt, depublishedat)

I-Valchev:
With credits to @Dsduit for pointing this one out, so that _finally_ I could reproduce it reliably.


bobdenotter (approved):
Huzzah! 👍

PR #2233, opened 2020-12-10
Warn of outstanding migrations

I-Valchev:
This is for `bolt/core`, but we can/should also do it in project. To do that, though, we'd need to revise `bolt/project` and move the post-update things in here, so we can version them.

image

### DIFF
--- a/bin/composer-script/post-update-cmd.php
+++ b/bin/composer-script/post-update-cmd.php
@@ -26,4 +26,8 @@ run('php bin/console extensions:configure --with-config --ansi', $symfonyStyle);
 run('php bin/console cache:clear --no-warmup', $symfonyStyle);
 run('php bin/console assets:install --symlink --relative public', $symfonyStyle);
 
+$migrate = "Database is out-of-date. To update the database, run `php bin/console doctrine:migrations:migrate`.";
+$migrate .= " You are strongly advised to backup your database before migrating.";
+run('php bin/console doctrine:migrations:up-to-date', $symfonyStyle, false, $migrate);
+
 run('php bin/console bolt:info --ansi', $symfonyStyle, true);
--- a/bin/composer-script/run.php
+++ b/bin/composer-script/run.php
@@ -2,12 +2,16 @@
 
 use Symfony\Component\Console\Style\SymfonyStyle;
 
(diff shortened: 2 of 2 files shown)


bin/composer-script/post-update-cmd.php
```diff
@@ -26,4 +26,8 @@
 run('php bin/console cache:clear --no-warmup', $symfonyStyle);
 run('php bin/console assets:install --symlink --relative public', $symfonyStyle);
 
+$migrate = "Database is out-of-date. To update the database, run `php bin/console doctrine:migrations:migrate`.";
+$migrate .= " You are strongly advised to backup your database before migrating.";
+run('php bin/console doctrine:migrations:up-to-date', $symfonyStyle, false, $migrate);
```
bobdenotter, line 31: This doesn't look like it exists? Or is that part of the trick to make it work? 

```
php bin/console doctrine:migrations:up-to-date

  There are no commands defined in the "doctrine:migrations" namespace.

  Did you mean one of these?
      doctrine
      doctrine:cache
      …
```

bobdenotter: Never mind! I am an idiot! #2232

bobdenotter (approved):
Whoop whoop!

toofff:
@bobdenotter & @I-Valchev 

I find it a shame to have the launch of the doctrine migrations directly in a script that is launched during a composer update or install. Because the installation is not as flexible as it used to be.

With Docker and Kubernetes (and surely other technology for deployment) it may be a problem to no longer have this flexibility or otherwise we will have to make our own scripts and not use those from the core repository.

I-Valchev:
Hi @toofff, just to be on the same page hear: those scripts only alert users that there are migrations that have not been applied yet. No migrations are actually run during update or install.

Is that an issue in your opinion?

toofff:
@I-Valchev For me it's a problem to have this kind of script during a composer install or update, because it doesn't fit here.

But I'm explaining all this to you this afternoon in the Docker channel on slack. Can I put a summary of the conversation here afterwards if you wish?

I-Valchev:
@toofff Sure 👍

PR #2234, opened 2020-12-10
Canonical improvements

doup:
This PR fixes the following issues:

## Fix listing page Canonical

> Listing pages have different canonical when using singular or plural form: https://github.com/bolt/core/issues/2218

Example:

- Using `localization.remove_default_locale_on_canonical = true`
- Note how the URL contains the singular form but **the canonical is normalized to the plural form**
- Also the default locale (`en`) is removed (old behaviour)

image

## Force default locale on canonical

> Add a new config option to tell Bolt if the canonical URL should force-prefix the default locale or not: https://github.com/bolt/core/issues/2219

Example:

- Using `localization.remove_default_locale_on_canonical = false`
- Note how the **default locale (`en`) is maintained on the canonical**

image

### DIFF
--- a/src/Controller/Frontend/ListingController.php
+++ b/src/Controller/Frontend/ListingController.php
@@ -80,6 +80,16 @@ class ListingController extends TwigAwareController implements FrontendZoneInter
 
         $records = $this->setRecords($content, $amountPerPage, $page);
 
+        // Set canonical URL
+        $this->canonical->setPath(
+            'listing_locale',
+            array_merge([
+                'contentTypeSlug' => $contentType->get('slug'),
+                '_locale' => $this->request->getLocale(),
+            ], $queryParams),
+        );
+
+        // Render
         $templates = $this->templateChooser->forListing($contentType);
         $this->twig->addGlobal('records', $records);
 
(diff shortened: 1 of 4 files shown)


I-Valchev:
Hi @doup , thanks for this PR.

Looks like there are one or two syntax issues with the code, can you fix them? Please also run `make csfix` so that ecs and phpstan don't complain about the code styles :-)

Ready to merge this in once tests pass! 👍

doup:
Yeah… I though I ran the style checks on my machine but apparently I didn't. 😅 
I'll check it later, if I can't run the tests locally I'll ask in Slack for help. 👍

src/Controller/Frontend/ListingController.php
```diff
@@ -80,6 +80,16 @@ public function listing(ContentRepository $contentRepository, string $contentTyp
 
         $records = $this->setRecords($content, $amountPerPage, $page);
 
+        // Set canonical URL
+        $this->canonical->setPath(
+            'listing_locale',
+            array_merge([
+                'contentTypeSlug' => $contentType->get('slug'),
+                '_locale' => $this->request->getLocale(),
+            ], $queryParams),
```
bobdenotter, line 89: Only in PHP8 you are allowed to have trailing comma's is parameter lists. :-)

```suggestion
            ], $queryParams)
```

bobdenotter (approved):
Tests are passing, now that that comma has been removed. 

Thanks, @doup 👍

src/Controller/Frontend/ListingController.php
doup: Thx! It's been a couple of years I don't code in PHP… that's I suppose is TS habit. ^_^U

PR #2235, opened 2020-12-11
`|svg` filter uses correct path inside collection/set

### LINKED ISSUE #1892: |svg filter should use path not filename
The filter does not always work because it uses the filename, not the path.

I-Valchev:
Fixes #1892 

Thanks @eduardomart and @nestordedios for finding and confirming the issue.


bobdenotter (approved):
👍

PR #2236, opened 2020-12-11
Show the user avatar in the toolbar, if it exists

I-Valchev:
Little improvement, but looks nice :-)

image


bobdenotter (approved):
That's neat! I like it! 👍

PR #2247, opened 2020-12-14
Localize menu with `{locale}: {label}` configuration

### LINKED ISSUE #2222: Localize menu
We could localize the menu like so:

```
main:
  - en_label: Home
  - es_label: Incio
    link: homepage
    class: homepage
```

Then make sure that the default twig template that renders it uses something like this:

```
{{ item[app.request.locale ~ '_label'] }}
```

With proper fallbacks if it is missing.

I-Valchev:
Fixes #2222 

For example:

```yaml
main:
  - label: Home
    title: This is the <b>first<b> menu item.
    link: homepage
    class: homepage
  - label: 
        en: About
        nl: Over
    link: page/2
```

### DIFF
--- a/config/bolt/menu.yaml
+++ b/config/bolt/menu.yaml
@@ -6,7 +6,9 @@ main:
     title: This is the <b>first<b> menu item.
     link: homepage
     class: homepage
-  - label: About
+  - label: 
+        en: About
+        nl: Over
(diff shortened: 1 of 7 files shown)


bobdenotter (approved):
Neat! 👍

PR #2249, opened 2020-12-15
[WIP] Always return field objects (not Markup)

I-Valchev:
Prior to this, some fields were passed to twig as object, some as Markup.

Apart from the ugly inconsistency, the issue is that the Markup ones don't then work with twig functions/filters that expect fields. This is specifically an issue with set, where `set.field` is different than `set.value.field`.

This PR fixes it, however it introduces a breaking change by deprecating the `ScalarCastable` interface in favour of `HtmlRenderable`.

Things that need to be updated:
https://github.com/bolt/article/blob/master/src/Entity/ArticleField.php
https://github.com/bolt/redactor/blob/master/src/Entity/RedactorField.php
https://github.com/LordSimal/bolt-geolocation-field/blob/master/src/Entity/GeolocationField.php

### DIFF
--- a/src/Controller/TwigAwareController.php
+++ b/src/Controller/TwigAwareController.php
@@ -8,6 +8,7 @@ use Bolt\Canonical;
 use Bolt\Configuration\Config;
 use Bolt\Configuration\Content\ContentType;
 use Bolt\Entity\Content;
+use Bolt\Entity\Field\HtmlRenderable;
 use Bolt\Entity\Field\TemplateselectField;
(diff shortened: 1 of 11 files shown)


I-Valchev:
@bobdenotter as discussed, we can't do this yet because of the internal implementation of symfony/twig safe functionality.

One day, when it is possible to mark an object safe (as opposed to class/interface right now), we'll be able to fix this. In the meantime, let's close the PR.

PR #2256, opened 2020-12-19
Make the migrations conditional

bobdenotter:
On an installation where I already ran `bin/console doctrine:schema:update` a while ago, the migration broke and was stuck from then on. 

by checking if the table and column don't exist yet, it'll continue and prevent breakage.


I-Valchev (approved):
👍  Cool!

Issue #2087, opened 2020-11-01
PHP 8.0 compatibility 


LordSimal:
Details
-------

| Question                      | Answer
|------------------------- | ------
| Relevant Bolt Version   | 4.1.4
| Install type                   | Composer install
| BC Break                      | no
| PHP version                 | 8.0 RC3
| Web server                  | Nginx
| For UX/UI issues          | Latest Chrome

Reproduction
------------

### Bug summary

BoltCMS currently breaks when trying to run it on PHP 8.0 RC3

### Specifics

After installing the current PHP 8.0 RC3 release I got the following error when trying to access the frontpage

![image](https://user-images.githubusercontent.com/9105243/97802230-3fe00f80-1c42-11eb-8b3c-46a89d1d3143.png)

```
Compile Error: Cannot declare interface Psr\Container\ContainerInterface, because the name is already in use
```

This is related to https://github.com/symfony/symfony/issues/36676

Basically you can either have the psr/container interface installed via composer or as a PHP module.

To circumvent that we need to add the following to the composer.json

```
{
    "replace": {
        "psr/container": "1.*"
    }
}
```

After a new `composer install` the following error occurs

![image](https://user-images.githubusercontent.com/9105243/97802278-92b9c700-1c42-11eb-917e-cd18118e4c8d.png)

```
syntax error, unexpected token "match"
```

This is realted to https://github.com/willdurand/Negotiation/pull/106

The module https://github.com/willdurand/Negotiation is being required by `api-platform/core` which is present in our composer.json

But as can be seen in that Pull Request there is no stable release for that fix at the moment.

I tried that fix on my instance and BoltCMS works after these changes.

![image](https://user-images.githubusercontent.com/9105243/97802309-d01e5480-1c42-11eb-96a1-43e8a8e90605.png)

I would say we'll have to wait for a stable release and then adjust our TravisCI tests to check if other things break as well under 8.0

LordSimal:
PHP 8.0 has just been released
https://www.php.net/archive/2020.php#2020-11-26-3

LordSimal:
https://github.com/willdurand/Negotiation/releases/tag/3.0.0 has been released, so there should be no problem to at least start a Travis CI Test started with the needed changes.

bobdenotter:
@LordSimal Sure! Let's see how it goes! #2184

LordSimal:
I just retried this whole 8.0 problem again.
[...]

PR #2257, opened 2020-12-20
allow tightenco/collect 8.0 to be installed

LordSimal:
Related to #2087 

I have no idea if 8.0 is downwards compatible with 7.0 nor do I have an idea how and where collections are used here.
It is just one step closer to allowing bolt to be installed with a PHP 8.0 CLI

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
         "symfony/webpack-encore-bundle": "^1.7",
         "symfony/yaml": "^5.1",
         "symfonycasts/reset-password-bundle": "^1.1",
-        "tightenco/collect": "^7.25",
+        "tightenco/collect": "^7.25 || ^v8.0.0",
         "twig/twig": "^3.0",
         "ua-parser/uap-php": "^3.9",
         "webimpress/safe-writer": "^2.1",
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
         "symfony/webpack-encore-bundle": "^1.7",
         "symfony/yaml": "^5.1",
         "symfonycasts/reset-password-bundle": "^1.1",
-        "tightenco/collect": "^7.25 || ^v8.0.0",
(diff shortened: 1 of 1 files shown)


I-Valchev (commented):
Nice PR, thanks @LordSimal. I've asked about it here: https://github.com/tighten/collect/issues/259

Let's see what we have to change. Additionally, I'll create an issue for deprecating `tightenco/collect` in favour of `illuminate/collections` once we're ready to drop support for PHP 7.2.

bobdenotter:
I'm not a fan of dependencies with two major versions. In `bolt/project` we have to include the composer.lock, so it will be locked to _something_. If only one works for different setups, the `bolt/project` installation will definitely break for people.

I-Valchev:
Oh that's a good point!

LordSimal:
Then I would guess we leave it at the 7.0 till PHP 7.2 support is dropped.
Then we can go further with the PHP 8.0 compatibility.

bobdenotter:
Are they really mutually exclusive? No `tightenco/collect` version that works on Both PHP 7 and 8? 

Since we're sticking with the Symfony requirements, we will drop PHP 7.2 support in the fall of 2022 at the earliest. (Since that's when SF 6 will be released, and they will not raise the minimum PHP requirement in a minor release, which I'm very thankful for)

LordSimal:
This commit added the 8.0 "support" to the composer.json and therefore allows composer with PHP 8.0 CLI to not break.
https://github.com/tighten/collect/commit/995114659a88f8546b1ad06d895295d23943e63c

Since no 7.0 release has been made since that commit I doubt they will port it.
So let me try to just change it to ^8.0

LordSimal:
seems like its working 🤷‍♂️

bobdenotter:
Whoop! Thanks, @LordSimal 👍

bobdenotter (approved):

PR #2268, opened 2020-12-28
 Put "Reset Password" functionality on Backend controller collection, so it doesn't need changes in routing

bobdenotter:


bobdenotter:
Does this change seem OK to you, @LordSimal ?

LordSimal:
I dont even know what the `reset-password` route was so I have no problem removing that form the routes.
The only thing I can find is the logic where you can reset the password via the CLI console command.

Other then that I can't see any problems with that change.

LordSimal:
I just first tought that all the Controllers in the `Backend` folder are only accessible via the `/bolt` backend URL and therefore need an account to be accessed (which is not really good for a ResetPasswordController 😛 )

bobdenotter:
> I dont even know what the reset-password route was so I have no problem removing that form the routes

It was from the old boilerplate i used for auth, and was never really used. We can safely remove it. :-)

> I just first tought that all the Controllers in the Backend folder are only accessible via the /bolt backend URL

Not necessarily. By mounting them on `%backend_url%` we not only keep it out of the 'main' routing, but we'll also allow developers to configure it to be something else than `/bolt` entirely. :-)

PR #2281, opened 2021-01-05
Show number of records on listing

I-Valchev:
@bobdenotter I noticed we sometimes calculate the number of records by multiplying the page * record listing thing 

image


bobdenotter (approved):
Cool!

Issue #2284, opened 2021-01-05
[RFC] Textarea attributes (add maxlength, remove pattern)


napley:
The pattern attribute is not working for textarea input : https://developer.mozilla.org/fr/docs/Web/HTML/Element/Textarea

- So I propose to remove the pattern option for textarea fieldtype.
- Moreover I think useful to add the maxlength attribute support for the textarea fieldtype to compensate for the lack of pattern support.

I already made a branch in my repo, I can create a PR

I-Valchev:
Nice! I'm going to assign this issue to you then, thanks @napley !

PR #2286, opened 2021-01-05
Feature/maxlength textarea #2284

napley:
- remove the pattern option for textarea fieldtype
- add the maxlength attribute support for the textarea fieldtype


napley:
I made a mistake when I create the PR

PR #2289, opened 2021-01-06
Fix allow twig and deprecate `record|allow_twig` requirement

I-Valchev:
One less `Request::createFromGlobals()` :innocent:


bobdenotter (approved):
Oooh, so much better!! 👍 

tenor

PR #2296, opened 2021-01-07
Fix `default` option for select field

### LINKED ISSUE #2152: The default select field doesn’t work properly
This is for bolt 4.1.
The default select field didn't apply when you create a new item.

Code in contenttype file: 

my_field:
            label: "Name"
            type: select
            values: [ Value1, Value2, Value3 ]
            default: [ Value1 ]

I-Valchev:
Fixes #2152


bobdenotter (approved):
👍

PR #2299, opened 2021-01-07
Fix finder quickselect dropdown descriptions

### LINKED ISSUE #1140: Config dropdown is not _very_ readable
Screenshot 2020-02-26 at 15 02 21

The option's descriptions are too long and get cropped ugly.

I-Valchev:
Fixes #1140

image

### DIFF
--- a/templates/finder/_quickselect.html.twig
+++ b/templates/finder/_quickselect.html.twig
@@ -27,7 +27,7 @@
             },
             options: [
                 {% for file in allfiles %}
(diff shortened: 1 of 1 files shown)


bobdenotter:
The reason why I initially gave it long descriptions, is because it then picks up on that in the 'typeahead'. making it easier to find something. 

This fix will reduce that, and it'll still be messy on mobile. I'd prefer if we could keep the descriptions intact, and instead fix the CSS that corrupts the display.

I-Valchev:
Ah, I see! Okay let's close this and find a better (CSS) solution :+1:

PR #2300, opened 2021-01-08
[WIP] Accessibility (a11y) automated tests

### LINKED ISSUE #2084: Integrate accessibility checks as part of CI
The goal is part of the Editor Accessibility milestone.

The idea behind this is to see if we can integrate a tool that checks accessibility issue as part of or in addition to the E2E tests during CI. If we do this and clean up all existing issues, by including it in CI, we'll make sure to not introduce new issues with future PRs. Aiming for accessibility by design here, rather than "fixing" and "patching" things every now and then.

If anyone has experience or thoughts about it, they're much appreciated!

I-Valchev:
Fixes #2084

A couple of notes about the current configuration:

- `threshold` in CI is set to 82, meaning tests will only fail the build if more than 82 errors are on a page. Right now, the most problematic page has 82 errors, hence if we start introducing new issues, tests will fail, but they won't fail for existing issues. That gives us time to fix the outstanding problems.
- `concurrency` is set to 1, meaning only one page at a time. This is so the first (authentication) test can be ran first, to the backend is accessible. Ideally, we'd be able to run the checks concurrently one day.

### DIFF
--- a/.github/workflows/a11y_tests.yaml
+++ b/.github/workflows/a11y_tests.yaml
@@ -1,0 +1,32 @@
+name: Accessibility (a11y) Tests
+
+on:
+    pull_request:
+
+jobs:
+    a11y_tests:
+        strategy:
+            fail-fast: false
+            matrix:
+                actions:
+                    -
+                        name: Run pa11yci
+                        run: npm run a11y:pa11y-ci
+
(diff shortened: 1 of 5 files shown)


bobdenotter (approved):
Awesome! 🏅

PR #2301, opened 2021-01-08
Run bolt in a11y tests

### LINKED ISSUE #2084: Integrate accessibility checks as part of CI

I-Valchev:
Fixes #2084 

Roughly speaking, this will make sure that the tests aren't failing for outstanding issues. But if we introduce new problems, they will most likely be flagged by pa11y and tests will fail. Seems like a good compromise as we undertake the big work of making the whole of Bolt accessible.

### DIFF
--- a/.github/workflows/a11y_tests.yaml
+++ b/.github/workflows/a11y_tests.yaml
@@ -27,6 +27,7 @@ jobs:
                 uses: actions/setup-node@v1
                 with:
                     node-version: ${{ matrix.node-version }}
+                    coverage: none
             - uses: shivammathur/setup-php@v1
               with:
                   # test the lowest version, to make sure checks pass on it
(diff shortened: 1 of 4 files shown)


I-Valchev:
This seems to depend a lot on https://github.com/pa11y/pa11y-ci/issues/70

bobdenotter (commented):
Whoop whoop!

bobdenotter (approved):

PR #2308, opened 2021-01-12
Open secure link new tab

I-Valchev:
Make sure you don't close the tab, losing all outstanding changes


bobdenotter (approved):
Ah, nice catch!

PR #2313, opened 2021-01-14
Use default locale when not explicitly stated

I-Valchev:
I'm not even sure why this is broken in the first place 🤔


bobdenotter (approved):
Whoop!

Issue #2241, opened 2020-12-13
Issue with 'orderby' in query


Sandrine-D:
Hello,
I don't really know if it's a bug or a problem on my side but I'm facing an exception in queries where the "orderby" clause contains fields other than the basic fields (datepublish...)

Details
-------
| Relevant Bolt Version | 4.1.5
| Install type                   | Composer install
| PHP version                 | 7.3.23 |
| Web server                  | Apache

Reproduction
------------

**Query OK:** 
`{% setcontent records = 'articles' where { region:region.id } orderby '-publishedAt' %} `
          
**Query NOK**                             
`{% setcontent records = 'articles' where { region:region.id } orderby 'titre' %}`
Here, 'titre' is a text field. It is the same if I sort with select fields or other filed types.      

These queries are correct with bolt 3.7, but failed with bolt 4.1. This only works when I delete the "orderby" clause or replace it with `orderby -publishedAt`                                

### Actual result

```
[...]

PR #2320, opened 2021-01-15
Migrate `JSON_GET_TEXT` and `CAST` in `doctrine.yaml` since 4.1.11

I-Valchev:
Adds the migrations for the recent fix for #2241 and an order change to `JSON_GET_TEXT`


bobdenotter (approved):
👍

PR #2322, opened 2021-01-15
Don't require user to re-authenticate after changing user locale

I-Valchev:
Fixes a bug that requires the user to re-authenticate in order to apply the locale settings.


bobdenotter (approved):
Awesome!

PR #2325, opened 2021-01-18
Bolt generates pretty URLs without `contentTypeSlug` in path or query params

### LINKED ISSUE #2261: Pretty URLs without ContentType
Support "pretty" short URLs without the ContenType for multiple/all ContentTypes, like so:

```
/pages/about-us -> /about-us
/reviews/john-doe -> /john-doe
/posts/how-to-learn-js -> /how-to-learn-js
```

**Known issues**
There is a conflict if two records of different ContentTypes have the same slug. However, in most cases this is an edge-case. 

To achieve this, the `contentTypeSlug` in DetailController should be optional. Then in routing we should warn the user of potential slug clash, _**but**_ allow them to take the risk, if they choose to.

I-Valchev:
Fixes #2261

Examples that work with this PR:

```
/page/hello-world -> /hello-world
/nl/page/hallo-wereld -> /nl/hallo-wereld
```

To configure this, put 2 routes in `routes.yaml`:

```yaml
content_seo_locale:
  path: /{_locale}/{slugOrId}
  methods: [GET|POST]
  defaults:
      _controller: Bolt\Controller\Frontend\DetailController::record

content_seo:
    path: /{slugOrId}
    methods: [GET|POST]
    defaults:
        _controller: Bolt\Controller\Frontend\DetailController::record

```

**Note: Naming is important here.** The localised route should be `{{ routeName }}_locale`, whereas the non-localised (generic) route should be `{{ routeName }}`. 

Then, for each Content Type that should have pretty URLs, set `record_locale` to `{{ routeName }}_locale`. This way, Bolt's existing URL shortcut for the default locale will be used. Otherwise, you'd generate links like `/en/hello-world` even for the default locale.

Exceptions can still be handled if you want to use `contentTypeSlug` by using Twig's `path` or the UrlGenerator in code.

### DIFF
--- a/src/Canonical.php
+++ b/src/Canonical.php
@@ -208,7 +208,7 @@ class Canonical
     {
         $routes = $this->router->getRouteCollection();
 
-        /** @var Route $routeDefinition */
+        /** @var Route|null $routeDefinition */
         $routeDefinition = $routes->get($route);
 
         if (! $routeDefinition) {
--- a/src/Canonical.php
+++ b/src/Canonical.php
@@ -211,6 +211,10 @@ class Canonical
         /** @var Route $routeDefinition */
         $routeDefinition = $routes->get($route);
 
+        if (! $routeDefinition) {
+            return false;
+        }
+
         /** @var CompiledRoute $compiledRoute */
         $compiledRoute = $routeDefinition->compile();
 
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):

bobdenotter:
Looking good!

PR #2333, opened 2021-01-22
Fix YouTube oEmbed missing `title` accessibility issue

### LINKED ISSUE #2331: YouTube embed video accessibility issue
It looks like there's a known issue with YouTube embeds missing the `title` attribute.

Check more here: https://adrianroselli.com/2017/08/inaccessible-youtube-embeds.html

Since we already fetch a title to show in the backend and elsewhere, it would make sense to include it in the Embed by default so it doesn't get flagged by accessibility tools such as Lighthouse

I-Valchev:
Fixes #2331


bobdenotter (approved):
👍

PR #2341, opened 2021-01-24
update/simplify permissions.yaml

simongroenewolt:
Update config in permissions.yaml so it will be (more or less) backwards compatible. Use only ROLE_ADMIN and ROLE_EDITOR in config to keep things simple, commented content-type specific configuration. Updated roles available in editor.


bobdenotter:
Looks good! Thank you, @simongroenewolt 👍

bobdenotter (approved):

PR #2347, opened 2021-01-29
Fix `record|thumbnail` getting different image than `record|image`

I-Valchev:
With special thanks to @mcdennem for working with me to find and fix this


bobdenotter (approved):
Cool! 👍

PR #2350, opened 2021-01-29
Fix new checkbox value on existing records

I-Valchev:
Fixes an issue where a new checkbox added to an existing record is always (wrongly) set to true.
Fixes the `default` value option for a checkbox to work.


bobdenotter (approved):
Sweet! 👍

PR #2356, opened 2021-02-01
Remove (abandoned) `sensiolabs/security-checker`

bobdenotter:


I-Valchev (approved):
👍

toofff (commented):
Why not replace it with [this repository](https://github.com/fabpot/local-php-security-checker) that Symfony recommends instead of just removing the dependency?

I-Valchev:
Hi @toofff , our understanding is that the security checker is very similar to `roave/security-advisories`, hence no need to keep two dependencies. Is that an incorrect conclusion in your view? @toofff

toofff:
@I-Valchev if it does the same job then yes :)

On the other hand it is not put in the CI, is this normal?

bobdenotter:
> Why not replace it with this repository that Symfony recommends instead of just removing the dependency?

It's not a matter of simply replacing it. The recommendation is for something tat's meant as a standalone tool, not something to do a one-on-one replacement with. We _could_ integrate that in the CI, but regardless the `sensiolabs/security-checker` has got to go. 
And like Ivo mentions: I don't think there's added value in having _both_ `roave/security-advisories` as well as using `github.com/fabpot/local-php-security-checker`.

toofff:
@bobdenotter 

The pull request did not mention the use of the `roave/security-advisories` package, that's why I commented on the pull request. But if the other package does the job, then yes ;)

On the other hand, could it be good to add it to the CI as a non-blocking job?

I-Valchev:
@toofff According to the documentation https://github.com/Roave/SecurityAdvisories, it already works on update, which we already do.

I don't if it is added explicitly though, in fact it looks like the better solution. Do you want to create a PR for it?

In the meantime, let's merge this in.

paras-malhotra:
@I-Valchev you can consider the [Enlightn Security Checker](https://github.com/enlightn/security-checker).

PR #2361, opened 2021-02-04
Use TemplateSelect `filter` option with directories

### LINKED ISSUE #2358: TemplateSelect filter doesn't seem to work with folders
For example, @nestordedios and I couldn't make it work for something like this:

```
parks/park_1.twig
parks/park_2.twig
```

Adding a regex like `/^parks\/park_.*.twig$/`, but it wouldn't work. `/park_.*.twig$/` on the other hand worked just fine.

I-Valchev:
Fixes #2358


bobdenotter (approved):
Whoop whoop! 👍

PR #2365, opened 2021-02-04
`ContentEditController` cleanup

I-Valchev:
Depends on https://github.com/bolt/common/pull/35

(And also I think I might look into a bit more refactoring of this before merging in)


bobdenotter (approved):
👍

PR #2373, opened 2021-02-08
Make multiselect fields iterable in Twig

### LINKED ISSUE #2368: Select field with multiple values has empty loop
Details
-------

|----------------- | ------
| Relevant Bolt Version |  4.1.11 
| PHP version                 | 7.4 |
| For UX/UI issues        | Chrome 88

Reproduction
------------

We have created a select field type with multiple paramter to true like this :
```
services:
            label: Ajouter des services
            type: select
            multiple: true
            autocomplete: true
            sortable: true
            group: Services
            values:
              {
                  mail-management: "Gestion, réexpedition, scan de votre courrier",
                  registered-mail: "Prise en charge de votre recommandés",
                  call-management: "Gestion personnalisée de vos appels/messages",
                  administrative-assistance: "Assistance administrative",
                  financial-support: "Support comptable, financier et juridique",
                  creation: "Aide à la création et la gestion d'entreprise"
              }
```
In database, we have this with three values selected:
`["it-support","registered-mail","administrative-assistance"]`

But when I dump field information in front I have this:

image

Finally, it doesn't loop like this for the values:
```
{% for value in solution.services %}
    {% dump(value) %}
{% endfor %}
```

For a fast fix I loop on select option and test if selected is true:
```
{% set listServices = select_options(solution.services) %}
{% for service in listServices %}
    {% if service.selected == true %}
            //Code
    {% endif %}
{% endfor %}
```

I don't found why there is no value on field information, have you an idea ?

I-Valchev:
Fixes #2368

### DIFF
--- a/src/Entity/Field/CollectionField.php
+++ b/src/Entity/Field/CollectionField.php
@@ -9,8 +9,8 @@ use Bolt\Entity\Field;
 use Bolt\Entity\FieldInterface;
 use Bolt\Entity\FieldParentInterface;
 use Bolt\Entity\FieldParentTrait;
+use Bolt\Entity\IterableFieldTrait;
 use Bolt\Entity\ListFieldInterface;
-use Bolt\Entity\ListFieldTrait;
 use Bolt\Repository\FieldRepository;
 use Doctrine\ORM\Mapping as ORM;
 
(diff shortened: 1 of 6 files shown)


bobdenotter (approved):
Clever! 👍

PR #2375, opened 2021-02-08
Content `toArray()` method returns fields according to definition

I-Valchev:
This is step 1 to making sure the content exporter works properly with CSVs.


bobdenotter (approved):
Cool! 👍

PR #2378, opened 2021-02-09
Keep cache timestamps for all `.env` files

### LINKED ISSUE #2125: Config.php timestamping for cache only uses .env, and not the (newer) other config possibilities.
`getConfigFilesTimestamps()` in `Config.php` records timestamps of config files to know when to re-build the cache. This code checks the various bolt config files and the .env file as well. Since DotEnv support has been improved there are (quite a lot) of .env files that could be in use (along with a cached php env file)

I guess the cache for the bolt config file should take changes to these into account as well. I don't see a way to use DotEnv itself to report the files that are used.

Details
-------

| Relevant Bolt Version                                 |  master
| Install type                                          |  GitHub checkout
| <abbr title="Backwards Compatibility">BC</abbr> Break | no (?)
| PHP version                                           | 7.2 / 7.3 / 7.4 |
| Web server                                            | any

I-Valchev:
Fixes #2125

### DIFF
--- a/src/Configuration/Config.php
+++ b/src/Configuration/Config.php
@@ -147,12 +147,30 @@ class Config
             }
         }
 
-        $envFilename = $this->projectDir . '/.env';
-        if (file_exists($envFilename)) {
-            $timestamps[$envFilename] = filemtime($envFilename);
+        return array_merge($timestamps, $this->getEnvFilesTimestamps());
+    }
+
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Nice catch! 👍

I-Valchev:
@simongroenewolt 's to thank for finding out about this :-)

PR #2385, opened 2021-02-10
Selects with `multiple: true` always have an array value

### LINKED ISSUE #2384: Content Select with `multiple: true` within a Set returns a single value if only one is selected
A content select with `multiple: true` should always return an array, even if only one record has been selected in the Editor

I-Valchev:
Fixes #2384


bobdenotter (approved):
👍

PR #2400, opened 2021-02-16
Don't show old fields in a `set` if they're removed from the definition

### LINKED ISSUE #2387: Fields in a set won't get deleted when removed from contenttypes.yaml
Hello,

I'm using Bolt 4 with PHP 7.3, and I keep getting the same bug over and over. When I delete or rename a field inside a set, the deleted/renamed field still show up in the already created content when I edit it. To get rid of it, I have to rename my set field, to sort of "re-create" the set, but if I rename it back to it's original name, the old fields show up again.

Here is how to reproduce the issue:

And here is what happen when I rename my "page_content" field to "page_content_2", the old field still show up, along with the "new/renamed" field:

Am I doing something wrong? Or is that a bug? The entries still seem to exist even if they are deleted from contenttypes.yaml.

Thank you!

I-Valchev:
FIxes #2387

### DIFF
--- a/src/Entity/Field/SetField.php
+++ b/src/Entity/Field/SetField.php
@@ -83,7 +83,10 @@ class SetField extends Field implements FieldInterface, FieldParentInterface, Li
     {
         $fieldsFromDefinition = $this->getFieldsFromDefinition();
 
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
tenor

PR #2404, opened 2021-02-17
Remove direct references to fields from twig templates

### LINKED ISSUE #1871: Remove direct references to `field.(something)` from twig templates
The `*.html.twig` where `*` is a bolt field should not use `field.` directly.

This is because in several places reuse them by overriding properties.

So if somewhere we use this for example:
```
{% set setPath = field.definition.get('upload')|placeholders({'contenttype': record.contenttype}) %} 
```

It should become:

```
{% if setPath is not defined %}
    {% set setPath = field.definition.get('upload')|placeholders({'contenttype': record.contenttype}) %} 
{% endif %}
```

tambait:
Fixes #1871

### DIFF
--- a/templates/_partials/fields/file.html.twig
+++ b/templates/_partials/fields/file.html.twig
@@ -7,7 +7,9 @@
 {% endset %}
 
 {% block field %}
-    {% set setPath = field.definition.get('upload')|placeholders({'contenttype': record.contenttype}) %}
+    {% if setPath is not defined %}
+        {% set setPath = field.definition.get('upload')|placeholders({'contenttype': record.contenttype}) %}
+    {% endif %}
     {% set directory = path('bolt_async_upload', {'location': location|default('files'), 'path': setPath}) %}
     {% set filelist = path('bolt_async_filelisting', {'location': location|default('files') }) %}
     {% set labels = {
--- a/templates/_partials/fields/filelist.html.twig
+++ b/templates/_partials/fields/filelist.html.twig
@@ -7,7 +7,9 @@
 {% endset %}
 
 {% block field %}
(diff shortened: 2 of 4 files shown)


tambait:
@I-Valchev 
What was causing /how did you fixed `API & Behavioural Tests (Behat) / E2E Tests`  :)

Not sure if I'm missing something, buit there are also, for example, variables for `field.definition.get('limit')` or `field.definition.get('extensions')` but I wasn't adding conditional for those since all have `default`. Although I found in the code something like below, so not sure if conditional is needed when `default` is defined for a variable.

```
{% if multiple is not defined %}
    {% set multiple = field.definition.get('multiple')|default ? 'true' : 'false' %}
{% endif %}
```

bobdenotter:
Hi @tambait,

> What was causing erorr/how did you fixed API & Behavioural Tests (Behat) / E2E Tests

Sometimes it just doesn't pass. Restarting it will then often make it work the second time. :-)

bobdenotter:
Hi @tambait, 

This PR is still marked as 'draft'. Is there more work to be done, or do you think it's 'ready to go'? 

We can always merge this one in as-is, and if needed do a follow-up PR. :-)

tambait:
> Hi @tambait,
> 
> This PR is still marked as 'draft'. Is there more work to be done, or do you think it's 'ready to go'?
> 
> We can always merge this one in as-is, and if needed do a follow-up PR. :-)

Hi @bobdenotter, 
I wasn't sure if some of the variables I mentioned in earlier comment should be also modified, but as you suggested maybe it's best to merge current modification and if needed I could return to the PR when I become more familiar with Bolt :)

bobdenotter (approved):
Thanks! Let's roll with this, then! :-)

PR #2407, opened 2021-02-18
Fix taxonomy `listing_template` inconsistency. Fix the canonical for `slug` and `singular_slug`

### LINKED ISSUE #2406: Canonical URL for taxonomy listing pages is not unique
For example, the canonical for those two pages is different:

`/group/main`
`/groups/main`

And, in addition as discussed on Slack, the template used for each is different too.

I-Valchev:
Fixes #2406


bobdenotter (approved):
Whoop! 👍

PR #2416, opened 2021-02-22
Don't break collection looping if definition changes

I-Valchev:
Special thanks to @nestordedios for solving this weirdness!


bobdenotter (approved):
👍

PR #2418, opened 2021-02-23
Use more specific selector for the `setcontent` [no] test page

I-Valchev:
Turns out that a lot of the times there's dumps in the Symfony debug toolbar that contain the `[no]` from the Twig template, which tricks the tests to assume there's an error.


bobdenotter (approved):
6d8956db30a984d9b964020f26315c5e

I-Valchev:
@bobdenotter "Behat says no" 🤡

Issue #2427, opened 2021-02-24
Global search: accented search queries don't return any results


EmilePerron:
In the documentation, the following statement is made:
> Searching is case insensitive, and normalised. This means that a search for 'CAFE' or 'cafe' is equal, and either will show records that have the word 'Café' in it.

However, in our project, the normalized aspect doesn't seem to hold up.

Details
-------

| Question                      | Answer
|----------------- | ------
| Relevant Bolt Version | 4.1
| Install type                   | Composer install
| BC Break                      | no
| PHP version                 | 7.3
| Web server                  | Apache
| Database                     | MySQL

Reproduction
------------

### Bug summary

In the global search, searching for any word with accented characters returns no results at all. 

For example, if you have a page with a redactor field that contains the following text:
> En s’abonnant au Musée du Fjord, vous contribuerez à l’élargissement des publics du Musée, à la qualité et à l’originalité de ses productions, ainsi qu’à son rayonnement international. Être abonné du Musée donne aussi droit à de nombreux privilèges.

You would expect the page to show up when searching for the words `privilèges` or `originalité`. However, it does not.

### Specifics

I haven't tested on any other versions, but the problem occurred on Bolt 4.1.13.

At first glance, this seems to be due to the UTF-8 encoding of accented characters in the `value` field of the `bolt_field_translation`. 

The runnable database query for the search, based on Symfony's profiler, looks like this:
```sql
SELECT b0_.id AS id_0, b1_.id AS id_1, b1_.name AS name_2, b1_.sortorder AS sortorder_3, b1_.version AS version_4, b1_.type AS type_5 
FROM bolt_content b0_ 
INNER JOIN bolt_field b1_ ON b0_.id = b1_.content_id AND b1_.type IN ('generic', 'block', 'number', 'collection', 'embed', 'email', 'select', 'imagelist', 'templateselect', 'set', 'textarea', 'html', 'markdown', 'hidden', 'slug', 'file', 'filelist', 'checkbox', 'text', 'date', 'image', 'data', 'article', 'seo', 'redactor') 
INNER JOIN bolt_field_translation b2_ ON b1_.id = b2_.translatable_id 
WHERE b2_.value LIKE '%privilèges%' 
ORDER BY b1_.sortorder ASC;
```

However, the `value` of that redactor field in `bolt_field_translation` looks like this:
```json
[...]

PR #2435, opened 2021-02-26
Allow search queries containing special characters (fix for #2427)

EmilePerron:
Hey folks!

This is a fix for the issue I described in #2427.
I added search term encoding & escaping to match the format of the content in `bolt_field_translation`.

There doesn't seem to be any tests for this repository/method, but I tested the change on a Bolt project with: 
- the examples mentioned in the issue
- other strings with accented/special characters
- regular strings without special characters
- email adresses
- numbers

... and they all worked fine. 

If there are tests set up for this and I just missed them somehow, just let me know and I'll take the appropriate steps.

Cheers!


bobdenotter (approved):
Hi @EmilePerron ! 

Maybe someday we'll have a full-fledged search, that works across different databases. For now, your fix seems fine. Thanks! 👍

PR #2454, opened 2021-03-04
Identical record slugs in `menu.yaml` don't create conflicting URIs

### LINKED ISSUE #2432: Menu disregards contentType and bugs on 'identical' slugs
Using the built in menu builder, it is possible to create a menu that utilises links to contentTypes and records of contentTypes. However, when there exist two records with the same name (hence slug) in different contentTypes, the menu builder does not generate a link using `<a href='{{ item.uri }}'>Link</a>`. The `{{ item.uri }}` is empty.

Details
-------

|----------------- | ------
| Relevant Bolt Version |  master
| BC Break                     | N/A
| Web server                  | Built-in, symfony
| For UX/UI issues        | N/A

Reproduction
------------

### Bug summary

Using the built in menu builder, it is possible to create a menu that utilises links to contentTypes and records of contentTypes. However, when there exist two records with the same name (hence slug) in different contentTypes, the menu builder does not generate a link using `<a href='{{ item.uri }}'>Link</a>`. The `{{ item.uri }}` is empty.

### Specifics

See table above.

### Steps to reproduce

Example reproduction:
1. Create a record of contentType 'page' and name it 'contact'. Note that the slug will show `/page/contact`
2. Create a record of contentType 'block' and name it 'contact'. Note that the slug will show `/block/contact`
3. Add one item to the menu in `menu.yaml` like so:

```
main:
...
  - label: Contact
    title: Contact
    link: page/contact
```

4. As you will see, no `href` will be set (`{{ item.uri }}` is empty)

Example fix:
1. Change the name of the 'contact'-record of contentType 'block' to 'contactblock'. Note that the slug will show `/block/contactblock`
2. Add the item to the menu in `menu.yaml` as described above.
3. The link will now correctly displayed, the `href` attribute is set as `{{ item.uri }}` is not empty.

### Expected result

The link is properly generated.

### Actual result

The link is not generated.

### LINKED ISSUE #2453: Bolt menu ignores Content Type slug and only searches for record slug
If we have:
```
/page/test
/entry/test
```

And in `menu.yaml`:
```
/page/test
```

The menu URI could be either `/page/test/` or `/entry/test` because when fetching content, it does not specify the contenttype slug. That's buggy.

I-Valchev:
Fixes #2432 
Fixes #2453

### DIFF
--- a/src/Menu/FrontendMenuBuilder.php
+++ b/src/Menu/FrontendMenuBuilder.php
@@ -6,6 +6,7 @@ namespace Bolt\Menu;
 
 use Bolt\Collection\DeepCollection;
 use Bolt\Configuration\Config;
+use Bolt\Configuration\Content\ContentType;
 use Bolt\Entity\Content;
 use Bolt\Repository\ContentRepository;
 use Bolt\Twig\ContentExtension;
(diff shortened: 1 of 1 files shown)


nestordedios (approved):
This looks good! 👍

PR #2459, opened 2021-03-08
Run yaml migrations on `composer update`

### LINKED ISSUE #2451: [5.0] Versioned update scripts don't run yaml migrations
Between the versioned script changes, I've lost the `yaml-migrations` tool thingy.

We also need to check if the `yaml-migrations` folder will get downloaded/updated when updating from 4.x and from 5.0 onwards (so, if it gets there if it is missing, and if it updates once it's there already)

I-Valchev:
Fixes #2451


bobdenotter (approved):
Whoop whoop!

PR #2461, opened 2021-03-09
Temporary version lock for `api-platform` bug

I-Valchev:
The main issue, see discussion https://github.com/api-platform/core/issues/3349

image

Fixed by locking `"api-platform/core": "2.5.3"`, which then causes:

image

See https://github.com/api-platform/core/issues/3683


bobdenotter (approved):
Let's roll wit this for now! 👍

PR #2473, opened 2021-03-17
Accept less strict search like `%foo%bar%`

I-Valchev:


bobdenotter (approved):
As discussed, let's roll with this! 👍

PR #2475, opened 2021-03-18
The `|excerpt` filter returns the excerpt in the correct locale

### LINKED ISSUE #2474: |excerpt filter always shows in the default locale when an excerpt_format is set for a contenttype
When I set an `excerpt_format` with specific fields in a `contenttype`, whenever I use the filter `|excerpt` in a template, no matter what the current `locale` is, it will always show the excerpt text in the default locale

### Details
Relevant Bolt Version : 4.1.15
Install type: Composer install
PHP version: 7.3
Web server: Apache
Database: MySQL

### Reproduction
For example, this contenttype:
```

expos:
    name: Expositions et activités
    singular_name: Exposition ou Activité
    show_in_menu: Expositions et Activités
    fields:
        texte:
            type: redactor
            localize: true
        other_fields:
            [...]
    taxonomy: [ expo_categories ]
    viewless: false
    viewless_listing: true
    record_template: exposition-detail.twig
    excerpt_format: "{texte}"
    default_status: published
    order: -publishedAt
    records_per_page: 50
    icon_many: "fa:camera"
    icon_one: "fa:camera"
    locales: ['fr', 'en']

```

I set the `excerpt_format` to `"{texte}"`, to make sure the excerpt of the record is made from the `texte` field. The field has content for the `fr` locale and also has content for the  `en` locale.

Now on my page, even if the `current locale` is set to `en`, `record|excerpt` will always show the content in my default locale, which is `fr`.

### My code in my template file:
image

### The result:
image

This is especially problematic in the global search. The results shown can be from different contenttypes, none of them has the same `excerpt_format`, so there is no work around.

Am I doing something wrong? Please tell me if I'm not clear of if you need more information.

Thank you!

I-Valchev:
Fixes #2474

### DIFF
--- a/src/Twig/ContentExtension.php
+++ b/src/Twig/ContentExtension.php
@@ -255,7 +255,7 @@ class ContentExtension extends AbstractExtension
         }
 
         if (ContentHelper::isSuitable($content, 'excerpt_format')) {
-            $excerpt = $this->contentHelper->get($content, $content->getDefinition()->get('excerpt_format'));
+            $excerpt = $this->contentHelper->get($content, $content->getDefinition()->get('excerpt_format'), $this->request->getLocale());
         } else {
             $excerpt = $this->getFieldBasedExcerpt($content, $length, $includeTitle);
         }


M-Stgelais (approved):
That works perfectly and fixes the issue

bobdenotter (approved):
👍

PR #2476, opened 2021-03-18
Allow `--roles` option using the `bin/console bolt:add-user` command

I-Valchev:
Example:

`php bin/console bolt:setup ivo my_password ivo@example.org Ivo --roles=ROLE_ADMIN --roles=ROLE_USER --roles=ROLE_FOO` as well as interactive.


bobdenotter (approved):
Neat! 👍

PR #2483, opened 2021-03-23
Fix the `allow_twig` option to work in the preview

I-Valchev:
Fixes a few things related to the preview, among which:

- the `allow_twig`
- makes sure the `record` global variable is set in the preview
- makes sure there's _minimal_ differences between the preview and the actual record handling

LESS CODE IS MORE 🚀


bobdenotter:
Closing this, because I broke it. See #2486 instead.

PR #2491, opened 2021-03-25
Use form and shape in addition to colour for content statuses

I-Valchev:
**Before:**
Screenshot 2021-03-25 at 11 57 08
Screenshot 2021-03-25 at 11 56 47

**After:**
Screenshot 2021-03-25 at 11 54 46
Screenshot 2021-03-25 at 11 59 12

Bit of an ugly SCSS implementation, but it makes sure it's done across the board and keeps simple `status is-published`, `status is-held`, etc. classes.

### DIFF
--- a/assets/scss/modules/base/_status.scss
+++ b/assets/scss/modules/base/_status.scss
@@ -1,26 +1,67 @@
 //** Base | Status
 
+.apply-fa {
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
✅ Looks good

PR #2492, opened 2021-03-25
Don't hide last ellipsis dot, search button covers input height

I-Valchev:


bobdenotter (approved):
I love small changes like this, that makes things look ever so slightly developed with an eye for detail!

PR #2495, opened 2021-03-25
Add `copy to clipboard` for suggested password. Fix view password glitch

I-Valchev:
image

And fixes a glitch where clicking the eye icon moved it a couple of pixels.


bobdenotter (approved):
👍

PR #2499, opened 2021-03-30
On `Forbidden`, only redirect to the dashboard if the user is granted access to it. Otherwise show the `403`

bobdenotter:


I-Valchev (approved):
As discussed, good way to handle this gracefully! 👍

PR #2500, opened 2021-03-31
Only show generic `Invalid credentials` error if error is credentials related

I-Valchev:
Don't show that a username is "guesstimated" by switching between `Username not found` and `Invalid credentials` exception. Instead, only ever show `Invalid credentials` for credentials issues.


src/Security/LoginFormAuthenticator.php
```diff
@@ -154,6 +156,11 @@ public function onAuthenticationSuccess(Request $request, TokenInterface $token,
 
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception): RedirectResponse
     {
+        // Don't reveal a UsernameNotFound exception.
+        if ($exception instanceof BadCredentialsException || $exception instanceof UsernameNotFoundException) {
```
bobdenotter, line 160: Isn't it enough to do this, since in the other case it already is `BadCredentialsException`? 

```suggestion
        if ($exception instanceof UsernameNotFoundException) {
```

bobdenotter (approved):
👍

PR #2501, opened 2021-04-01
Tweaking users and permissions

bobdenotter:
Applying some spit-polish to the default users/permissions structure. Most notably: 

 - The first user is now `developer`, 
 - `CAN_SWITCH_USER` is no longer hard-coded to `ROLE_DEVELOPER` 
 - `ROLE_USER` is now assignable, because otherwise the user would lose this role if you edit it
 - Tweaked some of the default permissions


src/DataFixtures/UserFixtures.php
```diff
@@ -120,17 +120,17 @@ private function getUserData(): array
...
-                'email' => 'developer@example.org',
-                'roles' => ['ROLE_DEVELOPER'],
+                'displayname' => 'Eddie Enduser',
+                'username' => 'eddie',
+                'password' => $this->append ? Str::generatePassword(10) : 'eddie%1',
+                'email' => 'eddie@example.org',
+                'roles' => ['ROLE_USER'],
             ],
         ];
     }
```
I-Valchev, line 136: @bobdenotter General question: given that now all roles (not just `ROLE_ADMIN`) have access to the backend, should all these accounts be enabled by default? Can it be an issue in the event somebody setting up a site forgets to delete all the fixture users?

bobdenotter: I'm assuming one of these happens: 

 - Somebody runs `make db-reset` (us developing, or automated tests) and `admin` has `admin%1` as password. 
 - Somebody runs `bin/console bolt:setup`, and the first `developer` user gets setup. Then, the fixtures will add these users, but with a proper, encoded password. 

So, in practice, this shouldn't be a problem.

I-Valchev: Ah, good point! Thanks! 👍

I-Valchev (approved):
👍

PR #2515, opened 2021-04-07
Fix bug where `anyField` search no longer works on listing

I-Valchev:
That's one of the things that would be good to have a test for, perhaps 🤔

### DIFF
--- a/src/Controller/Frontend/ListingController.php
+++ b/src/Controller/Frontend/ListingController.php
@@ -91,7 +91,7 @@ class ListingController extends TwigAwareController implements FrontendZoneInter
         $allowedParams = array_merge(
             $contentType['fields']->keys()->all(),
             $contentType['taxonomy']->all(),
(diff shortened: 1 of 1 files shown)


src/Controller/Frontend/ListingController.php
```diff
@@ -91,7 +91,7 @@ private function parseQueryParams(Request $request, ContentType $contentType): a
         $allowedParams = array_merge(
             $contentType['fields']->keys()->all(),
             $contentType['taxonomy']->all(),
-            ['order', 'earliest', 'latest', 'offset', 'page', 'random', 'author']
+            ['order', 'earliest', 'latest', 'offset', 'page', 'random', 'author', 'anyField']
```
I-Valchev, line 94:
```suggestion
            ['order', 'earliest', 'latest', 'offset', 'page', 'random', 'author', 'anyField', 'anything']
```

bobdenotter (approved):
Good catch 🙈

PR #2518, opened 2021-04-07
Allow parsing of `localdate`s like `now` or `today + 1 week, 21:00`

### LINKED ISSUE #2516: Allow |localdate() filter to format "now" date
Hi!

It seems the Twig filter `|localdate()` does not support format of dates like `"{{ now"|localdate("j F Y") }}`. This breaks the rendering of the template.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 5.0 beta 3
| PHP version                 | 7.3  |
| Web server                  | Symfony local server

Reproduction
------------

Use `{{ "now"|localdate("j F Y") }}` in your template.

### Expected result

Considering the `"now"` string would return the date `April 7, 2021 16:11` I would expect the date to be formatted as `7 April 2021`

### Actual result

Broken page with a RuntimeError exception thrown.

bobdenotter:
Fixes #2516

### DIFF
--- a/src/Twig/LocaleExtension.php
+++ b/src/Twig/LocaleExtension.php
@@ -169,7 +169,7 @@ class LocaleExtension extends AbstractExtension
         } elseif (empty($dateTime)) {
             $dateTime = Carbon::now();
         } else {
-            $dateTime = Carbon::createFromTimeString($dateTime);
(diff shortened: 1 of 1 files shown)


I-Valchev (approved):
Nice catch!  👍

PR #2523, opened 2021-04-09
Show `404` instead of Exception when missing secret

bobdenotter:
Fixes this error / exception: 

Schermafbeelding 2021-04-09 om 15 26 48


I-Valchev:
Ah, good catch! I saw it a couple of times, but wasn't sure where it's coming from or why... any ideas?

PR #2533, opened 2021-04-14
Improve contrast ratio in filemanager

### LINKED ISSUE #1997: File manager accessibility issues
image

nestordedios:
Fixes #1997.

The button contrast seems to be already fixed. This PR only fixes the contrast of the text. See screenshot.

Before:
before

After:
after


I-Valchev (approved):
👍  and as a side benefit, cleaner code with less inlined CSS! Cool!

Issue #2429, opened 2021-02-25
General Accessibility


I-Valchev:
For all PRs and discussions regarding to accessibility in general, which do not fit specifically into one of the existing accessibility issues.

Vondry:
as part of https://github.com/bolt/core/pull/3751 there is new `eslint-plugin-vuejs-accessibility` plugin to eslint which enforces some of the accessibility improvements

PR #2534, opened 2021-04-14
Improve backend search box accessibility

nestordedios:
This PR is part of #2429 .

It adds a label to the search input field visible only to screen readers .


I-Valchev (approved):
Nice! These gradual improvements are neat, traceable, and ever so slightly improving the backend 😇

PR #2537, opened 2021-04-20
Allow `SelectField` options from callables and services

I-Valchev:
Two new options:

**Option 1: Static callable**

Example `contenttypes.yaml` configuration:

```yaml
        books:
            type: select
            values: App\Books::getOptions
```

Allows you to implement an `App\Books` class like so (where the `$field` argument is optional):

```
<?php

declare(strict_types=1);

namespace App;

use Bolt\Entity\Field\SelectField;

class Books
{
    public static function getOptions(SelectField $field): iterable
    {
        return [
            'Animal Farm',
            '1984',
        ];
    }
}
```

**Option 2: Callable as a Symfony service (with autowiring and autoloading)**

Example `contenttypes.yaml` configuration:

```yaml
        books:
            type: select
            values: App\Books
```

Allows you to implement an `App\Books` class like so (where the `$field` argument is optional):
**Important**: the `App\Books` service **must** explicitly be [public](https://symfony.com/doc/current/service_container/alias_private.html#marking-services-as-public-private), so that it does not get automatically removed from the container if unused 

```php
<?php

declare(strict_types=1);

namespace App;

class Books
{
    public function __construct(SomeNiceService $service)
    {
        $this->service = $service;
    }

    public function getOptions(SelectField $field): iterable
    {
        // $this->service is available here
        return $this->service->getBooks();
    }
}
```
_Note:_ the method name `getOptions` is significant in this case.

### DIFF
--- a/src/Entity/Field/SelectField.php
+++ b/src/Entity/Field/SelectField.php
@@ -8,6 +8,7 @@ use Bolt\Entity\Field;
 use Bolt\Entity\FieldInterface;
 use Bolt\Entity\IterableFieldTrait;
 use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Tightenco\Collect\Support\Collection;
 
 /**
@@ -18,6 +19,9 @@ class SelectField extends Field implements FieldInterface, RawPersistable, \Iter
     use IterableFieldTrait;
 
     public const TYPE = 'select';
+
+    /** @var ContainerInterface|null */
+    private static $container = null;
 
     public function setValue($value): Field
     {
@@ -66,7 +70,24 @@ class SelectField extends Field implements FieldInterface, RawPersistable, \Iter
 
     public function getOptions()
     {
-        return $this->getDefinition()->get('values');
+        $values = $this->getDefinition()->get('values');
+
+        // Check if it is a service
+        if (self::$container && self::$container->has($values)) {
+            $class = self::$container->get($values);
+            // the name of the function
+            $func = 'getOptions';
+
+            return $class->{$func}($this);
+        }
+
+        // Check if it is a callable
+        if (is_callable($values)) {
+            return call_user_func_array($values, [$this]);
+        }
+
+        // Assume it's an array of values
+        return $values;
     }
 
     public function getSelected()
@@ -87,7 +108,7 @@ class SelectField extends Field implements FieldInterface, RawPersistable, \Iter
 
     public function getContentType()
     {
-        $values = $this->getDefinition()->get('values');
+        $values = $this->getOptions();
 
         if (is_string($values) && mb_strpos($values, '/') !== false) {
             return current(explode('/', $values));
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
This is f'ing brilliant! I love it! 🧡

Spidlace:
Thanks @I-Valchev , exactly what I searched !

Spidlace:
Hi @jvanoostrom 

Can you show the code of your service, please ?

jvanoostrom:
Hi guys,

Thanks for the assist! Here are my files:

My contenttype:
```
commissies:
    name: Commissies
    singular_name: Commissie
    fields:
        title:
            type: text
            class: large
            group: content
        image:
            type: image
            label: Highlighted image
        body:
            type: redactor
            height: 300px
        commissieleden:
            type: collection
            fields:
                commissielid:
                    type: set
                    label: Commissielid
                    fields:
                        rol:
                            type: text
                            label: Rol in de commissie
                        persoon:
                            type: select
                            label: Lid
                            values: App\Service\MemberCommissionLoader
                            autocomplete: true
                            limit: 500
        slug:
            type: slug
            uses: title
            group: Meta
    taxonomy: [ groups ]
    records_per_page: 5
    order: id
    record_template: templates/commissie.twig
    listing_template: templates/list_commissies.twig
    icon_many: "fa:user-tie"
    icon_one: "fa:user-tie"
```

My service:
```
<?php

namespace App\Service;

use App\Repository\MemberRepository;
use Bolt\Entity\Field\SelectField;

class MemberCommissionLoader
{

    public $memberRepository;

    public function __construct(MemberRepository $memberRepository)
    {
        $this->memberRepository = $memberRepository;
    }

    public function loadOptions()
    {
        $members = $this->memberRepository->getMemberListForCommissions();

        $memberArray = [];

        foreach($members as $member)
        {
            array_push($memberArray, $member->getEendenNaam()." (memberid:".$member->getId().")");
        }

        return $memberArray;
    }

    public function getOptions(SelectField $field): iterable
    {

        return $this->loadOptions();

    }

}
```

And in my services.yaml
```
    App\Service\MemberCommissionLoader:
        public: true
```

Thanks again!

Spidlace:
Hi @jvanoostrom,

> How can I get the actual value back in my templates? If I try to access them now, it gives me the position of the entry as it is in the list in the admin panel (eg. 'car' is third in the select options, so in my template I get 3).

In your service, in ```loadOptions```, you do an array_push but in the template, it will get the key and not the value
In my opinion, you should do (example) : 

```php
$memberArray[$member->getId()] = $member->getEendenNaam()." (memberid:".$member->getId().")");
```

And in your template, it will send you back : ```$member->getId()``` selected

jvanoostrom:
Hi @Spidlace,

Thanks for the suggestion, it works like a charm! I would, however, expect that when one would feed a simple array (such as in the example in the docs), eg, ["1984", "Animal Farm"], the actual output in the template would be the actual text (value), rather than the auto generated value based on order of appearance (key). Or is there a way to print both key and value in the template?

For my project, I needed the $member->getId() anyways, so it doesn't really matter for me.

Thanks again!

PR #2539, opened 2021-04-20
Fix datepicker to only attempt to use existing locales, ensuring Time Picking still works

bobdenotter:


templates/_partials/fields/date.html.twig
```diff
@@ -24,13 +24,18 @@
   {% endif %}
 
   {% if not valueonly %}
+      {# Ensure we use only locales bundled with Flatflicker, otherwise default to 'en' #}
+      {% set locale = app.user.locale|split('_')|first %}
+      {% if not locale in 'ar|at|az|be|bg|bn|bs|cs|cy|da|de|eo|es|et|fa|fi|fo|fr|ga|gr|he|hi|hr|hu|id|is|it|ja|ka|km|ko|kz|lt|lv|mk|mn|ms|my|nl|no|pa|pl|pt|ro|ru|si|sk|sl|sq|sr|sv|th|tr|uk|uz|vn|zh'|split('|') %}
+          {% set locale = 'en' %}
```
I-Valchev, line 30: @bobdenotter is it an overkill to use the default locale, rather than `en` here? 🤔

bobdenotter: I hardcoded it, because if the default locale would be `nl_NL`, it'll still break in that case.

I-Valchev: Ah, good point! Agree!

I-Valchev (approved):
👍

PR #2548, opened 2021-04-26
Support multiple targets for widget, keeping `$target` too

### LINKED ISSUE #1585: New Feature: Widgets should be able to have multiple targets
Currently, a widget can target only one location, like this: 

https://github.com/bobdenotter/configuration-notices/blob/master/src/ConfigurationWidget.php#L24

```php
    protected $target = AdditionalTarget::WIDGET_BACK_DASHBOARD_ASIDE_TOP;
```

`$target` should be able to take an array of multiple targets, so we can re-use a widget in different locations. 

See also #1506

When this is done, we should fix https://github.com/bobdenotter/configuration-notices/issues/3

I-Valchev:
Fixes #1585

It also keeps a single `$target` class variable, for ease of use and backwards compatibility. Apart from that, all other handling should be handled as an array of targets.

### DIFF
--- a/src/Widget/SnippetWidget.php
+++ b/src/Widget/SnippetWidget.php
@@ -11,19 +11,19 @@ class SnippetWidget extends BaseWidget
 {
     protected $name;
     protected $type;
-    protected $target;
+    protected $targets;
     protected $zone;
     protected $priority;
 
     public function __construct(
         string $snippet = '<!-- snippet -->',
         string $name = 'Nameless Snippet',
-        string $target = Target::NOWHERE,
+        array $targets = [Target::NOWHERE],
         string $zone = RequestZone::NOWHERE
     ) {
         $this->setTemplate($snippet);
         $this->setName($name);
-        $this->setTargets([$target]);
+        $this->setTargets($targets);
         $this->setZone($zone);
     }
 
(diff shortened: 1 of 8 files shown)


src/Widget/SnippetWidget.php
```diff
@@ -21,10 +21,10 @@ public function __construct(
         string $target = Target::NOWHERE,
```
bobdenotter, line 21: Should we allow `string|string[]` here?

bobdenotter (changes requested):
I can's seem to do something like this yet, right? 

```
    protected $targets = [AdditionalTarget::WIDGET_BACK_DASHBOARD_ASIDE_TOP, AdditionalTarget::WIDGET_BACK_EDITCONTENT_BELOW_HEADER];
```

(as a replacement of this: https://github.com/bobdenotter/weatherwidget/blob/master/src/WeatherWidget.php#L23 )

src/Widget/SnippetWidget.php
I-Valchev: I agree this should be changed, wonder if sticking to just `$targets` is better for consistency. Like, _always_ expect an array, without that many exceptions. So far, the only exception is in the `BaseWidget` because it's most widely used.

bobdenotter: Yeah, plural is better for consistency.. We'd just have to prevent BC-breaks if possible. If we can _say_ people should use the plural, but with a quick `if/else` we can prevent BC-breakage, that'd be best, imho. :-)

I-Valchev: Agreed! The latest changes should allow that :-)

bobdenotter (approved):
He, ho, let's go! 👍

I-Valchev:
source

PR #2551, opened 2021-04-27
Allow sensio/framework-extra-bundle version 6

bobvandevijver:
I'm not sure if PSR-7 is actually used by this bundle, but that is the only major change between v5 en v6 of `sensio/framework-extra-bundle` (see https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/710 / https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/CHANGELOG.md). 

It seems to work fine for me when forcing composer to install v6.1.2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "psr/event-dispatcher": "^1.0",
         "psr/simple-cache": "^1.0",
         "scienta/doctrine-json-functions": "^4.1",
-        "sensio/framework-extra-bundle": "^5.6",
+        "sensio/framework-extra-bundle": "^5.6 || ^6.0",
         "sensiolabs/security-checker": "^6.0",
         "siriusphp/upload": "^3.0.1",
         "squirrelphp/twig-php-syntax": "^1.5",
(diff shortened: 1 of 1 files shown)


bobdenotter:
Hi, @bobvandevijver 👋 

I personally don't like `PSR-7`, and I'm confident Symfony doesn't use it either. 

Shouldn't we just bump it to `^6.1`? I'm not a fan of having `||` dependencies in Core. (long story short, it adds overhead in resolution, and it makes providing `bolt/project` with a lock file much harder and more fragile.

bobvandevijver:
Hi 👋

Bumping it to 6.1 is fine by me 👍🏻

bobdenotter:
Cool! It tests are passing, i'll merge it in. 

Thanks!

bobvandevijver:
You might want to squash it before merge (if that isn't enabled by default), as two commits for this change might be a bit too much 😅

bobdenotter (approved):

bobdenotter:
Done, thanks again! 👍

PR #2554, opened 2021-04-28
Fix a CSS quirk in listing overview pages

bobdenotter:
Before: 

Screenshot 2021-04-28 at 17 18 20

After: 

Screenshot 2021-04-28 at 17 37 27

It would be better if it were dynamic, but that might have downsides as well (like, columns shifting). This fix makes it better, with low impact.


I-Valchev:
Neat, definitely much better already!

Have you tried it with `text-overflow: ellipsis;`? We already use that in a few places
https://github.com/bolt/core/search?q=ellipsis

bobdenotter:
I keep forgetting that support for that is now in all major browsers. 🤦 

Updated PR!

I-Valchev (approved):
👍 Nice! Let's roll with this!

PR #2566, opened 2021-05-10
Fix ecs breakage with `friendsofphp/php-cs-fixer`

I-Valchev:
Fixes `Warning: count(): Parameter must be an array or an object that implements Countable in /[...]/bolt/core/vendor/friendsofphp/php-cs-fixer/src/AbstractProxyFixer.php on line 73`


bobdenotter (approved):
👍

PR #2570, opened 2021-05-11
fix issue with migration that wouldn't run on MariaDB 10 by specifying a primary key

simongroenewolt:
see https://github.com/bolt/core/issues/2569


bobdenotter:
Hi @simongroenewolt, 

The tests are failing because of some bitrot in dependencies. If you could rebase on `master`, tests will likely pass! :-)

PR #2576, opened 2021-05-11
Several fixes for handling Relations

bobdenotter:
This PR fixes three issues related to Relations:

 - Default to `required: false`, since it makes most sense if you can _clear_ a Relation.
 - Allow deletion of a Relation from either side in the Backend
 - Fix logic error in `getFromMemory` in `RelationFactory`


I-Valchev (approved):
👍  Looks good, gradually cleaning up and fixing bugs!

PR #2582, opened 2021-05-11
fix issue with migration that wouldn't run on MariaDB 10

simongroenewolt:
by specifying a primary key

This should fix https://github.com/bolt/core/issues/2569

This has not been tested on sqlite!


bobdenotter (approved):
Did some monkeytesting, seems to work on SQLite too! 👍 

Thanks, @simongroenewolt

PR #2583, opened 2021-05-14
Renaming files for consistency

bobdenotter:


I-Valchev (approved):
👍 Looks good, let's keep it as consistent as we can!!

PR #2585, opened 2021-05-14
Autowire twig environment to Field for rendering

I-Valchev:
Also allows to use the `record` global variable in fields with `allow_twig`


bobdenotter (approved):
Neat! 👍

PR #2587, opened 2021-05-17
Fix breaking kernel and create new `EntityServiceProvider`

I-Valchev:


bobdenotter (changes requested):
Looks good, but we should also remove `m_2021-05-11-services.yaml`

I-Valchev:
@bobdenotter YES! Good spot, thanks! Removed it with the last commit.

I-Valchev:
Ugh, what a clusterfuck!

So, when autowiring twig's `Environment`, under the hook it initialises all twig extensions. It turns out, however, that the RequestStack does not _yet_ have the request object passed onto it executing this from the `Kernel`'s `boot` method.

Yet, we use it in two Twig extensions: the `ContentExtension` and the `FrontendMenuExtension`. The first had a fallback for a missing request, the latter did not.

https://github.com/bolt/core/pull/2587/files#diff-6ddaed7c4ec1937ac3db455d18baccf61ba91cad4149ae0af2fd4115773b7bd2L108

https://github.com/bolt/core/pull/2587/files#diff-56dce87e76671d948431b2603a320f447dceafef39569760f996b37da03fb78eL25

So, the next time the `FrontendMenuExtension` is used, it's not created again (as it is a reused service), hence `$requestStack->getCurrentRequest();` is not called, and there's still no request object.

🤯

I'm guessing we shouldn't use the shortcut of setting the request from the request stack in the constructor!

mcdennem (approved):

mcdennem (approved):

bobdenotter:
Looks good to me, too! Let's quickly merge it in, before it breaks again! 😅

bobdenotter (approved):

PR #2592, opened 2021-05-19
#2429 Fix some accessibility issues

DorienD:
- Add correct ID's to fields
- Change color of code to be contrast compliant
- Remove space from aria-selected
- And some more minor changes


I-Valchev (changes requested):
Hey @DorienD ,

Thanks a lot for the PR! Looking good!

One question: it seems like there are already translated labels for the 4 items here. https://github.com/bolt/core/blob/master/translations/messages.en.xlf

Am I overlooking something, or the changes to `translations/messages.en.xlf` are not needed?

DorienD:
I will look into this.

translations/messages.en.xlf
```diff
@@ -2655,6 +2655,29 @@
...
+      </segment>
+    </unit>
+    <unit id="OYUVA5." name="general.label.search">
+      <segment>
+        <source>general.label.search</source>
+        <target>Search</target>
+      </segment>
+    </unit>
+    <unit id="hKmDb7q" name="Share secure preview link">
+      <segment>
```
DorienD, line 2671: @I-Valchev I can't really figure out why it thinks I Added this translation. But its still the same as on master, so I guess it's ok?

DorienD:
@I-Valchev Fixed it, can you check it again :)

I-Valchev (approved):
@DorienD awesome, thanks! Sometimes GitHub is being a little weird with those differences, let's merge it in as it is. I'll check if it is duplicated and remove it afterwards, if necessary.

Thanks for fixing these accessibility issues 👍

mcdennem (approved):

PR #2597, opened 2021-05-30
Performance boost: Keep in-memory copy of `currentLocale`

bobdenotter:
It turns out that `getCurrentLocale` can be pretty inefficient, especially when a page has a lot of fields and/or relations. Every time _any_ field is fetched, like `{% record.title %}`, the `currentLocale` is checked, triggering a database lookup. 

See these screenshots from a real-world site, where a page has 180 related records. Before: 

Screenshot_2021-05-30_at_11_19_09

After:

Screenshot_2021-05-30_at_11_23_12

In this out-of-the-ordinary case, performance increases by a factor of 60. 

On 'regular' sites, performance gains are around 10 to 15% for detail pages.

### DIFF
--- a/src/Utils/LocaleHelper.php
+++ b/src/Utils/LocaleHelper.php
@@ -36,6 +36,9 @@ class LocaleHelper
     /** @var string */
     private $defaultLocale;
 
+    /** @var Collection */
+    private $currentLocale;
+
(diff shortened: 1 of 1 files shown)


I-Valchev (approved):
👍 Already a nice improvement, let's follow up in #2599.

PR #2603, opened 2021-06-01
Fix `|filter` and `|order` chaining returning no results

### LINKED ISSUE #2601: Using Pagerfanta (records) with Twig's `|filter` and Bolt's `|order` breaks
Something like this:

```twig
{% set records = 'showcases' limit 500 %}

{% set pages = records|filter(p => p.id > 50)|order('-priority') %}
```
^ The `pages` will be an iterator of iterators (or something..) and it won't work in the for loop correctly. We need to fix that.

I-Valchev:
FIxes #2601


nestordedios (approved):
Looks good! 👍

PR #2610, opened 2021-06-04
Pin Chromedriver to version 90.x 

bobdenotter:
We pin Chromedriver to `90.something`, because `91` _appends_ text when using `.setValue()` instead of replacing it.


nestordedios (approved):
Looks good! 👍

I-Valchev (approved):
👍 Let's hope it gets fixed soon!

PR #2611, opened 2021-06-04
Adding Jasny Twig Extensions for Bolt 5.

bobdenotter:
See https://github.com/jasny/twig-extensions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
         "embed/embed": "^3.4",
         "erusev/parsedown": "^1.7",
         "fakerphp/faker": "^1.10",
+        "jasny/twig-extensions": "^1.3",
         "knplabs/doctrine-behaviors": "^2.0.3",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -24,6 +24,7 @@
         "embed/embed": "^3.4",
         "erusev/parsedown": "^1.7",
         "fakerphp/faker": "^1.10",
+        "jasny/twig-extensions": "^1.3",
```
I-Valchev, line 27: 👍  Love these!

Wonder whether our `localdate` or the `jasny/twig-extensions` one will take precedence in a project?

bobdenotter: @I-Valchev Good call. I've changed the PR to not add Jasny's DateExtension to prevent the conflicting tag.

nestordedios (approved):
Looks good and nice addition! 👍

PR #2614, opened 2021-06-05
Add restriction for (global) Composer 2 to `composer.json`

bobdenotter:
Fixes the issue where some people bumped into upgrade problems with vague descritpions, because they were still on composer 1. 

If this is you, please upgrade to composer 2, by running `composer selfupdate --2`. 

The main difference you'll notice is that Composer 2 is an order of magnitude faster than Composer 1 is.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,6 +8,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-pdo": "*",
+        "composer-runtime-api": "^2",
         "api-platform/core": "2.5.3.*",
         "babdev/pagerfanta-bundle": "^2.5",
         "beberlei/doctrineextensions": "^1.2",


I-Valchev (commented):
I think I spotted something like this as well, wonder what causes it and if it is a breaking change?

bobdenotter:
> if it is a breaking change?

Well, opinions differ.. Some people say raising requirements is fine within SemVer. Personally, I'm leaning towards considering bumping requirements as a BC break. 

In this case, i don't think we have much choice. Our _existing_ dependencies are causing breakage on _existing_ installs. It basically requires an update to Composer 2. As in: If you don't update, it'll break in an obscure, hard to debug way. 

This change makes the new requirement much more visible, and it will also prevent the breakage _before_ it happens, in stead of afterwards.

I-Valchev (approved):
@bobdenotter gotcha, thanks for explaining!

In that case, let's roll with this 👍

nestordedios (approved):
Looks good! 👍

PR #2618, opened 2021-06-10
`extensions:configure` accepts both `.yaml` and `.yml` for config, routes and services in extensions

### LINKED ISSUE #2615: Field Type extension notice from configuration-notice
Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.1.20
| BC Break                      | no

Reproduction
------------

* Install current BoltCMS 4 via composer + demo data
* Install https://github.com/LordSimal/bolt-geolocation-field
* Add test content-type + geolocation field
* See dashboard notice on the right

Basically I already did some debugging in https://github.com/LordSimal/bolt-geolocation-field/issues/7 but couldn't get to the core because it has something to do with doctrine magic.

Bob said it probably has something to do with the core, so here is that issue 😄

I-Valchev:
Fixes #2615 

@LordSimal I'd also advise renaming it to `services.yaml` (note the A), that way it will work for people with `4.1.21` as well :-)

### DIFF
--- a/src/Command/ExtensionsConfigureCommand.php
+++ b/src/Command/ExtensionsConfigureCommand.php
@@ -64,14 +64,9 @@ class ExtensionsConfigureCommand extends Command
         // @todo: Combine this with Bolt\Extension\ConfigTrait.php
         foreach ($packages as $package) {
             $path = $this->getPackagePath($package);
-            $origin = $this->getRelativePath($path) . '/config/config.yaml';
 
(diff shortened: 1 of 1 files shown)


I-Valchev:
(I mean lower than 4.1.21) :-)

bobdenotter (approved):
Whoop!

mcdennem (approved):
👍

PR #2619, opened 2021-06-10
Fix `maximum_listing_select` in `config.yaml`

I-Valchev:


nestordedios (approved):
Good catch @I-Valchev !! 👍  We can be 100% sure this works :)

bobdenotter (approved):
Looks good! 👍

PR #2623, opened 2021-06-11
Fix error in custom code producing seemingly unrelated message. Catch `LoaderLoadException` instead of `Throwable`

### LINKED ISSUE #2586: "Cannot autowire service" error, when you have a syntax error in an extension.
When you're developing an extension, and have a file in that extension with a Syntax error, you might get this confusing error:

```
Cannot autowire service "Bolt\Article\Controller\Upload": argument "$articleConfig" of method "__construct()" references class "Bolt\Article\ArticleConfig" but no such service exists.
```

This is confusing, because `Bolt\Article` seems unrelated. Working theory is that it breaks adding the Bolt extensions to the container, and that "Article" happens to be the first it then tries to instantiate. 

Schermafbeelding 2021-05-14 om 14 13 43

### LINKED ISSUE #2622: Error in custom code produces seemingly unrelated message
If you're building a Bolt project, and you've unnoticedly made an error in your custom code, you'd expect a descriptive error. 

For example, notice the trailing comma below (which is only valid in PHP 7.3 and up, not on 7.2)

```php
<?php

namespace App;

use Bolt\Extension\BaseExtension;

class Extension extends BaseExtension
{
    public function getName(): string
    {
        return sprintf(
            '%s %s', 
            "foo", 
            "bar",
        );
    }
}
```

The error you're getting is this: 

```bash
$ bin/console bolt:info

In DefinitionErrorExceptionPass.php line 54:

  Cannot autowire service "Bolt\Article\Controller\Upload": argument "$articleConfig" of method "__constr
  uct()" references class "Bolt\Article\ArticleConfig" but no such service exists.
```

This is terribly confusing, obviously. 

PR with a fix is incoming. 

Schermafbeelding 2021-06-11 om 13 55 30

bobdenotter:
Fixes #2622
Fixes #2586

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -10,6 +10,7 @@ use Bolt\Extension\ExtensionCompilerPass;
 use Bolt\Extension\ExtensionInterface;
 use Bolt\Repository\FieldRepository;
 use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
+use Symfony\Component\Config\Exception\LoaderLoadException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\Config\Resource\FileResource;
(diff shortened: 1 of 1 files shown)


I-Valchev (approved):
👍 Nice fix! FINALLY 🥳

mcdennem (approved):
Yes!

PR #2633, opened 2021-06-16
Update Symfony components to 5.3

bobdenotter:
Despite what #2605 says, this PR actually updates our bundled Symfony components to 5.3.

Fun fact: Dependency resolution on composer update takes about 180mb less memory.


I-Valchev (approved):
👍

PR #2635, opened 2021-06-16
Filter by numeric fields in `setcontent`

### LINKED ISSUE #2620: Bolt 4 Filter / Query Search With Number Types Treated as Text
Details
-------

|----------------- | ------
| Relevant Bolt Version | 4

Reproduction
------------

### Bug summary
I am using query search to filter by price. Works great with a URL like below on a list page:

`?price=>100000`

Price is a number field and I have filled out some entries with value over 1 million however strangely when I query for over (or under) 1 million I seem to just get no results.

`?price=>1000000`

### Specifics
So entity with number field type, have tried both number and text however same result. Filled out with prices. Prices can go over £1,000,000. Content type with query_search enabled and querying the listing page with POST params like above.

### Steps to reproduce

1. Create content_type with number type field e.g. price.
2. Create records, atleast one must have a value above 1000000
3. Go to list page and filter by that field for a value above 1000000 e.g. www.bolt.com/records?price=>1000000

### Expected result

Record created with value over 1000000 should be shown in list of results.

### Actual result

Nothing shows in results.

I-Valchev:
Fixes #2620

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -24,7 +24,7 @@ parameters:
         # Parameters in Storage\Directive\OrderDirective::orderByNumericField() aren't seen as ints
         -
             message: '#of method Doctrine\\ORM\\Query\\Expr::substring\(\) expects int#'
-            path: %currentWorkingDirectory%/src/Storage/Directive/OrderDirective.php
+            path: %currentWorkingDirectory%/src/Storage/FieldQueryUtils.php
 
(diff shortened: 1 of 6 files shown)


bobdenotter (approved):
Looking good! 👍

mcdennem (approved):
Looks nice!

PR #2638, opened 2021-06-17
Decouple directive handling from the `ContentQueryParser`

### LINKED ISSUE #2637: Refactor Bolt Storage `DirectiveHandler` classes as services
All these classes:

https://github.com/bolt/core/tree/4.1/src/Storage/Directive

Mainly, the `OrderDirective.php` now requires a lot of services for it to create a proper order query. But it's not a service at the moment (having naturally grown):

https://github.com/bolt/core/blob/0916ee977a04f1147f987304421081a8957f5285/src/Storage/FrontendQueryScope.php#L83
https://github.com/bolt/core/blob/0916ee977a04f1147f987304421081a8957f5285/src/Storage/ContentQueryParser.php#L124

If we convert them into services, we won't need to instantiate them manually anymore.

I-Valchev:
Fixes #2637 in -33 lines!

image

### DIFF
--- a/src/Storage/ContentQueryParser.php
+++ b/src/Storage/ContentQueryParser.php
@@ -5,27 +5,14 @@ declare(strict_types=1);
 namespace Bolt\Storage;
 
 use Bolt\Configuration\Config;
-use Bolt\Doctrine\Version;
 use Bolt\Entity\Content;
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):

mcdennem (approved):
👍

PR #2649, opened 2021-06-21
Additional Filter / Order By Number Field Types Fix

harrigo:
Fixes issues with number field types in query parameters showing as a json array and not directly comparable as numbers. Further issues with cast detection fixed.

Fixes: #2620

### DIFF
--- a/src/Storage/SelectQuery.php
+++ b/src/Storage/SelectQuery.php
@@ -324,7 +324,7 @@ class SelectQuery implements QueryInterface
 
         $dateFields = $this->getDateFields();
 
-	$numberFields = $this->getNumberFields();
+	    $numberFields = $this->getNumberFields();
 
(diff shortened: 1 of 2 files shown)


src/Storage/SelectQuery.php
```diff
@@ -339,7 +341,7 @@ public function build(): QueryBuilder
                 $param = date('c', strtotime($param));
             }
 
-            if (in_array($fieldName, $this->regularFields, true)) {
+            if (in_array($fieldName, $this->regularFields, true) && !in_array($fieldName, $numberFields, true)) {
```
bobdenotter, line 344:
```suggestion
            if (in_array($fieldName, $this->regularFields, true) && ! in_array($fieldName, $numberFields, true)) {
```

bobdenotter (approved):

bobdenotter:
Thanks, @harrigo 👍

PR #2651, opened 2021-06-23
Allows running `bin/console` from other folders than just ProjectDir

bobdenotter:


I-Valchev (approved):
👍 Yep, that's nice to have it working

mcdennem (approved):
👍

PR #2659, opened 2021-06-29
Fix incorrect error when uploading an allowed filetype, that isn't an image

bobdenotter:
For example, you have this: 

```yaml
accept_file_types: [ …, mp4, …]

accept_media_types: [ gif, jpg, jpeg, png, svg, pdf, mp3, tiff, avif, webp ]
```

Uploading an `mp4` into a `type: file` field _should_ work, but it doesn't, because it doesn't generate a Media Entity. This PR fixes that.

Screenshot 2021-06-29 at 20 07 50

### DIFF
--- a/.github/workflows/behavioural_tests.yaml
+++ b/.github/workflows/behavioural_tests.yaml
@@ -109,5 +109,5 @@ jobs:
                 run: ./run_behat_tests.sh && make behat-js-quiet
 
             -   name: Upload Behat logs
-                run: ./bin/upload-textfiles "var/log/behat-reports/*.html"
+                run: ./bin/upload-textfiles "var/log/behat-reports/*"
                 if: always()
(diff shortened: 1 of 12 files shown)


I-Valchev (changes requested):
@bobdenotter there are a couple of these `isImage` methods now, see here:
https://github.com/bolt/core/search?q=isimage 

Should we refactor?

bobdenotter:
Yeah, let's see if we can! 

The methods do something similar, but not exactly the same (one operates on a string, the other on `Media`. But, that hardcoded list in two places in an eyesore. :-)

bobdenotter:
Updated!

bobdenotter:
This is becoming a snake pit, and the PR is getting much too large. Going to close this one, open a new one, and leave the refactoring for later. 

See #2663

PR #2660, opened 2021-06-30
Support Twig chain loaders if available

emodric:
One of our Symfony bundles that we're testing on Bolt uses a custom Twig loader.

This results in `$this->getTwig()->getLoader()` code used in Bolt in several places to return an instance of `Twig\Loader\ChainLoader` instead of `Twig\Loader\FilesystemLoader` making the code in question fail in Bolt, which results in Bolt not being able to find any of its templates.

This fix adds support for `ChainLoader`, which basically iterates over any Twig loader in the chain loader, and if it is an instance of `FilesystemLoader`, does what Bolt did previously (adds or prepends a new path, based on context).

Thanks to @ilukac for discovering the cause of the issue.

### DIFF
--- a/src/Controller/TwigAwareController.php
+++ b/src/Controller/TwigAwareController.php
@@ -23,6 +23,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Tightenco\Collect\Support\Collection;
 use Twig\Environment;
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
I haven't bumped into this yet, but I can see how this would be problematic. Thanks for the fix, @emodric and @ilukac 👍

emodric:
Thanks @bobdenotter !

PR #2663, opened 2021-07-02
Fix incorrect error when uploading an allowed filetype, that isn't an image

bobdenotter:
For example, you have this: 

```yaml
accept_file_types: [ …, mp4, …]

accept_media_types: [ gif, jpg, jpeg, png, svg, pdf, mp3, tiff, avif, webp ]
```

Uploading an `mp4` into a `type: file` field _should_ work, but it doesn't, because it doesn't generate a Media Entity. This PR fixes that.

Screenshot 2021-06-29 at 20 07 50


bobdenotter:
We'll leave the refactor for another time: #2659 #2664

I-Valchev (approved):
Sure, this is an improvement already so let's merge! :-)

mcdennem (approved):
👍

PR #2669, opened 2021-07-02
Created cypress test for display record and fixed a typo in content fixtures

Blepeian:
ContentFixtures.php had a typo on row 441: There was a double space before "with" resulting in the test failing.


I-Valchev (approved):
👍  Looking good!

bobdenotter (approved):
This is exactly how tests should function, by making an error that went under the radar for a long time actually visible so we can fix it! 👍

PR #2670, opened 2021-07-02
Created cypress test for displaying taxonomies + added cypress library for easier work

stellakasabova:
…y for easier work


I-Valchev (approved):
Looks good, nice addition with the library 🙌

bobdenotter (approved):
Good call on adding that library! 👍

PR #2675, opened 2021-07-05
Fix `If condition is always true` on new phpstan

I-Valchev:


bobdenotter (approved):
Always fun when a new version of PHPstan finds new things to fix! 🤣

PR #2688, opened 2021-07-08
Created cypress test for record localization and fixed a formatting error in the edit page's header

Blepeian:


I-Valchev (approved):
Ugh, have to admit Gherkin/Behat had a nicer way of testing tables... but, the tests look good!

Better to have properly behaving tests than tests that just look a little better 😉

bobdenotter (approved):

templates/content/edit.html.twig
```diff
@@ -19,8 +19,7 @@
 
 {# The 'title' and 'shoulder' blocks are the main heading of the page. #}
 {% block shoulder %}
-    {{ (app.request.get('_route') == 'bolt_content_duplicate') ? 'caption.duplicate'|trans : 'caption.edit'|trans }}
-    {{ record.definition.singular_name }}
+    {{- (app.request.get('_route') == 'bolt_content_duplicate') ? 'caption.duplicate'|trans : 'caption.edit'|trans }} {{ record.definition.singular_name -}}
```
nestordedios, line 22: Line 22 seems hard to read since it holds the ternary operator and a property that prints something.

I would keep it separate in two lines as it was.

nestordedios:
Hahaha!! Whatever! @bobdenotter already merged it 😂

templates/content/edit.html.twig
bobdenotter: I think this whitespace was what was tripping up the tests :-)

Let's do a small follow-up PR where we use ` -}} {{- ` to keep 'm on two lines, but without Twig adding the whtespace

nestordedios: Ah ok, I did not know this 😓 hehe.

Blepeian: About keeping them one two lines
At first I tried using different combinations of `-}} {{-` and `~}} {{~` to keep it on two lines, but adding these would either remove all spaces between the resulting words (example: Word1Word2) or it would remove everything between them except the new line (example: Word1\nWord2) neither of which is the desired result of Word1 Word2 
So I'm not sure if they can be used in this case, since what's needed is a single space to be left

bobdenotter: The _really_ ugly way to go around that could be to use  `{{- ' ' ~ record.definition.singular_name -}}`

But, that really sort of defeats the purpose of making it legible! 😅 

Let's just keep it like this. It's not like it's terrible.

nestordedios: Yes I think that is the best. 

I didn't know it would be so complicated and what the reason of having it in one line was. 

Better to leave as it is 👍 .

PR #2696, opened 2021-07-12
Created cypress tests for editing users and removed users.feature

Blepeian:


nestordedios:
Hi @Blepeian ,

The tests you wrote do look good but some other tests seem to be failing. 

Those tests which are failing need to be fixed before we approve and merge. 

The [logs](https://github.com/bolt/core/runs/3046077752?check_suite_focus=true) in the terminal show where the test is failing, but I don't really know how to fix it.

Maybe @I-Valchev can help out here.

PR #2715, opened 2021-07-19
[4.x] Simplify, don't mangle username on incorrect input for Login

bobdenotter:
input `admin@example.org` becomes: 
Screenshot 2021-07-19 at 20 15 35


I-Valchev (approved):
👍 Good catch

mcdennem (approved):
👍

PR #2716, opened 2021-07-19
[5.x] Simplify, don't mangle username on incorrect input for Login

bobdenotter:
This is  #2715, but for Bolt 5.0.


I-Valchev (approved):
👍 Likewise, good catch

mcdennem (approved):
👍

PR #2727, opened 2021-07-23
Fix issue with incorrectly named `autocomplete` option

### LINKED ISSUE #2720: `type: select` needs `searchable: true` in order to autocomplete
This should be fixed.

I-Valchev:
Fixes #2720 and makes `autocomplete: true` the default


bobdenotter (approved):
Neat!

PR #2737, opened 2021-07-29
Fix `group: sidebar` conflicting with Bolt's sidebar

### LINKED ISSUE #2735: group `sidebar` does not work
Grouping fields in a content group called `sidebar` does not work, because Bolt's sidebar already uses that id.

I-Valchev:
This will cause the 5.0 cypress tests to fail. Will fix separately.

Fixes #2735


nestordedios (approved):
Looks good! 👍

bobdenotter (approved):
👍

PR #2740, opened 2021-07-29
Cleanup behat after moving to cypress

I-Valchev:
This is a partial cleanup of the behat setup. The full cleanup can be made after #2739


nestordedios (approved):
👍

bobdenotter (approved):

PR #2745, opened 2021-08-03
Update the warning color to a11y compatible

I-Valchev:
Changing orange to red, because it was the only accessibility-compliant color that I could find 🤷‍♂️


nestordedios (approved):
Looks good! 👍

bobdenotter (approved):
🧧

PR #2750, opened 2021-08-04
Unlimited select field options

JTNMW:
Allows the number of options displayed in select fields to be set by the limit value in the contenttypes.yaml

### DIFF
--- a/assets/js/app/editor/Components/Select.vue
+++ b/assets/js/app/editor/Components/Select.vue
@@ -7,7 +7,8 @@
             :limit="1000"
             :multiple="multiple"
             :options="options"
-            :searchable="taggable"
+            :optionsLimit="optionslimit"
+            :searchable="autocomplete || taggable"
             :show-labels="false"
             :taggable="taggable"
             :disabled="readonly"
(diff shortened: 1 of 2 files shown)


bobdenotter:
Hi @JTNMW, 

CI seems to be complaining about the line with `optionsLimit`, because it disapproves of CamelCase. I wonder if it's treated as case sensitive, or if we could just use `optionslimit` there too? 🤔

JTNMW:
hi @bobdenotter ,

I followed the syntax given in the vue-multiselect documentation https://vue-multiselect.js.org/#sub-props. I'm not sure how strict these need to be followed. I'll give it a go in lower case and let you know if i still get the same results.

I-Valchev (changes requested):
Looks solid, thanks @JTNMW.

If you would like to also introduce the `autocomplete` option in 4.2, I'd suggest checking here: https://github.com/bolt/core/pull/2727/files

A few things are missing in the current PR. It maybe best to not introduce this here and, instead, create a separate PR that makes the autocomplete work for 4.2 :-)

bobdenotter:
Thanks,  @JTNMW 👍 

Let's roll with this! 

> It maybe best to not introduce this here and, instead, create a separate PR that makes the autocomplete work for 4.2 :-)

If we feel we should backport fixes, let's do so in a separate PR!

bobdenotter (approved):

PR #2767, opened 2021-08-12
Search now also finds words with special characters, like `büßen`

### LINKED ISSUE #2758: Search for strings containing german special characters (Umlaute)
Bolt Search does not find content with german special characters (Umlaute)

Details
-------

|----------------- | ------
| Relevant Bolt Version | 4.2
| PHP version                 | 7.3 |
| For UX/UI issues        | Not relevant

Reproduction
------------
Entering words  with german special characters into a text or html field like _"Äpfel", "Prüfung", "daß", "Ölpreis"_ etc. and later searching for those strings thru the Bolt search functionality (search?searchTerm=Äpfel) doesn't return any matches.

However performing the **same search thru the admin panel works** and returns the content containing the special
characters as expected.

This bug appeared during migration from bolt v3 to bolt v4.2. However it not only happens with imported (conimex) content 
from Bolt v3 but **also when entering new content** containing special characters in Bolt 4.2. So it seems to be a general
problem.

If you need more information, feel free to contact me. Thank you in advance.

bobdenotter:
Fixes #2758

afbeelding

### DIFF
--- a/src/Repository/ContentRepository.php
+++ b/src/Repository/ContentRepository.php
@@ -105,7 +105,7 @@ class ContentRepository extends ServiceEntityRepository
 
         // The search term must match the format of the content in the database
         // Therefore, it is JSON encoded and escaped with backslashes
-        $encodedSearchTerm = addslashes(trim(json_encode($searchTerm), '"'));
+        $encodedSearchTerm = addslashes(trim(json_encode($searchTerm, JSON_UNESCAPED_UNICODE), '"'));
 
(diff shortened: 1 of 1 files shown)


mcdennem (approved):
Nice 👍

eherber:
Hi Bob,

thank you for fixing this bug.

I already tested the fix and it works fine, german special characters could now be used in the search query string
and results are found.

Thank you.

-Eric

PR #2781, opened 2021-08-19
feat: separate the persist logic from the back-office controller

Deuchnord:
(Work in progress)

Bolt has an event mechanism that allows developers to get notified and execute code when content or users are getting changes. But these events are currently dispatched by the `Bolt\Controller\Backend\ContentEditController` controller. As a consequence, developers cannot take advantage of these events in their own code.

In this PR, I propose to create two new _data manager_ classes (one per data type: user and content) that would take care of the task to persist the entities, and dispatch the relevant events.
This PR introduces four new events too, `{PRE,POST}_PERSIST` and `{PRE,POST}_FLUSH`, so the developers can subscribe to new intermediate states without breaking the backward compatibility.

### DIFF
--- a/src/Controller/Backend/ContentEditController.php
+++ b/src/Controller/Backend/ContentEditController.php
@@ -9,6 +9,8 @@ use Bolt\Common\Json;
 use Bolt\Configuration\Content\ContentType;
 use Bolt\Controller\CsrfTrait;
 use Bolt\Controller\TwigAwareController;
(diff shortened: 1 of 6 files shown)


bobdenotter:
Neat! 👀

I-Valchev:
Very cool stuff, @Deuchnord 👏

Deuchnord:
Just need to write tests, and I'll make this RFR.

PR #2788, opened 2021-08-23
Preview is frontend zone, fix for viewing content in another locale

I-Valchev:
Fixes an issue causing the backend locale to be interpreted as the frontend locale during preview, rolls back #2772.

Good example of how refactoring the `ContentEditController` will be good!


nestordedios (approved):
Nice @I-Valchev 👍 , It wouldn't be bad to rip out that huge Controller class if possible.

bobdenotter (approved):

PR #2817, opened 2021-09-03
Fix missing id's for integer/float field

Joossensei:
Id was missing for the integer and float field which broke both of the labels.


bobdenotter (approved):
👍

Issue #2813, opened 2021-09-02
Remove dummy users (apart from admin) now that they can login to the backend


I-Valchev:

I-Valchev:
Fixed by #2819

PR #2819, opened 2021-09-06
Addressed remove dummy users issue.

Joossensei:
Instead of removing the dummy users (#2813) the status is set to disabled by default. Only admin's status is set to enabled so it can access the backend.


bobdenotter (approved):
Nice way to fix this! Thanks! 👍

Issue #2748, opened 2021-08-03
setcontent condition not using json_extract


matesko:
Bolt 5 beta12

When fetching content via setcontent if you have two fileds one of which is 
```
       showInMenu:
            type: select
            values: 
                yes: 'Yes'
                no: 'No'
```
and another 

```
        inMenu:
            type: checkbox
```

then do:

```
    {% setcontent ps = 'pages' where { status:'published', 'showInMenu': 'yes' } orderby 'menuOrder'  printquery %}
    {% setcontent ps1 = 'pages' where { status:'published', 'inMenu': true } orderby 'menuOrder' printquery %}
```

you get following result:

```
SELECT content FROM Bolt\Entity\Content content LEFT JOIN content.fields fields_showInMenu LEFT JOIN
 fields_showInMenu.translations translations_showInMenu LEFT JOIN content.fields fields_order_2 LEFT JOIN 
fields_order_2.translations translations_order_2 WHERE content.contentType = :ct0 AND (content.status = :status_1 AND 
(translations_showInMenu.value = :showInMenu_1 AND fields_showInMenu.parent IS NULL AND fields_showInMenu.name = 
:field_showInMenu)) AND fields_order_2.name = :order_2 ORDER BY CAST(SUBSTRING(translations_order_2.value, 3, 
LENGTH(translations_order_2.value)) as decimal) ASC

    ct0: pages
    status_1: published
    showInMenu_1: yes
    field_showInMenu: showInMenu
    order_2: menuOrder

SELECT content FROM Bolt\Entity\Content content LEFT JOIN content.fields fields_inMenu LEFT JOIN fields_inMenu.translations 
translations_inMenu LEFT JOIN content.fields fields_order_3 LEFT JOIN fields_order_3.translations translations_order_3 WHERE 
content.contentType = :ct0 AND (content.status = :status_1 AND (JSON_EXTRACT(translations_inMenu.value, '$[0]') = :inMenu_1 
AND fields_inMenu.parent IS NULL AND fields_inMenu.name = :field_inMenu)) AND fields_order_3.name = :order_3 ORDER BY 
CAST(SUBSTRING(translations_order_3.value, 3, LENGTH(translations_order_3.value)) as decimal) ASC

    ct0: pages
    status_1: published
    inMenu_1: 1
    field_inMenu: inMenu
    order_3: menuOrder`

```

ps will be empty and ps1 will be correct resultset.

[...]

PR #2820, opened 2021-09-06
issue #2748 - select filters

matesko:
How to deal with issue #2748


bobdenotter:
hi @matesko,

Thanks for the patch! Could you run ECS on it, so we can keep the coding style consistent? 

```
```

PR #2828, opened 2021-09-13
Fix querying selects and multiselects

### LINKED ISSUE #2806: {{ setcontent }} with where clause on a select field does not fetch records
Hi all!

It seems that `setcontent` is not working when trying to fetch records using the where clause on a select field.

Details
-------

| Relevant Bolt Version                                 | 5.0
| Install type                                          | Composer
| Web server                                            | Symfony Local server

Reproduction
------------

Given the following Contenttype configuration:
```
wikipaginas:
    name: Wiki Paginas
    singular_name: Wiki Pagina
    slug: wikipaginas
    singular_slug: wikipagina
    fields:
        project:
            type: select
            values: projecten/{projectnummer},{title}
            keys: id
            autocomplete: true
            sort: -projectnummer
            class: narrow
            variant: inline
            postfix: <br><hr>
            searchable: true
```
and trying to fetch records like `{% setcontent wikipaginas = 'wikipaginas' where { project : record.id } %}` does not return anything.

### Expected result

Fetch existing records which match this query

### Actual result

No record are fetched even though there are matches in the database

I-Valchev:
Fixes #2806 select for MySQL where we can use `JSON_SEARCH`.

Doesn't yet fix the multiselect, but there's the groundwork. Not yet sure how to best parse the multiple expression...

### DIFF
--- a/yaml-migrations/m_2021-09-16-doctrine.yaml
+++ b/yaml-migrations/m_2021-09-16-doctrine.yaml
@@ -1,0 +1,10 @@
+# Adding the JSON_SEARCH for Bolt 5
+file: packages/doctrine.yaml
+since: 4.2.0
+
+add:
+    doctrine:
+        orm:
+            dql:
+                string_functions:
+                    JSON_SEARCH: Scienta\DoctrineJsonFunctions\Query\AST\Functions\Mysql\JsonSearch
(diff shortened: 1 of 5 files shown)


yaml-migrations/m_2021-09-16-doctrine.yaml
```diff
@@ -0,0 +1,10 @@
+# Adding the JSON_SEARCH for Bolt 5
+file: packages/doctrine.yaml
+since: 4.2.0
```
bobdenotter, line 3:
```suggestion
since: 5.0.2
```

bobdenotter (changes requested):

I-Valchev:
Yes good catch thanks @bobdenotter

bobdenotter (approved):
👍

PR #2831, opened 2021-09-17
Ensure admin/developer option is respected

bobvandevijver:
When providing a complete add-user command (`bin/console bolt:add-user --admin admin admin admin@localhost Admin`), the `interact` method is not called, resulting in the admin and developer options to be completely ignored. By detecting their presence in the `execute` method as well, the user is created correctly.

Workaround till merged/released is to use the `bin/console bolt:add-user --role=ROLE_ADMIN admin admin admin@localhost Admin` with the role argument instead.

### DIFF
--- a/src/Command/AddUserCommand.php
+++ b/src/Command/AddUserCommand.php
@@ -240,7 +240,11 @@ class AddUserCommand extends Command
         $plainPassword = $input->getArgument('password');
         $email = $input->getArgument('email');
         $displayName = $input->getArgument('display-name');
-        $roles = $input->getOption('roles');
+        
+        // Always merge the admin/developer role when the parameter has been given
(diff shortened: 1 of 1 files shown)


I-Valchev (changes requested):
Hey @bobvandevijver , thanks for the PR!

Should we then remove these lines: https://github.com/bolt/core/blob/master/src/Command/AddUserCommand.php#L207-L210

bobvandevijver:
@I-Valchev I would not do so because (if I'm correct) that would always trigger the interactive role question, even when you provided the `admin` or `developer` option.

I-Valchev (approved):
👍 Thanks, in that case let's roll with this!

bobdenotter (approved):
Just in time for 5.0.2 ;-)

PR #2833, opened 2021-09-17
Added save-method to RelationFactory. Closes #2832

### LINKED ISSUE #2832: Add save-method to RelationFactory
Add a missing save-method to RelationFactory to streamline it with ContentFactory. This method will persist the Relation entity or collection of Relation entities.

andersbjorkland:
Added save-method to RelationFactory, and tests to check if the save method persists a single Relation as well as a collection of Relation.

### DIFF
--- a/src/Factory/RelationFactory.php
+++ b/src/Factory/RelationFactory.php
@@ -7,18 +7,24 @@ namespace Bolt\Factory;
 use Bolt\Entity\Content;
 use Bolt\Entity\Relation;
 use Bolt\Repository\RelationRepository;
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Tightenco\Collect\Support\Collection;
 
 class RelationFactory
 {
+    /** @var EntityManagerInterface */
+    private $em; 
+
     /** @var RelationRepository */
     private $repository;
 
     /** @var Collection */
     private $relations;
 
-    public function __construct(RelationRepository $repository)
+    public function __construct(RelationRepository $repository, EntityManagerInterface $em)
     {
+        $this->em = $em;
         $this->repository = $repository;
         $this->relations = collect([]);
     }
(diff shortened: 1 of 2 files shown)


bobdenotter (approved):
**You included tests!! 💛**

bobdenotter:
Could you run ECS and PHPStan on your branch? 

```
vendor/bin/ecs check tests/php --fix
vendor/bin/phpstan --memory-limit=1G analyse -c phpstan.neon src
```

andersbjorkland:
Ah, I see I didn't run ECS. I did run phpstan though.

andersbjorkland:
> Could you run ECS and PHPStan on your branch? 
> 
> ```
> vendor/bin/ecs check tests/php --fix
> vendor/bin/phpstan --memory-limit=1G analyse -c phpstan.neon src
> ```

Yeah, I'll do that. I'm sitting on a train, so just gotta dig up the charger.

andersbjorkland:
What is up with `Ignored error pattern #Unreachable statement - code above always terminates# `?
image

bobdenotter:
> What is up with Ignored error pattern #Unreachable statement - code above always terminates# ?

Sometimes we have a PHPstan, that is either a false positive, or we think it's _fine_. So we add it to this file to ignore: https://github.com/bolt/core/blob/master/phpstan.neon#L14

However, if an ignore is _never_ triggered, PHPstan _also_ flags it as a potential issue. So, i think in this case, the bug that resulted in the false positive might be fixed, you got a new PHPStan tag pulled in, and it flags it. 

If  that's the only failing test, we'll merge in your PR, and fix PHPStan in a follow up PR.

andersbjorkland:
PHPUnit 7.2 didn't like my tests. I can write my arrow functions as regular anonymous functions instead 🤔

bobdenotter:
Yes, let's.. 

With Bolt, we're sticking with the same minimum requirements as Symfony, which means that we should support PHP 7.2.9, for the time being.

andersbjorkland:
Had to debug my unit test. I filtered an array and reset the index of the array - is what I thought. I actually just reset the pointer. So I switched out the reset()-function for the array_values()-function in the method getNonRelatedEntryIds.

bobdenotter:
Thanks again, @andersbjorkland

PR #2835, opened 2021-09-19
Added possible null return for getId. Closes #2834

### LINKED ISSUE #2834: Relation getId is not specified for null
getId() can return null for entities that are not yet persisted. The method getId() in the Relation-class specifies that it will return an int. This can trigger a TypeError when this method is called. One use-case when this can happen is Unit Testing where we may check if a certain entity is persisted in the database by checking if the id has been set.

andersbjorkland:
Added ?int for return type of Relation getId().

### DIFF
--- a/src/Entity/Relation.php
+++ b/src/Entity/Relation.php
@@ -77,7 +77,7 @@ class Relation
         $toContent->addRelationsToThisContent($this);
     }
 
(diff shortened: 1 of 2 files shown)


bobdenotter (approved):
Excellent! Thank you, @andersbjorkland 🍻

PR #2849, opened 2021-09-30
Fixed not rendering ellipsis

Joossensei:
Default pager didnt render the ellipsis because it was missing a definition for path


bobdenotter (approved):
Aaaaaiighhht 👌

PR #2854, opened 2021-10-06
Add `cascade={"persist"}` to Field#content relation

bobdenotter:
When programmatically creating content, this: 

```
$this->entityManager->persist($this->currentrecord);
```

Could give the following error:

zoho-error

This PR fixes this error.


mcdennem (approved):
👍

nestordedios (approved):
Yes! This works! 👍

PR #2889, opened 2021-10-26
Cache "related options"

bobdenotter:
Another one split off from #2853. Will probably give merge conflicts, as they affect the same bit of code in `ContentSaveSubscriber`, but that shouldn't be too hard to fix. 

This speeds up the generation of "Related options" a lot, especially on databases with more than a few hundred items. It's what shows up under `@bolt/content/_relationships.html.twig` in the Profiler's Timeline.


bobdenotter:
Superseded by #2957

PR #2904, opened 2021-11-03
Redirect 'GET on Preview' link to edit page

bobdenotter:
Sometimes clients or editors paste links to specific pages with a URL they've copy/pasted from a preview. Like `https://example.org/preview/123`. 

This will give an error like: 

image

This PR allows a `GET` request to that page (assuming it's published, similar to the existing "detail view")

### DIFF
--- a/src/Controller/Frontend/DetailController.php
+++ b/src/Controller/Frontend/DetailController.php
@@ -36,6 +36,7 @@ class DetailController extends TwigAwareController implements FrontendZoneInterf
      *     name="record_locale",
      *     requirements={"contentTypeSlug"="%bolt.requirement.contenttypes%", "_locale": "%app_locales%"},
      *     methods={"GET|POST"})
+     *
      * @param string|int $slugOrId
      */
     public function record($slugOrId, ?string $contentTypeSlug = null, bool $requirePublished = true, ?string $_locale = null): Response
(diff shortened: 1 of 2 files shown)


nestordedios (approved):
Nice fix! Looks good 👍

I-Valchev (changes requested):
I like the idea of fixing that scary error screen... but I'm not sure if rendering the saved version for a preview link is the way to go.

If somebody copies the preview and shares it, they'd expect the unsaved changes to be shown, which we can't do.

Perhaps we could either show a nicer error/explanation why sharing the preview does not work that way, or redirecting to the backend 🤔

bobdenotter:
The thing is, the people grabbing these links and sharing them will not see this nicer error, because it already works _for them_. it only breaks for the people _receiving_ the link, and i don't think they'll benefit from a link to the backend.. 

On the other hand, if we redirect it to the correct "Edit" page for that record.. At least you'd end up in the correct spot without having to dig much deeper. Would that be a good compromise? 🤔

I-Valchev:
@bobdenotter yes, that would make sense! Sure, I'm on board 🚂 🚋 🚋

bobdenotter:
@I-Valchev Updated!

I-Valchev (approved):
👍 Yesss, looks good! Let's roll.

PR #2908, opened 2021-11-08
Block active sessions from disabled users

I-Valchev:


bobdenotter (approved):
Such a nice little fix! 👨‍🍳

PR #2911, opened 2021-11-08
Enable 'Remember me' by default, and make duration configurable

bobdenotter:
Note, the diff on `src/Form/LoginType.php` seems large, but apparently every line in that file was indented.


I-Valchev (approved):
✅  This is so smart, me likes!

nestordedios (approved):
Nice PR! 🎉 Looks good! 👍

PR #2912, opened 2021-11-08
Multiple sessions reflect correctly in the `Current sessions` area

I-Valchev:


I-Valchev:
(Adding BC for now: if we go ahead with this, we can add the `security.yaml` migration pre-release)

src/Security/LogoutListener.php
```diff
@@ -19,31 +21,32 @@ class LogoutListener implements LogoutHandlerInterface
...
-        ];
-
-        if ($user->getUserAuthToken()) {
-            $userArr['token_id'] = $user->getUserAuthToken()->getId();
-            $this->em->remove($user->getUserAuthToken());
-            $this->em->flush();
-        }
+        ]);
 
         $this->logger->notice('User \'{username}\' logged out (manually, auth_token: {token_id}, {ip})', $userArr);
```
bobdenotter, line 51:
```suggestion
```

bobdenotter (changes requested):

PR #2915, opened 2021-11-08
Don't fail on upsert or create if programatically creating content without authentication

I-Valchev:


bobdenotter (approved):
Nice catch! 👍

PR #2919, opened 2021-11-09
Fix unique slugs if title is empty

### LINKED ISSUE #2718: Saving with empty title can generate empty slug
127 0 0 1_8000_bolt_edit_10

If you leave the title empty. It generates an empty slug. 
The empty slug persists after the first save unless changed manually.

This is an issue, because users may be inclined to think of a title later after which the slug is already set.
Doing so causes the "view saved version" link to reload the backend page in a new tab instead of opening the actual saved version because the slug is invalid. There is no signal to point users to the actual issue, which is the empty slug.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 5.3.2 
| Install type                   |   Composer install 
| Web server                  | Apache / Nginx / Built-in, version [ version ] < really?
| For UX/UI issues        | Brave > Chromium <<<< really this doesnt matter here

Reproduction
------------

create record without title > save > forget to change slug > attempt to open saved version of page

### Bug summary

empty slugs don't throw errors

### Specifics

It's a simple issue. doesn't get more specific than this.

### Steps to reproduce

* create record without title 
*  save 
*  forget to change slug
*  attempt to open saved version of page

### Expected result

some kind of error or notice slugs are not allowed to be empty if I try to open saved version

### Actual result

record edit page keeps opening in new tab

*sorry for the originally large picture format

### LINKED ISSUE #2809: Empty slug field when saving a record without title
If your contenttype uses a "title" field to create a slug (Bolt's out of the box behaviour) no slug is added when a user saves the record without adding a title.
Also, when adding a title later, the slug field remains empty.

I made a screencast to show you what happens:
https://www.dropbox.com/s/8kdcd7yyigcaw9d/Screen%20Recording%202021-08-30%20at%2016.19.27.mov?dl=0

This results in records not having a valid url in the frontend templates.

---
Bolt 5.0.0 b14

I-Valchev:
Fixes #2809 
Fixes #2718

### DIFF
--- a/src/Event/Listener/ContentFillListener.php
+++ b/src/Event/Listener/ContentFillListener.php
@@ -92,14 +92,14 @@ class ContentFillListener
             $slugField = null;
         }
 
-        $fields = $slug ? $this->fieldRepository->findAllBySlug($slug) : null;
+        $safe = true;
 
-        if (! $fields) {
-            // No slug field with that slug exists. We're done here.
-            return;
+        if (! $slug) {
+            $slug = $this->getSafeSlug($content->getContentTypeSingularSlug());
+            $safe = false;
         }
 
-        $safe = true;
+        $fields = $slug ? $this->fieldRepository->findAllBySlug($slug) : null;
 
         // Clone fields for each locale
         $tempFields = [];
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Huzzah! 🎉

PR #2922, opened 2021-11-10
An empty image alt made the page throw an error

Joossensei:
If you create a field with an empty alt attribute it returns an error in the standard skeleton theme


bobdenotter (approved):
Thanks! :+1:

PR #2943, opened 2021-11-17
Fix regex for email pattern

### LINKED ISSUE #2940: Email pattern does not validate subdomain emails
It seems that when using the `pattern: email` in a text field it does not validate emails with subdomains like `nestor@mysubdomain.example.org`

Details
-------

| Relevant Bolt Version                                 | 5.0.3
| Web server                                            | Symfony server

Reproduction
------------

Add the pattern option to a text field like: 
```
email:
    type: text
    label: "Email address"
    pattern: email
```

Try entering an email like `nestor@mysubdomain.example.org` and save the record.

### Expected result

The record is saved with the text `nestor@mysubdomain.example.org` as the value of the email field

### Actual result

A pop up that says `Please match the requested format: fields[email].` appears under the email field and the record can not be saved.

bobdenotter:
Fixes #2940

### DIFF
--- a/config/bolt/contenttypes.yaml
+++ b/config/bolt/contenttypes.yaml
@@ -208,6 +208,7 @@ showcases:
         email:
             type: email
             info: "This is an <i>info box</i> shown as a popover next to the field label."
+            pattern: email
         embed:
             type: embed
             group: Media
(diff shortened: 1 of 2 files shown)


I-Valchev (approved):
👍 Cool!

PR #2953, opened 2021-11-20
Bumping versions of Glide and PHPStan

bobdenotter:


src/Command/CopyAssetsCommand.php
```diff
@@ -77,22 +77,20 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             try {
                 $this->filesystem->remove($targetDir);
                 $this->hardCopy($originDir, $targetDir);
-
+                dump($originDir, $targetDir);
```
I-Valchev, line 80: Ooops, that probably shouldn't be here @bobdenotter ?

src/Api/Extensions/ContentExtension.php
```diff
@@ -69,7 +69,7 @@ private function filterUnpublishedViewlessContent(QueryBuilder $queryBuilder): v
         $queryBuilder->setParameter('status', Statuses::PUBLISHED);
 
         //todo: Fix this when https://github.com/doctrine/orm/issues/3835 closed.
-        if (! empty($this->viewlessContentTypes)) {
+        if ($this->viewlessContentTypes->isNotEmpty()) {
```
I-Valchev, line 72: 👍 I like this a lot, manly because of this: https://www.zachstronaut.com/posts/2009/02/09/careful-with-php-empty.html

I-Valchev (changes requested):

src/Twig/FieldExtension.php
```diff
@@ -121,7 +121,9 @@ public function getLabel(Field $field): string
 
     public function getType(Field $field): string
     {
-        return $field->getType() ?? $field->getDefinition()->get('type');
+        // Note, this used to be a ternary, with `?? $field->getDefinition()->get('type');`. I'm not sure
```
I-Valchev, line 124: Hm, looks like it was left as a "safe" choice in #1709 . Which means it's probably safe as-is now.

src/Twig/FieldExtension.php
```diff
@@ -131,7 +133,7 @@ public function getSelected(Field\SelectField $field, $returnsingle = false, $re
     {
         $definition = $field->getDefinition();
 
-        if ($definition->get('type') !== 'select' || ! $field->isContentSelect() || ($field->isContentSelect() && $definition->get('mode') === 'format')) {
+        if ($definition->get('type') !== 'select' || ! $field->isContentSelect() || $definition->get('mode') === 'format') {
```
I-Valchev, line 136: @bobdenotter Doesn't this one have a different truth table compared to the one on the left? 🤔

I-Valchev (changes requested):

src/Command/CopyAssetsCommand.php
bobdenotter: Thx!

src/Twig/FieldExtension.php
bobdenotter: Ok!

src/Twig/FieldExtension.php
bobdenotter: if `$field->isContentSelect()` is `false`, it falls through in the second part of the statement. It it _reaches_ the third part, we can already be sure that `$field->isContentSelect()` is `true`, so it is meaningless.

Joossensei (approved):
Looks good 👍

I-Valchev (approved):
👍 Looks good

PR #2957, opened 2021-11-22
Toggle cache for related options

I-Valchev:


bobdenotter (commented):
Looks good at first glance, but I'm going to actually check this one out and monkey-test it, before moving forward! :-)

bobdenotter (approved):
Really like this approach! Let's roll with it!

PR #2967, opened 2021-11-23
Fixed a few cypress tests again

Joossensei:


Joossensei:
Tests are failing here https://github.com/bolt/core/blob/aa2701083af309abf586aa4ea6f9bad0bcde23c1/src/Doctrine/Version.php#L34-L46 

But didnt edit these files so something in Bolt I presume

Joossensei:
Made new PR with more fixes for Cypress

PR #2968, opened 2021-11-24
Remove some of the describe from Cypress / Fix some common Cypress errors

### LINKED ISSUE #2961: Cypress tests improvements
I think we can improve duration and readability for the cypress test by removing some of the `describe`.  Just discovered you can use multiple `it` inside a single `describe`.

I think it would look cleaner and maybe even shave some time off the cypress tests.

With multiple `it`: 
Screenshot 2021-11-23 at 10 14 04

Without multiple `it`:   
Screenshot 2021-11-23 at 10 14 28

Joossensei:
Fixes #2961


bobdenotter (approved):

bobdenotter:
Looks good! 👍

PR #2976, opened 2021-11-26
Make it possible to add custom class in Select field

### LINKED ISSUE #2975: It is not possible to add custom class to Select field
Having a Select field defined like it follows, won't add the class `number-attribution-group` to the field.

```
selectfield:
    type: select
    values: [ foo, bar, baz ]
    class: number-attribution-group
```

Details
-------

| Relevant Bolt Version                                 | 5.0.6
| Web server                                            | Symfony local server

Reproduction
------------

Add a custom class to a select field like for example:
```
selectfield:
    type: select
    values: [ foo, bar, baz ]
    class: number-attribution-group
```

### Expected result

The custom class is added to the select field.

### Actual result

The custom class is **not** added to the select field.

nestordedios:
Fixes #2975 

It seems only the `"wide-options"` class was being added by default. I've kept this class to be rendered since there is some CSS targeting that class.

### DIFF
--- a/templates/_partials/fields/select.html.twig
+++ b/templates/_partials/fields/select.html.twig
@@ -25,7 +25,7 @@
 
 {% block field %}
     <editor-select
-        :classname="{{ "wide-options"|json_encode }}"
+        :classname="{{ ['wide-options ', class]|join|json_encode }}"
         :value="{{ value|json_encode }}"
         :name='{{ name|json_encode }}'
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Thanks, @nestordedios 

The failures in CI are caused already fixed in a later PR, so i'm going to go ahead and merge this in!

PR #2977, opened 2021-11-26
Fix getting db version (lock `doctrine/dbal`)

I-Valchev:
Fixes this:
image


bobdenotter (approved):
Let's roll with this, and see how we can use `dbal` 3 later.

PR #2981, opened 2021-11-28
Check on null value added to calling setCurrentLocale to prevent null value parameter

nicodeboer:


bobdenotter (approved):
Thanks, @nicodeboer 👍 

The fails in CI are fixed in #2977, so i'm going to merge this one in.

PR #2984, opened 2021-11-29
Allow symfony/flex v2

bobvandevijver:
Symfony/flex v2 should be compatible with Bolt, so lets allow it. See https://github.com/symfony/flex/releases/tag/v2.0.0 for the release notes, it is basically a versions which only support composer 2.1+ & PHP 8.0+.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "symfony/dependency-injection": "^5.3",
         "symfony/dotenv": "^5.3",
         "symfony/expression-language": "^5.3",
-        "symfony/flex": "^1.17",
+        "symfony/flex": "^1.17|^2",
         "symfony/form": "^5.3",
         "symfony/framework-bundle": "^5.3",
         "symfony/http-client": "^5.3",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -47,7 +47,7 @@
         "symfony/dependency-injection": "^5.3",
         "symfony/dotenv": "^5.3",
         "symfony/expression-language": "^5.3",
-        "symfony/flex": "^1.17",
+        "symfony/flex": "^1.17|^2",
```
bobdenotter, line 50: Tiny nitpick, let's use the same syntax as above. :-)

```suggestion
        "symfony/flex": "^1.17 || ^2.0",
```

bobvandevijver:
Don't know why the test fail for this change though..

composer.json
bobvandevijver: Doesn't matter to me, I use the single one as it is what is used by Symfony.

bobdenotter (approved):
According to @Joossensei, it's failing due to an error in Cypress, that's already fixed in master. (#2968, iirc) 

So, let's merge this one in! :shipit:

PR #2986, opened 2021-11-29
Date issues in Bolt 5

### LINKED ISSUE #2993: Date formats are stored differently in a database
I recently discovered that the dates of "publishedAt" and a normal date field get stored differently in a database.

This is the format for the publishedAt fields:
Screenshot 2021-12-02 at 11 12 44

And this is the format for the usual date field:  
Screenshot 2021-12-02 at 11 13 22

I think we should make these two formats the same.

Joossensei:
This (#2993) bug has a bunch of things that nee to be done:

- [x] Find out where the data gets saved as Zulu
- [x] Change this to the normal Y-m-d h:i format
- [x] See if date filtering works with a 'between' query ( ' >= date ' & ' <= date' )

Fixes #2993

### DIFF
--- a/assets/js/app/editor/Components/Date.vue
+++ b/assets/js/app/editor/Components/Date.vue
@@ -105,7 +105,7 @@ export default {
                 wrap: true,
                 altFormat: 'F j, Y',
                 altInput: true,
-                dateFormat: 'Z',
+                dateFormat: 'Y-m-d h:i',
                 enableTime: false,
             },
         };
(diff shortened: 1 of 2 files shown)


Joossensei:
Screenshot 2021-12-03 at 10 35 26
New format they are saved in

nestordedios (approved):
Nice @Joossensei!! It looks good! 👍

nestordedios:
@Joossensei as showed in your screenshot... Does the new format also allow for strings like "dummie-2"? I don't think that is really what we would want.

Joossensei:
@nestordedios that is a title generated by Bolt I think it isnt a date field😅

bobdenotter (approved):
👍

I-Valchev (approved):
👍 Let's roll with it!

PR #2994, opened 2021-12-02
Add new paginate function for arrays

Joossensei:
If you filter the results from a setcontent query it returns an array with the records but the normal pager puts these records on the page they used to be. This paginate function will group these records to the same page


I-Valchev (approved):
👍 Nice!

nestordedios (approved):
Looks good! 👍

PR #3004, opened 2021-12-10
Don't set "linked media", if we don't have a proper filename

bobdenotter:


nestordedios (approved):
Looks good! 👍 Conimex import is happy again!

I-Valchev (approved):
👍  Hey ho, let's go

PR #3006, opened 2021-12-10
If the given "values" for a `type: select` is not a string like `pages/{title}`, we shouldn't parse it

bobdenotter:


bobdenotter:
Let's park this one for now.. There's more to it than is seems at first sight.

giphy

simongroenewolt:
Hmm @bobdenotter this looks a lot like my issue https://github.com/bolt/core/issues/3085. Please don't just ignore the null value, as at least in my situation there _should_ not have been a null and the real problem is in the initialization of the select field.

bobdenotter:
@simongroenewolt Yes, I'm now quite certain that the underlying issue here is what you're seeing in #3085. Especially now that i'm able to reproduce it reliably.

bobdenotter:
Fixed in #3101

PR #3015, opened 2021-12-13
Fix php unit relation test

### LINKED ISSUE #3012: Failing test on PHP 7.2
This one boggles my mind: It's only failing on PHP 7.2, and I can't yet reproduce it locally:

https://github.com/bolt/core/runs/4497542902?check_suite_focus=true

Schermafbeelding 2021-12-12 om 13 15 10

I-Valchev:
Fixes #3012 

But...unit tests making a DB connection 😐 
Besides the code is pretty complex, which is also not in the spirit of unit tests:

https://github.com/bolt/core/blob/master/tests/php/Factory/RelationFactoryTest.php#L89-L105

I think we need to refactor this.


bobdenotter (approved):

bobdenotter:
For now though, let's roll with this! :-)

nestordedios (approved):
👍

PR #3022, opened 2021-12-17
[WIP] Making it easier to set field values for collections

Joossensei:


I-Valchev:
That'd be an awesome feature to have, and will make API integrations a lot cleaner in code.

I'd suggest taking a look at implementing it within the `CollectionField.php` and `SetField.php`, because this is where related logic for these fields should live. That way the `Content::setFieldValue` will also work, because it calls `setValue` of the field:

https://github.com/bolt/core/blob/master/src/Entity/Content.php#L533

So, perhaps we can take a look at these:
https://github.com/bolt/core/blob/master/src/Entity/Field/CollectionField.php#L58-L73
https://github.com/bolt/core/blob/master/src/Entity/Field/SetField.php#L33-L57

A couple hints:
1. `SetField` should be easier to do than `CollectionField`
2. Start by defining an example structure (call to the `setField->setValue()`) that will work, and then build the code to implement it

Joossensei:
Hey @I-Valchev, I think im a little lost in how I should approach this. 
I dont really know what the `$field->setValue()` should do? 
Can you give me an example of what it is supposed to be doing.

Joossensei:
Gonna close it for now maybe ill get back to it later

PR #3028, opened 2021-12-21
Make `LoginFormAuthenticator` compatible with Symfony 5.3

### LINKED ISSUE #3023: Bolt 5.1 is incompatible with Symfony 5.3
The latest Bolt version is incompatible with Symfony 5.3 due to the new authenticator system. This was added with https://github.com/bolt/core/pull/3007. 

It throws the following during the composer installation scripts:

```
PHP Fatal error:  Declaration of Bolt\Security\LoginFormAuthenticator::authenticate(Symfony\Component\HttpFoundation\Request $request) must be compatible with Symfony\Component\Security\Http\Authenticator\AuthenticatorInterface::authenticate(Symfony\Component\HttpFoundation\Request $request): Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface in /var/www/website-v2-dev/vendor/bolt/core/src/Security/LoginFormAuthenticator.php on line 43
```

Details
-------

| Relevant Bolt Version                                 | 5.1
| <abbr title="Backwards Compatibility">BC</abbr> Break | yes
| PHP version                                           | 8.0 |
| Web server                                            | Apache

I-Valchev:
Fixes #3023

### DIFF
--- a/src/Security/LoginFormAuthenticator.php
+++ b/src/Security/LoginFormAuthenticator.php
@@ -16,6 +16,7 @@ use Symfony\Component\Security\Http\Authenticator\Passport\Badge\CsrfTokenBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Credentials\PasswordCredentials;
 use Symfony\Component\Security\Http\Authenticator\Passport\Passport;
+use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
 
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Neat! 👍

bobdenotter:
> But, I have to ask, would it not be easier to just drop Symfony 5.3 and require 5.4 or higher?

I think we should definitely do that, but not just yet.. Doing so means we'll bump to PHP `^8.0` as well, which is nice, but i don't want to to it too soon. (case in point: Still quite a few stragglers on Bolt 3.x, to name one thing)

PR #3039, opened 2022-01-07
Update serializable interface

bobvandevijver:
This should prepare for the future deprecation of the `Serializable` interface.


bobdenotter (approved):
Thanks @bobvandevijver   👍

The failures in the CI are fixed in #3053

PR #3041, opened 2022-01-07
New feature: Allow empty value for select form controls, based on user setting.

vlad-ghita:
## Use case
- have a required select field;
- select field must not pre-select any option;
- user must deliberately choose an option;

Current Bolt behaviour automatically selects first option if field is required.

## Solution

Add a configuration setting called `allow_empty` which enables developer to force an empty option for select control, even when control is required.

`allow_empty` setting is available in following definitions:
- select field
- relations
- taxonomies

Image below shows definitions with these settings:
```
required: true
allow_empty: true
```

sample_relations_and_taxonomies

## Testing

Manually tested these combinations:

| required | allow_empty |
| - | - |
| none | none |
| none | true |
| none | false |
| true | none |
| true | true |
| true | false |
| false | none |
| false | true |
| false | false |

## Further improvements
- automatic validation. Given empty option is added automatically, Bolt can automatically validate select value and return an error.
- unit test

### DIFF
--- a/assets/js/app/editor/Components/Select.vue
+++ b/assets/js/app/editor/Components/Select.vue
@@ -3,7 +3,6 @@
         <multiselect
             ref="vselect"
             v-model="selected"
-            :allow-empty="allowempty"
             :limit="1000"
             :multiple="multiple"
             :options="options"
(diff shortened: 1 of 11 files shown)


bobdenotter:
Hi @vlad-ghita I'm going to review this PR. Thanks! 

I've fixed a tiny CS issue directly in the PR. It's still complaining about a PHPStan error, but this is fixed upstream, already.

PR #3046, opened 2022-01-11
Fix date format to save PM time

### LINKED ISSUE #3024: Bolt 5.1 All datetime fields are saved as AM
I just updating to 5.1 but i have a doctrine issue, i fixed whit **bin/console doctrine:schema:update --force**, so after i have the next bug:

### Example:
I select 18:00hrs (6:00 PM) in the datetime field:
image

But when i save changes, the page reload and shows 6:00 AM:
image

It happens in all my dates... so i have tu get back to v5.0.6

fsidobre:
Hello! 
The recent changes to the format handled by date & datetime fields introduced a few bugs : 

Having the `h` format instead of `H` for hour will :
- make default hour stored for date fields as `12:00` 
- make default hour for query comparison as `12:00` (e.g. when using `where { date: '>yesterday' }`)
- only store the hour in a range between 0 and 12 on a `datetime` field when you type a PM hour (with or without the AM/PM selector) 

Fixes #3024

### DIFF
--- a/assets/js/app/editor/Components/Date.vue
+++ b/assets/js/app/editor/Components/Date.vue
@@ -105,7 +105,7 @@ export default {
                 wrap: true,
                 altFormat: 'F j, Y',
                 altInput: true,
-                dateFormat: 'Y-m-d h:i',
+                dateFormat: 'Y-m-d H:i',
                 enableTime: false,
             },
         };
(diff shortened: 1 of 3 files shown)


Joossensei (approved):
Looks good! And nice that you wrote some tests for these as well!

bobdenotter (commented):
Thanks @fsidobre  👍

The failures in the CI are fixed in #3053

PR #3048, opened 2022-01-12
Fix wrong subtask called in makefile for docker

fsidobre:
Hello! 

Some cleaning was made in the `db-*` and `docker-db-*` sub-tasks, but an old one is still called, breaking the `make docker-install` procedure.

This fixes it!


bobdenotter (approved):
Thanks, @fsidobre 👍

PR #3055, opened 2022-01-21
Bugfix Fix autocomplete for Relation 

ionutlip:
it looks like searchable not used in assets/js/app/editor/Components/Select.vue   only  **:searchable="autocomplete || taggable"**


bobdenotter (approved):
This is a good catch. Thanks, @ionutlip 👍 

(CI is acting up, but locally tests pass fine)

Issue #3059, opened 2022-01-21
Attempting to create a relationship via API POST to /api/relations endpoint returns an error


akersj:
If you attempt to create a relationship via API POST to /api/relations with a valid payload the caller is returned 500 Internal Server Error with payload:

```
{
  "@context": "/api/contexts/Error",
  "@type": "hydra:Error",
  "hydra:title": "An error occurred",
  "hydra:description": "Variable \"‘api\" is not valid around position 12 for expression `is_granted(‘api:post’)`.",
  "trace": [
    {
      "namespace": "",
      "short_class": "",
      "class": "",
      "type": "",
      "function": "",
      "file": "/Users/jamie/Desktop/Hotarucon/vendor/symfony/expression-language/Parser.php",
      "line": 230,
      "args": []
    },
```
(rest of stack trace removed as it adds little value)

Details
-------

| Question                                              | Answer
|------------------------------------------------------ | ------
| Relevant Bolt Version                                 | 5.1.0 |
| Install type                                          |  Composer install |
| <abbr title="Backwards Compatibility">BC</abbr> Break | n/a |
| PHP version                                           | 7.4 |
| Web server                                            | Apache |
| For UX/UI issues                                      | n/a |

Reproduction
------------

Send a valid POST request to /api/relations using Postman or Bolt's built-in API tools.  E.g.
```
{
  "fromContent": "/api/contents/123",
  "toContent": "/api/contents/116",
  "position": 0
}
```

### Expected result

The caller is returned 200 OK with a payload describing the new relationship.

### Actual result

The caller is returned 500 Internal Server Error with a payload including a stack trace.

Potential cause
------------

(My apologies I am a software tester by trade not a developer!)
In /src/Entity/Relation.php on line 18 the wrong quotation marks are used around _api:post_.

Bug: 
`*          "post"={"security"="is_granted(‘api:post’)"}`

Correct: 
`*          "post"={"security"="is_granted('api:post')"}`

PR #3060, opened 2022-01-24
Update Relation.php

Joossensei:
Fixes issue #3059 
Wrong quotation marks were used which caused an error.


bobdenotter (approved):
Cool! 👍 

The (unrelated) failure in the CI-tests is fixed in #3061

PR #3066, opened 2022-01-27
Add the possibility to sort and filter users in the backend

damien-jourdan:
By default, this feature is disabled, but can be activated by adding user_show_sort&filter: true in bolt/config.yaml


bobdenotter (approved):
Thanks, @damien-jourdan 👍 

I'm not convinced yet about the default value of the setting.. Let's see how it goed, but we might toggle it, so it's _enabled_ by default.

PR #3068, opened 2022-01-27
Fix configuration values not resolving correctly

ionutlip:
https://github.com/bolt/core/issues/3067


bobdenotter (approved):
thanks again, @ionutlip 👍

PR #3073, opened 2022-02-01
Avoid running node service as root user

fsidobre:
Hello!

The current node image used for compiling assets is run using it's default user, which is `root`. However, the image is providing an existing non-root user which is `node`. This one should be used, as mentioned in the [docker node best practices](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user).

## Current problem 

When starting the project, assets under the `/public/assets` folder are generated using the `node` image as `root` user, so the files and subfolders are created with `root:root` ownership.

This causes a following Webpack subtask to fail : 

webpack-cli-error

## Solution

As Bolt is not running a custom image for this service, but is using docker-compose, the easiest tweak would be to just switch the user for the `node` service to the `node` user. Generated file would then inherit the ownership of parents folders on host.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,6 +47,7 @@ services:
     volumes:
       - ./:/home/node/app
     working_dir: /home/node/app
+    user: node
     command:
       - ./node_modules/.bin/encore
       - dev


bobdenotter (approved):
Thanks, @fsidobre 

I didn't know about this! 👍 

I've restarted Cypress (which can be fickle), and if it passes, I'll merge it in.

fsidobre:
Hello @bobdenotter ! 

From what I see, it seems there are errors in the cypress tests, making them fail until this happens : 
`You've exceeded the limit of test results under your free billing plan this month.`

Unless it's just a bunch of false positives ?

PR #3074, opened 2022-02-02
Update AuthSubscriber.php to prevent logout error

matesko:
Prevent error on logout when user is not logged in


bobdenotter (approved):
Thanks, @matesko 👍 

I've restarted Cypress (which can be fickle), and if it passes, I'll merge it in.

PR #3081, opened 2022-02-05
Make it possible for GlobalVoter.php to vote on non-bolt users.

simongroenewolt:
This is a suggested 'fix' for issue https://github.com/bolt/core/issues/3079 

This needs to be discussed a bit before merging it in I'd say, especially if this could be the right fix for the wrong problem. 

Should global permissions used in the admin be mixed with permissions used for the front end? If the answer is 'yes' than I'd say this change should probably go in, if the answer is 'no' there should be another solution I'd say.

The global permissions I'm referring to are of course the API ones: 

```
    api:get: [ ROLE_WEBSERVICE ] # allow read access to Bolt's RESTful and GraphQL API
    api:post: [ ROLE_WEBSERVICE ] # allow write access to Bolt's RESTful and GraphQL API
    api:delete: [ ROLE_WEBSERVICE ] # allow delete access to Bolt's RESTful and GraphQL API
```

Reading the docs here https://github.com/bolt/core/blob/02b828cbe08cb7b3ee9c273ee8aa27cd1b91749d/config/bolt/permissions.yaml#L14 state explicitly the permissions are meant for the admin back-end. That documentation is not in line with the usage for the api access at the moment.

### DIFF
--- a/config/bolt/permissions.yaml
+++ b/config/bolt/permissions.yaml
@@ -10,8 +10,10 @@
 assignable_roles: [ROLE_DEVELOPER, ROLE_ADMIN, ROLE_CHIEF_EDITOR, ROLE_EDITOR, ROLE_USER, ROLE_WEBSERVICE]
 
 # These permissions are the 'global' permissions; these are not tied
-# to any content types, but rather apply to global, non-content activity in
-# Bolt's backend. Most of these permissions map directly to backend routes;
+# to any content types. Most of them apply to global, non-content activity 
+# in Bolt's backend but there are exceptions like the api:* permissions
+# used to manage access to the api-platform based requests.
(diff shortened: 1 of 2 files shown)


bobdenotter:
> If the answer is 'yes' than I'd say this change should probably go in, 

Actually, I'd say "yes". :-)

It makes sense for some people to _not_ want public access to the (read-)API,  but it also makes sense to allow it.

bobdenotter (approved):

simongroenewolt:
(Updated doc accordingly)

simongroenewolt:
@bobdenotter I think this one is ready to be merged now, but even though I don't understand how the errors from the Cypress Tests _could_ be permission related.

PR #3086, opened 2022-02-09
Adds the ability to change dynamic the bolt configuration directory

ionutlip:


src/Configuration/Parser/BaseParser.php
```diff
@@ -73,6 +73,17 @@ public function getInitialFilename()
         return $this->initialFilename;
     }
 
+    protected function getProjectConfigDir(): string
+    {
+        $projectConfigDir = $_ENV['BOLT_CONFIG_FOLDER'] ?? 'bolt';
+
+        if (empty($value) && getenv('BOLT_CONFIG_FOLDER')) {
```
bobdenotter, line 80:
```suggestion
        if (empty($projectConfigDir) && getenv('BOLT_CONFIG_FOLDER')) {
```

simongroenewolt:
While this probably works, I don't think using `$_ENV` is the 'Symfony way' of solving this, is it? I'd say the way more in line with the Symfony configuration expectations would be to update the paths that get passed to Config.php ($projectDir is defined in services.yaml), and have the Config class pass a full path to the different Parser implementations instead of handing them a 'basePath' and requiring the parser to append the path to the actual config location?

bobdenotter:
@simongroenewolt That's a valid point, but then where/how would we add the logic to make it fall back to a sensible default, if it's omitted from `services.yaml`? 🤔

bobdenotter:
@simongroenewolt I've thought about it some more, and we _might_ want to expand this to be able to configure the location of more config files at some point.

In that case, i think it makes sense to do it in the `ENV` (or `.env`), because if we put it in configuration itself somewhere, it would first need to read configuration and _then_ determine the location. 

Let's roll with this as-is! It's not breaking in any BC way, nor it is obtrusive for those who don't need it. 🚀

simongroenewolt:
@bobdenotter I agree it should work from the .env and that it should not be done in the `config/bolt/...` files, I'd only say that to make it work like symfony generally does this config should be passed by the dependency injection mechanism + services.yaml (or related) to the Config.php class, instead of using $_ENV in a class that is instantiated by Config.php. This is how `$publicFolder` is passed to Config.php as well at the moment, so why should this config option be different? 

I'm guessing your main concern is backwards compatibility?

simongroenewolt:
I did a quick experiment. You could add an additional constructor argument (string) for this to Config.php, with a default value. Now you don't _have_ to supply it, but the moment you add this parameter to services._default.bind (probably using an env var) you can set the value from either services.yaml or .env / environment variables.

bobdenotter:
| I'm guessing your main concern is backwards compatibility?

Correct

| I did a quick experiment. 

Interesting! I'm guessing you did set it in Kernel.php, then? We do other `ENV` / `.env` juggling there too.

simongroenewolt:
@bobdenotter hmm I got a bit of a surprise while trying to code my suggestion so I could show what I mean. And I now also understand you mentioning `Kernel.php` -- I didn't realize the ContentTypesParser is initialized twice. One time in Kernel.php and one time via Config.php. My suggestion works perfectly for Config.php, but not for the use in Kernel.php -- I'll see if I can come up with a suggestion that doesn't require a full rewrite of the bolt core ;-)

simongroenewolt:
Ok, the only way I see my suggestion working is if the parsers are not used from the kernel. I'd say that would be a good thing, but doing so means probably means writing a [custom route loader](https://symfony.com/doc/5.4/routing/custom_route_loader.html#loading-routes-with-a-custom-service) to replace the use of `bolt.requirement.contenttypes` params etc in `@Route` annotations. And even if @bobdenotter would agree with that solution, I don't think I've got the time at the moment to ... ah, whatever, I gave it a go just for fun, and this is the resulting example commit: https://github.com/simongroenewolt/core/commit/5760eec185bc4a15b90b71e3014b658f2aed2a06 (demo, not fully finished, but testable)

Sorry @ionutlip for hijacking your PR for my ramblings here, as this might be a bit 'out of scope'. But then again, I'd say that removing the use of the parsers from Kernel.php is a good thing and improves the overall architecture.

ionutlip:
> Ok, the only way I see my suggestion working is if the parsers are not used from the kernel. Doing so probably means writing a [custom route loader](https://symfony.com/doc/5.4/routing/custom_route_loader.html#loading-routes-with-a-custom-service) to replace the use of `bolt.requirement.contenttypes` params etc in `@Route` annotations. And even if @bobdenotter would agree with that solution, I don't think I've got the time at the moment to ... ah, whatever, I gave it a go just for fun, and this is the resulting example commit: [simongroenewolt@5760eec](https://github.com/simongroenewolt/core/commit/5760eec185bc4a15b90b71e3014b658f2aed2a06) (demo, not fully finished, but testable)
> 
> Sorry @ionutlip for hijacking your PR for my ramblings here, as this might be a bit 'out of scope'. But then again, I'd say that removing the use of the parsers from Kernel.php is a good thing and improves the overall architecture.

ionutlip:
Hi @simongroenewolt no problem, I agree with the suggestions given but at the moment these changes would involve more work and more tests to make sure that we do not spoil something of the current functionality. I analyzed and looked for the best possible solution to do this change to be easy as possible. but maybe we can add to Todo to find a much more optimal and global solution.

bobdenotter:
@simongroenewolt Well, that looks pretty neat, and a definite improvement to get that piece of black magic out of `Kernel.php`. However, this seems like a change we should _not_ do in a `0.0.1`-release. 

What should we do with this? Merge it, and mark it as "experimental", so I can put out 5.1.3, and then we go ahead with refining @simongroenewolt's WIP?

ionutlip:
Thanks, @simongroenewolt  very good these suggestions, I will prepare a fix immediately

bobdenotter:
Ok, let's roll with this! We'll improve it for Bolt 5.2.0 (and we can add some "yaml migrations" to  update config automatically. 

Going to merge this, and rebase #3088

PR #3087, opened 2022-02-09
Add a username to the backendmenu cache key

Joossensei:
When an admin logs out and right after that a editor logs in the editor sees the cached backendmenu of the admin. This will prevent that.


bobdenotter (approved):
Whoop! 🎉

PR #3093, opened 2022-02-17
Fix custom homepage setting: allow singular ContentType slugs

### LINKED ISSUE #3092: Changing homepage entry will cause "ContentType is not defined" error
Changing the `homepage` option in the `config.yaml` to e.g. `page/1` will throw the following error:
```
Homepage is set to `page/1`, but that ContentType is not defined
```

The contenttype `page` exists and an page entry with id 1 also exists.
The same error wil be thrown if you try to use the slug instead e.g. `homepage: page/start`.

Details
-------

| <abbr title="Backwards Compatibility">BC</abbr> Break | unknown
| PHP version                                           | 8.1 |
| Web server                                            | Built-in
| For UX/UI issues                                      | Firefox 97.0

bobdenotter:
Fixes #3092

### DIFF
--- a/src/Controller/Frontend/HomepageController.php
+++ b/src/Controller/Frontend/HomepageController.php
@@ -29,7 +29,7 @@ class HomepageController extends TwigAwareController implements FrontendZoneInte
         }
 
         $homepageTokens = explode('/', $homepage);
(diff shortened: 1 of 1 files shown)


Joossensei (approved):
Looks good!

UtechtDustin:
Thank you for the fast fix!

PR #3095, opened 2022-02-19
update 'dynamic' routing based on names of content types and taxonomies

simongroenewolt:
This change moves content-type and taxonomy routing config from Kernel + annotations to a RouteLoader service class. By doing so it removes (some of) the dependencies of Kernel.php on config parsers, the use of these parsers by Kernel.php can be problematic/confusing for developers working on or using Bolt because it removes the possibility to override (parts of) the Config.php class, for example during testing.
As the Config.php class is itself a service, so it could not be used by the kernel (you cannot use services while configuring the service container)

### DIFF
--- a/config/routes.yaml
+++ b/config/routes.yaml
@@ -16,6 +16,14 @@ frontend:
   prefix: /
   type: annotation
 
+#
+# Special loader for dynamic routes based on names of content types and taxonomies.
+# Both for front-end and for the admin.
+#
+dynamic_routes:
+    resource: 'bolt.dynamic_route_loader::loadRoutes'
+    type: service
+
 # ------------------------------------------------------------------------------
 # Place your own routes here, that have a LOWER priority than the default routes.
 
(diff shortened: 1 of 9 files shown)


simongroenewolt:
Hmm, the phpunit errors do not look like they are related to my changes, or are they?

simongroenewolt:
I've been thinking a bit more about the general Bolt configuration setup, and while I do think this PR could make sense, I'm wondering if it is the right approach after all. This is because of on the way Config.php is used, being injected as a service, vs. the 'general' way configuration is handled by symfony and related packages/extensions.
If I understand correctly most bundles handle config parsing in the container building phase (via an Extension class), which is more or less the use in the same phase as I've _removed_ in this PR. 
My guess is that when you decided on the bolt config setup there probably were good reasons to make it like it is now - it would be interesting to know more about the reasons why it is like it is.

bobdenotter:
> My guess is that when you decided on the bolt config setup there probably were good reasons to make it like it is now - it would be interesting to know more about the reasons why it is like it is.

It is the way it is now, because I didn't know how else I could (programmatically) add the requirements to the routing. Personally, I think this is an improvement, because it removes black magic from Kernel. It might look like it's _more_ code than before, but if you take into account a large part of the new code is comments, it seems to be about the same. 

I'm 👍 on this

PR #3096, opened 2022-02-20
Update Date.vue dateFormat to include seconds

simongroenewolt:
Added seconds to dateFormat to fix issue with date-compare between database original and edited version, causing users that didn't have the right to change publication dates to be denied making edits.

This fixes https://github.com/bolt/core/issues/3082

Note, this is only for the `dateFormat`, this is used 'behind the scenes' in this vue component and is used to post to the server, the actual format shown is specified in `altFormat`.

### DIFF
--- a/assets/js/app/editor/Components/Date.vue
+++ b/assets/js/app/editor/Components/Date.vue
@@ -105,7 +105,7 @@ export default {
                 wrap: true,
                 altFormat: 'F j, Y',
                 altInput: true,
-                dateFormat: 'Y-m-d H:i',
+                dateFormat: 'Y-m-d H:i:S',
                 enableTime: false,
             },
         };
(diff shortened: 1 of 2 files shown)


Joossensei:
Hey! First of all nice fix! But I don't think it happens on the background only as the Cypress tests are failing. 

One way to fix this is by editing the failing tests over here: 
[https://github.com/bolt/core/blob/8cbe02383a3b31b71cbf3e7b3fe35ba29ceacde4/tests/cypress/integration/edit_record_1_field.spec.js#L54-L181](url)

It should be fixed by just adding the seconds it should expect.
Screenshot 2022-02-22 at 08 50 27


If the tests work then we can merge it in 😄

simongroenewolt:
Thanks @Joossensei for the tip on where to fix the tests. It looks like these tests do check a value in the dom that is not visible to the user, but as it is the actual value that _should_ be posted it is important to fix them anyway -- I'll try to get that done this week.

Joossensei (approved):
Cool! 👍🏻

PR #3099, opened 2022-02-25
Update cypress tests so it retries on failure

Joossensei:
For more info see:
[https://docs.cypress.io/guides/guides/test-retries#Global-Configuration](https://docs.cypress.io/guides/guides/test-retries#Global-Configuration)


bobdenotter (approved):
Cool! 👍

PR #3101, opened 2022-02-25
change initialization of fields on postLoad doctrine event

simongroenewolt:
This is a PR to illustrate a possible solution for https://github.com/bolt/core/issues/3085 -- to be a real fix the logic probably needs to be moved.


bobdenotter:
I spent some time this morging doing a bunch of monkey-tests, and for me it works like a charm ✨ 

Would be great if we could (like you say) use this to get rid of one of the magic `…FillListeners`.. I mean, we'll still have a piece of Black Magic Voodoo, but at least, it'll be __*one less*__ piece of Black Magic Voodoo!

Filoz:
Hi
we had the same issue as https://github.com/bolt/core/issues/3085 and this pull fixed the problem

simongroenewolt:
PR is no longer 'Draft' so I can check the tests.

simongroenewolt:
Hmm, tests are failing on a missing `public/assets/manifest.json`, which is a file that is probably generated when running encore (webpack) when installing bolt. The _best_ way around this would be to introduce a service and move almost all code from ContentEditController.php to this service. That way the test wouldn't have to do a 'real' request at all. But that is quite a change.

bobdenotter:
> which is a file that is probably generated when running encore (webpack) when installing bolt.

the cypress tests do this by running 

```
npm set progress=false
npm ci
```

But, it doesn't really make sense to do this for the phpunit tests, right? 

> The best way around this would be to introduce a service and move almost all code from ContentEditController.php to this service. That way the test wouldn't have to do a 'real' request at all. But that is quite a change.

Yes, that would probably be best. And i also agree that that'd be quite a bit of work to refactor that. It should be out of scope for this specific PR. 

The testing code has keys like `622e624a526f0`.. Is it wise to rely on these? Won't they potentially change over time? 

Final thought for now.. How about calling `contentEditController->save()`, without doing the actual request? 

If that doesn't work, i see two potential ways to move forward: 

 - Forego this as a unit test for now, until we can refactor contentEditController to move the logic to a class we can Unit test?
 - Convert this test to a Cypress test after all

simongroenewolt:
> > which is a file that is probably generated when running encore (webpack) when installing bolt.
> 
> the cypress tests do this by running
> 
> ```
> npm set progress=false
> npm ci
> ```
> 
> But, it doesn't really make sense to do this for the phpunit tests, right?

Yep, I think it would be really bad to start depending on that for the 'basic' phpunit tests. Although if the tests would use symfony/panther they could run browser requests with js enabled, and then it would start to make sense. (But these kinds of tests are now handled by cypress)

> 
> > The best way around this would be to introduce a service and move almost all code from ContentEditController.php to this service. That way the test wouldn't have to do a 'real' request at all. But that is quite a change.
> 
> Yes, that would probably be best. And i also agree that that'd be quite a bit of work to refactor that. It should be out of scope for this specific PR.
> 

I don't know if it would be _that_ much work, but I agree it is out of scope for this PR.

> The testing code has keys like `622e624a526f0`.. Is it wise to rely on these? Won't they potentially change over time?
> 

That could be the case, but I don't think so. From what I understand these keys are being 'made up' in the vue.js code when editing collections.

> Final thought for now.. How about calling `contentEditController->save()`, without doing the actual request?
> 

That might be the best course of action without introducing a service, but it could be hard to mock the request and other dependencies.

I don't like removing the test, as I think this is a part of Bolt that really should get more tests, not less. But maybe it will be the only solution if calling save() manually is too hard.

Cypress tests are not an option because they cannot interact and test directly on php objects and the db.

I-Valchev:
hey hey @bobdenotter @simongroenewolt 

Default values for collection fields work differently, because you can repeat them.
See: https://docs.boltcms.io/5.0/fields/collection

```
test_collection:
            type: collection
            fields:
                collection_text:
                    type: text
                    default: "default val" #this is not working
```
^ What would this mean? This only allows you to add one of the `collection_text` items with a default values. But there's no way to add multiple, which is the purpose of the collection. Also when there are multiple field types inside the collection, no way to define which comes after which, etc.

All I'm saying is that, when this was implemented some time ago, it was implemented in the following format with a purpose:

```yaml
 features:
            type: collection
            fields:
                title:
                    type: text
                content:
                    type: html
                image:
                    type: image
            default:
                0:
                    field: title
                    default: "Lets start with a title"
                1:
                    field: image
                    default:
                        filename: "kitten.jpg"
                        alt: "Picture of a kitten"
                2:
                    field: content
                    default: "<strong>Finish strong</strong> after the kitten with some content"
```

I-Valchev:
Maybe that's more appropriate as a reply to #3129 3113

bobdenotter:
This will also fix #3162, probably

PR #3104, opened 2022-02-28
Add helper to render a themed twig template.

vlad-ghita:
This make it easy to render templates located inside themes using TwigAwareController.


vlad-ghita:
Closing b/c I'll open a new PR with all changes.

PR #3108, opened 2022-03-03
Add map to htaccess

peterboorsma:
Web inspectors return a 403 error because css source maps were not allowed access
Add map filetype solves this.


bobdenotter:
Thanks, @peterboorsma 👍

PR #3119, opened 2022-03-14
Bugfix: verify setting correct tags for caching works correctly

### LINKED ISSUE #3114: Cache throws exception when using multiple contenttypes in a `type:select`
When having 

```yaml
caching:
    …
    selectoptions: 3600
    … 
```

Throws an exception, if you're using a select like 

```yaml
somevalue:
    type: select
    values: (events,news,pages)/{title} - {contenttype} № {id} ({status})
```

bobdenotter:
Fixes #3114


nestordedios (approved):
Looks good! 👍

PR #3122, opened 2022-03-15
Update float field so it uses step='any' by default

Joossensei:
This way the amount of decimals after the comma isn't fixed by default


bobdenotter (approved):
👍

PR #3132, opened 2022-03-19
Use requestStack to access current request when needed.

vlad-ghita:
Storing current request in constructor isn't a good idea.
In general, it works (eg: when request from browser or most tests).

But not always.
EG: when test extends WebTestCase and login is executed through `$this->loginUser()`, then container is built without an active request.
Thus, request is null in constructors ...

Proper fix implies storing `$requestStack` and accessing current request when required.

This PR fixes Canonical and FrontendMenu (these failed in my use-case, so far).

### DIFF
--- a/src/Canonical.php
+++ b/src/Canonical.php
@@ -24,7 +24,7 @@ class Canonical
     private $urlGenerator;
 
     /** @var Request */
-    private $request;
+    private $request = null;
 
     /** @var string */
     private $scheme = null;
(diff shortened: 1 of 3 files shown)


bobdenotter:
Hi @vlad-ghita, 

At first I though there might be danger of infinite recursion here? 


```
    public function getRequest(): Request
    {
        // always try to use current request
        if ($this->request === null) {
            $this->request = $this->requestStack->getCurrentRequest() ?? Request::createFromGlobals();
            $this->init();
        }

        return $this->request;
    }

    public function init(): void
    {
        // Ensure in request cycle (even for override).
        if ($this->getRequest() === null …) {
            return;
        }
```

But, that's not true, because `init` will only be called from `getRequest`. After which the Request is set.

PHPstan is complaining about it, though.. It looks like we can simplify it like this: 

```
    public function init(): void
    {
        // Ensure in request cycle (even for override).
        if ($this->getRequest()->getHost() === '') {
            return;
        }
```

Would that work for you?

src/Canonical.php
```diff
@@ -44,27 +44,39 @@ class Canonical
...
 
-        $this->init();
+        return $this->request;
     }
 
     public function init(): void
     {
         // Ensure in request cycle (even for override).
-        if ($this->request === null || $this->request->getHost() === '') {
+        if ($this->getRequest() === null || $this->getRequest()->getHost() === '') {
```
bobdenotter, line 73: Fix PHPStan's remark: 

> Strict comparison using === between Symfony\Component\HttpFoundation\Request and null         will always evaluate to false.

```suggestion
        if ($this->getRequest()->getHost() === '') {
```

vlad-ghita: Agreed with possible infinite loop.
I aimed to keep `init()` for backwards compatibility.
But maybe we should simply ditch it.

vlad-ghita: See latest changes.
I added a request setter.
Please re-run checks.

My tests passed.

bobdenotter (approved):
Looks good now, thanks @vlad-ghita 👍 

I'll merge in after tests are done!

PR #3135, opened 2022-03-22
Use JSON_VALUE on MySQL platforms

### LINKED ISSUE #2088: Setcontent with a LIKE at the beginning of the string doesn't match
I'm using the equivalent of a setcontent tag in some controller code to match all content records that start with a given letter.

However this doesn't work - it always returns 0 records.

I've narrowed this down in the following code and query:

```php
$contentType = 'lyrics-artists';
$filterParameters['title'] = 'A%';
$matchingContent = $this->boltQuery->getContent($contentType, $filterParameters);
```
```sql
SELECT content FROM Bolt\Entity\Content content LEFT JOIN content.fields fields_title LEFT JOIN fields_title.translations translations_title WHERE content.contentType = :ct0 AND (JSON_EXTRACT(translations_title.value, '$[0]') LIKE :title_1 AND fields_title.parent IS NULL AND fields_title.name = :field_title)

    ct0: lyrics-artists
    title_1: a%
    field_title: title
```

A content with the title field set to "A Great Big World" does not match in the above query.

If I run JSON_EXTRACT manually in MariaDB, it includes the quotes in the JSON, so the above string begins with `"A` not `A` as the LIKE statement is expecting.

```
SELECT value, JSON_EXTRACT(value, '$[0]') FROM `bolt_field_translation` WHERE translatable_id = 76 
```

| value | JSON_EXTRACT|
|-|-|
| ["A Great Big World"] | "A Great Big World" |

Including the " in the code does indeed return the record I was expecting:

```php
$contentType = 'lyrics-artists';
$filterParameters['title'] = '"A%';
$matchingContent = $this->boltQuery->getContent($contentType, $filterParameters);
```

I believe the solution is to use `JSON_VALUE` instead of `JSON_EXTRACT` - and I'm happy to send a PR, I'm just unsure if this is the right solution (will this break anything elsewhere in Bolt/MySQL?) and also the impact on PostgreSQL - @I-Valchev ?

```
SELECT value, JSON_EXTRACT(value, '$[0]'), JSON_VALUE(value, '$[0]') FROM `bolt_field_translation` WHERE translatable_id = 76 
```

| value | JSON_EXTRACT | JSON_VALUE |
|-|-|-|
| ["A Great Big World"] | "A Great Big World" | A Great Big World |

andysh-uk:
This is my attempt to fix #2088.

On MySQL/MariaDB platforms, a JSON string such as `"test"` (like Bolt stores for field translations) includes the JSON syntax, so a "LIKE" search fails when looking for the first character (e.g. `"test" LIKE 't%'` never matches because the first character is the quote, not the "t").

```
MariaDB [(none)]> select JSON_EXTRACT('"test"', '$[0]');
+--------------------------------+
| JSON_EXTRACT('"test"', '$[0]') |
+--------------------------------+
| "test"                         |
+--------------------------------+
1 row in set (0.001 sec)
```

However `JSON_VALUE` works as expected:

```
MariaDB [(none)]> select JSON_VALUE('"test"', '$[0]');
+------------------------------+
| JSON_VALUE('"test"', '$[0]') |
+------------------------------+
| test                         |
+------------------------------+
1 row in set (0.000 sec)
```

Two bits I need help with on this PR:

* there is a change to `config/packages/doctrine.yaml` which I don't know what I need to do (if anything) to automatically apply this on an update
* I only have access to fairly recent MariaDB, so I have not been able to test this on earlier MySQL servers.  Looking at [the docs](), this function was added to MySQL in 8.0.21 and MariaDB 10.2.3.  

Do you know of a way to further limit my restriction to only use JSON_VALUE on MySQL >= 8.0.21 and MariaDB >= 10.2.3, and fall back to JSON_EXTRACT on earlier servers?

What is Bolt's minimum supported version of MySQL and MariaDB?


andysh-uk:
I’ve been thinking about this some more and have a better implementation I think.

JSON_UNQUOTE exists on both MariaDB and MySQL at the same time (from what I can find) as JSON_EXTRACT (which Bolt is already using.)

Therefore a MySQL/MariaDB version that can’t use JSON_UNQUOTE in all likelihood can’t use JSON_EXTRACT either anyway.

Therefore if we wrap JSON_EXTRACT in JSON_UNQUOTE, we should get either an unquoted string, or a JSON array.

Side point: the use of JSON functions in Bolt 4/5 already limits Bolt to a minimum of MySQL 5.7.8 and MariaDB 10.2.3 (I believe.) Is it worth including this in the system requirements doc page, like the minimum PHP version?

Happy to submit a PR for this if the consensus is yes.

andysh-uk:
Also I’m a bit confused with the coding standards feedback.

Why is an “if / elseif / else” construct suggested to be “if / else / if / else” instead?

src/Doctrine/JsonHelper.php
```diff
@@ -31,8 +31,11 @@ public static function wrapJsonFunction(?string $where, $slug, Connection $conne
             if ($version->getPlatform()['driver_name'] === 'pgsql') {
                 // PostgreSQL
                 $resultWhere = 'JSON_GET_TEXT(' . $where . ', 0)';
+            } else if ($version->getPlatform()['driver_name'] === 'mysql') {
```
bobdenotter, line 34:
```suggestion
            } elseif ($version->getPlatform()['driver_name'] === 'mysql') {
```

bobdenotter:
Hey @andysh-uk 👋 

| Why is an “if / elseif / else” construct suggested to be “if / else / if / else” instead?

If you use `else if`, it's seen as two distinct cases, but if you use `elseif`, it'll be seen as a proper "else if".. I've taken the liberty of ninja-editing that in. 

In the other thread i've suggested a patch for the YAML migration to be added for this.

bobdenotter:
| Therefore if we wrap JSON_EXTRACT in JSON_UNQUOTE, we should get either an unquoted string, or a JSON array.

That's a fair point.. Would you suggest doing that as a _replacement_ of this PR, or as a followup? 

| Side point: the use of JSON functions in Bolt 4/5 already limits Bolt to a minimum of MySQL 5.7.8 and MariaDB 10.2.3 (I believe.) Is it worth including this in the system requirements doc page, like the minimum PHP version?

Yes, let's! (and we should add the minimum version of SQLite too, which is a bit more fuzzy.. IIRC: "SQLite 3.17 (with JSON1) or 3.38 and up" 

| Happy to submit a PR for this if the consensus is yes.

As always, would be much appreciated! :-)

andysh-uk:
> If you use else if, it's seen as two distinct cases, but if you use elseif, it'll be seen as a proper "else if".. I've taken the liberty of ninja-editing that in.

Perfect, thank you!  I can't honestly say I've ever used "elseif", I always write it as you say it!

> In the other thread i've suggested a patch for the YAML migration to be added for this.

I saw this, that looks great and really easy to add in changes, brilliant!

> That's a fair point.. Would you suggest doing that as a replacement of this PR, or as a followup?

Leave this with me for another day or two, ideally I'd like to replace this PR as the changes in this effectively *increase* the MySQL version requirement to v8, whereas if my thoughts on JSON_UNQUOTE work the same way, it'll keep the version requirement the same as it is now.

> As always, would be much appreciated! :-)

No worries, just created a PR on Github for this.

bobdenotter:
> Leave this with me for another day or two, ideally I'd like to replace this PR as the changes in this effectively increase the MySQL version requirement to v8, …

That's a good point. Yeah, we shouldn't bump the minimum required version, without bumping the major of Bolt itself. In other words: with Bolt 6, at the earliest. :-)

I'll keep an eye on this PR, for when it's good-to-go!

bobdenotter:
Hey @andysh-uk. Not to rush you or anything, but I was curious if you had an update on this one? :-)

andysh-uk:
Hey @bobdenotter 👋🏽 

I did have some trouble when I looked at this initially, it was complaining about an issue with the lexing/parsing (expecting a comma but getting a closing bracket.)

I've come back to it with fresh eyes today, and I think I've cracked it.

I'll get my branch updated with my fix (it's currently only in my personal dev site) and update the PR tomorrow.

bobdenotter (approved):
Thanks, @andysh-uk 👍

PR #3137, opened 2022-03-23
Reverse search and replace array's in str_replace

HeinDR:
The parameters in the search array are appended with an index number. This causes problems when the elements in the array exceed 10 because the first element will find a part in the 10th element (e.g. `content.topics = :topics_1` and `content.topics = :topics_10`). The resulting query will be corrupted because the 0  (in case of the 10th element) will not be replaced and left dangling in the string.
This change fixes this by reversing the array's, so the largest string will be looked up and replaced first.

### DIFF
--- a/src/Storage/SelectQuery.php
+++ b/src/Storage/SelectQuery.php
@@ -604,7 +604,7 @@ class SelectQuery implements QueryInterface
                 return sprintf("JSON_SEARCH(%s, 'one', %s) != ''", $valueAlias, $parameter[0]);
             }, $expressions);
 
-            return str_replace($expressions, $newExpressions, $filter->getExpression());
+            return str_replace(array_reverse($expressions), array_reverse($newExpressions), $filter->getExpression());
         }
 
         $originalLeftExpression = 'content.' . $filter->getKey();


bobdenotter:
Hi @HeinDR , 👋 

Looks like you bumped into a legitimate usecase, but I'm not a huge fan of this particular fix. It feels "off". 

Would it be a solution to use something like this, on line 583 of `SelectQuery.php`, instead? That way, they'd get params like `topics_005`.

```php
$nameParam = sprintf('field_%03d', $filter->getKey());
```

HeinDR:
Hi @bobdenotter,

It could be a solution, but then the issue will occur again with more than 10000 fields. I agree it's a highly unlikely scenario, but still. Line 583  would also not be the place to fix this because `getRegularFieldWhereExpression` is called on line 572 and the number index on the field is set by the `QueryParameterParser`  if I'm not mistaken. 
I agree that my solution feels a bit hackish, but as long as it is guaranteed that the index numbers are sequential, and I think it is, it is a solid solution I think. 
Fixing this in the  `QueryParameterParser`  would probably have more impact.

bobdenotter (approved):
Fair enough! In that case, let's get this merged! 🚂 

Thanks, @HeinDR!

PR #3139, opened 2022-03-23
handle 'double wrapped' database connections

simongroenewolt:
this fix is needed when runing tests using dama/doctrine-test-bundle db-rollback support.

This should fix https://github.com/bolt/core/issues/3138


bobdenotter (approved):
Neat, thanks! 👍

PR #3141, opened 2022-03-24
Remove unused method

### LINKED ISSUE #3140: findParents() is named incorrectly, _and_ the function is not in use
The function [findParents()](https://github.com/bolt/core/blob/39f9eccbdb8da207215326f4cd602180dae37b26/src/Repository/FieldRepository.php#L126) in FieldRepository.php is incorrectly named (If it finds anything, what it will return are children). It is not used anywhere in the project. I'd say it would be a good idea to delete it to prevent future confusion.

bobdenotter:
Fixes #3140, thanks @simongroenewolt

### DIFF
--- a/src/Repository/FieldRepository.php
+++ b/src/Repository/FieldRepository.php
@@ -123,18 +123,6 @@ class FieldRepository extends ServiceEntityRepository
         return $field;
     }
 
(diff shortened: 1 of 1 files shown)


Joossensei (approved):
Looks good 👍 !

PR #3142, opened 2022-03-25
Multiple EntityManagers prefixes

matesko:
Update TablePrefix.php

Update TablePrefixTrait.php

Update Vresion.php

### DIFF
--- a/src/Doctrine/Version.php
+++ b/src/Doctrine/Version.php
@@ -20,7 +20,7 @@ class Version
     /** @var string */
     private $tablePrefix;
 
-    public function __construct(Connection $connection, string|array $tablePrefix = 'bolt')
+    public function __construct(Connection $connection, $tablePrefix = 'bolt')
     {
         $this->connection = $connection;
         $tablePrefix = is_array($tablePrefix) ? $tablePrefix['default'] : $tablePrefix;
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
Thanks, @matesko 👍

bobdenotter:
I wonder why the tests aren't running on this PR, though.. 🤔

src/Doctrine/Version.php
```diff
@@ -23,6 +23,7 @@ class Version
     public function __construct(Connection $connection, string $tablePrefix = 'bolt')
     {
         $this->connection = $connection;
+        $tablePrefix = is_array($tablePrefix) ? $tablePrefix['default'] : $tablePrefix;
```
bobdenotter, line 26: `strict_types` is true, so `tablePrefix` can never be an array, right?

bobdenotter (changes requested):

src/Doctrine/Version.php
matesko: Yes, you are corect! I updated the code

bobdenotter: `string|array` won't work. Union types are only allowed in PHP 8 and up.. Perhaps do it like this, then? 

```php
    /**
     * @param string|array $tablePrefix
     */
    public function __construct(Connection $connection, $tablePrefix = 'bolt')
```

matesko: Of course, pushed...

bobdenotter (approved):
Thanks, @matesko

PR #3145, opened 2022-03-28
Ensure correct links in menu, when changing `backend_url` in `services.yaml`

bobdenotter:


Joossensei (approved):
"There are only two hard things in Computer Science: cache invalidation and naming things." -- Phil Karlton

PR #3160, opened 2022-04-04
Bump flex to ^1.17, v1.13 is going to stop working very soon

nicolas-grekas:
We're going to shutdown the flex-server very soon and this locked version will stop working.
It could be great to unlock this constraint on the 4.2 branch, which is the only one affected.
Unless you think this version is too old and it's considered out of maintenance?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "symfony/dependency-injection": "^5.1",
         "symfony/dotenv": "^5.1",
         "symfony/expression-language": "^5.1",
-        "symfony/flex": "1.13.*",
+        "symfony/flex": "^1.17",
         "symfony/form": "^5.1",
         "symfony/framework-bundle": "^5.1",
         "symfony/http-client": "^5.1",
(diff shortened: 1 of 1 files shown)


bobdenotter:
Hi, @nicolas-grekas.. Thanks for this PR. Bolt 4 is not maintained anymore, though. 

It's up-to-date in `master`: https://github.com/bolt/core/blob/master/composer.json#L53

Most users will intall Bolt using "composer create-project", though. This is also up to date: https://github.com/bolt/project/blob/master/composer.json

bobdenotter:
I've read the linked issue, and I now see why you're making this PR to an older version. The tests are failing because of issues that have been fixed later. I'll check to see if the patch works as expected, and if so, i' going to merge it in to not block people updating their old installs (and who don't want to upgrade to 5.x)

bobdenotter (approved):
Thanks, @nicolas-grekas 👍

nicolas-grekas:
Thank you for merging and releasing quickly!

PR #3170, opened 2022-04-08
fix(backend-menu): parent menu items use wrong icon

### LINKED ISSUE #3171: Parent menu items use wrong icon
Hello

#3167 introduces a problem : when we have multiple content types using differents settings for `show_in_menu` the icon of the first item cascades on others.
wrong
In this example, second one should have a "fa-cog" icon.

david-saisondor:
Fixes #3171


bobdenotter (approved):
Thanks again, @david-saisondor! 👍

PR #3182, opened 2022-04-17
Render .webp images

xxxLOGIATxxx:
If I'm not mistaken, in Bolt CMS we can't render images in the modern .webp format "out-of-the-box". In any case i have not found such information anywhere.


bobdenotter:
I don't think this PR is how you intended it, right? 

Anyways, I think Bolt handles `.webp` just fine, as long as the system's available GD or Imagick supports it. This is not something we have direct control over, though.

PR #3185, opened 2022-04-21
Update Number field to include the min and max

### LINKED ISSUE #3184: Field type number options max and min are not working
In the [Bolt 5 documentation](https://docs.boltcms.io/5.0/fields/number#options) the option to add a minimum and maximum accepted value should be available according to the documentation. But is not implemented in [`assets/js/app/editor/Components/Number.vue`](https://github.com/bolt/core/blob/5a5423d47004a7a078319c6e64e15580e3b891cc/assets/js/app/editor/Components/Number.vue)

Details
-------

| Relevant Bolt Version                                 | 5.1.7
| PHP version                                           | 7.2 |
| Web server                                            | Nginx

Reproduction
------------

Added the field as followed in the contenttypes.yaml:
```
number_of_reviews:
    type: number
    mode: integer
    required: true
    min: 0
    max: 6
```

### Bug summary

Currently the input field does not see a max or min and will increase the number in the input higher than 6

### Expected result

I would expect when filling in a number higher than 6 (in this case) it would not accept it and clicking out of the number input field it should show an error.

Joossensei:
Fixes #3184

Default values are now between 0 and 1000.
Probably make a doc where this is noted.

### DIFF
--- a/assets/js/app/editor/Components/Number.vue
+++ b/assets/js/app/editor/Components/Number.vue
@@ -7,6 +7,8 @@
             class="form-control"
             type="number"
             :step="step"
+            :min="min"
+            :max="max"
             :required="required"
             :readonly="readonly"
             :data-errormessage="errormessage"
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
Neat! 👍

PR #3188, opened 2022-04-27
Fix: MySQL was returning case sensitive results.

codemis:
Tested on MySQL 5.7 and 8.0.26, the bolt admin panel search widget on content was only returning case sensitive results.  When the original code was run on MariaDB, the results were case insensitive.  This fix makes MySQL return case insensitive results.  The issue is described in this StackOverflow answers:

- https://stackoverflow.com/a/59000485
- https://stackoverflow.com/a/63519138/4638563

By casting the JSON_EXTRACT to CHAR, it makes the search case insensitive. 

NOTE: Since this is database specific, I am unable to write a test for the issue.


bobdenotter (approved):
Thanks for the fix, @codemis 👍 

The PHPStan failure is unrelated (it's because PHPStan got a bump to 1.6.0 and found a new thingy)

PR #3196, opened 2022-05-06
Fix version compare for composer self-update versions.

vnagara:
To avoid error on installing:

[TypeError]                                                                                                                                                                                             
  Symfony\Component\Process\Process::__construct(): Argument #1 ($command) must be of type array, string given, called in bolt/web/vendor/bolt/core/bin/composer-script  
  /Script.php on line 57  

We need to fix compare version constraint. 
Composer self-update command updates composer to v2.3-dev+xxxx that is lower than current constraint.

### DIFF
--- a/bin/composer-script/Script.php
+++ b/bin/composer-script/Script.php
@@ -51,7 +51,7 @@ class Script
      */
     public static function runPHP(array $command): int
     {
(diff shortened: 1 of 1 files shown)


bobdenotter:
Thanks, @vnagara

vnagara:
@bobdenotter Thank you for making this project. It is a lot of efforts :), I know it.

PR #3198, opened 2022-05-06
Renamed the Icon method in the Content class

Spomsoree:
to avoid name conflicts with fields named icon.

With the current naming it's impossible to name a field `icon` and access it in twig.


bobdenotter (approved):
I missed this earlies, my apologies! 

Thanks, @Spomsoree 👍

PR #3200, opened 2022-05-09
Fix setting new fields on collection.

vlad-ghita:
Bug observed when previewing a record with collection field.
Only last item in collection was available.

`$newFields` variable must be initialized outside the loop in order to capture all collection fields.


bobdenotter (approved):
Nice catch! thanks, @vlad-ghita 👍

PR #3206, opened 2022-05-23
Cast the config maintenance value to boolean

Spomsoree:
to be able to use environment configs for this setting.

`.env`:
`APP_MAINTENANCE_MODE=false` (false/true/yes/no/0/1, whatever you prefer)

`config.yaml`:
`maintenance_mode:  '%env(APP_MAINTENANCE_MODE)%'`


bobdenotter (approved):
Thanks, @Spomsoree 👍

PR #3208, opened 2022-05-23
Improve image and files modals

nestordedios:
This PR adds some modifications to the cards show on file/image modals:

- Make modal larger to fit more images. On small screen 2 cards per row are displayed. On big screens 4 cards per row are displayed.
- If resource name is too large adds ellipsis at the end of the name.
- Make resource name use normal font weight instead of bold
- Resource is selected on click

image


bobdenotter (approved):
Looking good, @nestordedios 👍

PR #3211, opened 2022-05-27
Ensure `PublishedAt` gets set correctly when publishing a Record

bobdenotter:


bobdenotter:
(The failure with PHPStan was fixed in a later PR)

PR #3219, opened 2022-06-02
Fixed the config used for search result limits

Spomsoree:


Spomsoree:
I didn't break it, did I?

PR #3223, opened 2022-06-07
Allow "Create Content" as regular editor

bobdenotter:


bobdenotter:
Merging this one in as-is. Cypress tests will need more work in a follow-up PR.

PR #3224, opened 2022-06-09
Improved the twig filter "current"

Spomsoree:
which did return false on matching routes with parameters:

image

This returned `false`, because we compare the full strings:

```
public function isCurrent($item): bool
{
    $uri = $item['uri'] ?? '';

    return $uri === $this->requestStack->getCurrentRequest()->getPathInfo();
}
```


bobdenotter (approved):
Thanks, @Spomsoree 👍 

Merging this in as-is.. Failing tests are unrelated!

PR #3238, opened 2022-06-15
Fix avatar image sizes when image is not found

### LINKED ISSUE #3234: Missing user avatar makes backend unusable
I ran into this while copying a database from production to local, without copying the files directory where the avatar image files reside. In this site Avatars are used for the users/editors. The missing user avatar file causes a missing-image to be shown in the top right corner, on the top layer, pretty huge and obscuring half of the backend interface. 

It's clear that not copying the Avatars was the actual error but is this a scenario that we want Bolt to be prepared for? There may also be other scenarios in which the Avatar files can be deleted by unknowing users or so - and the impact can be big. 

Possibly one or 2 lines of CSS can solve this but I have not gone into this.

nestordedios:
This PR fixes #3234 .

**Before fix:**
image

**After fix:**
image

### DIFF
--- a/assets/scss/modules/admin/_toolbar.scss
+++ b/assets/scss/modules/admin/_toolbar.scss
@@ -161,6 +161,7 @@
 
     &__profile {
       padding-right: 1rem;
+
       img {
         max-width: 25px;
         height: 25px;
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Excellent! 👍

PR #3252, opened 2022-06-21
Additional fix for `type: number` field, set default to `min: 0`.

bobdenotter:
See also #3231


xiaohutai (approved):
Good one. I took over the `min` and `max` values that were placed there. Technically `1000` may be too low of a value too, right?

bobdenotter:
The issue with `1` was that it's the _default_ value of the field is `0`, so just putting in the filed would block you from posting, which was confusing. :-)

PR #3254, opened 2022-06-22
Add Fixtures for SEO field

### LINKED ISSUE #2946: Make db-reset (loading fixtures) in combination with the appolodev/bolt-seo extension results in exception on dashboard
While performing a `make db-reset` with the **appolodev/bolt-seo** extension results in the following error on the dashboard when trying to view the contenttype using said extention.

Capture_db-reset-fixtures-seo-error

It looks fixtures are not properly stored/saved in the SEO field. The extension is expecting a json type format. Bolt saves fixtures to the field as a 'normal' string type.

Capture_db_value

Uncommenting the extension seo fields in the contentypes allows one to gain access again to the contenttype in the dashboard and allows one to continue to use fixtures.

Another option is to manually input new content into Bolt (without commenting out SEO fields in the contenttypes) which circumvents this error altogether, however which tampers workflow with fixtures.

Details
-------

| Relevant Bolt Version                                 | 5.0.6
| Web server                                            | Symfony CLI

nestordedios:
Fixes #2946 

When the SEO extension is installed the Fixtures generate a non valid value for the the SEO field. This PR makes sure that the value assigned to a SEO field is valid and prevents the backend from breaking when rendering a non valid value for the SEO field.

### DIFF
--- a/src/DataFixtures/ContentFixtures.php
+++ b/src/DataFixtures/ContentFixtures.php
@@ -399,6 +399,13 @@ class ContentFixtures extends BaseFixture implements DependentFixtureInterface, 
 
 
                 break;
+            case 'seo':
+                $data = ['keywords' => '', 'shortlink' => '', 'canonical' => '', 'robots' => '', 'og' => ''];
+                $data['title'] = $this->faker->sentence(4, true);
+                $data['description'] = $this->faker->sentence(120, true);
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Ah, neat! 👍

PR #3262, opened 2022-06-29
Enable the hash functionality for tabs instead of pills

### LINKED ISSUE #3261: Hashes have disappeared from URL when clicking on tabs
When using the tabs of on a Record edit page the URL is no longer updated missing the hashes on the URL:

image

Details
-------

nestordedios:
Fixes #3261

I have replace the target `data-bs-toggle="pill"` by `data-bs-toggle="tab"` to fix that issue. It seems in our source code we are not using any HTML element with an attribute like `data-bs-toggle="pill"`. 

It should be safe to replace

### DIFF
--- a/assets/js/app/common.js
+++ b/assets/js/app/common.js
@@ -35,7 +35,7 @@ $(document).ready(function() {
 
     /*
      ** Hash on tabs functionality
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Nice catch! :-)

PR #3263, opened 2022-06-29
Replace manual copy to clipboard with tiny library

I-Valchev:
👋 Refactors the Copy filename feature for files, removing ugly hidden input field generation with a tiny library. 

I couldn't update `package-lock.json` without _too_ many unrelated changes because of this error 😔

`This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!`

### DIFF
--- a/assets/js/app/common.js
+++ b/assets/js/app/common.js
@@ -3,6 +3,7 @@ import { DateTime } from 'luxon';
 import { Popover } from 'bootstrap';
 import { Tab } from 'bootstrap';
 import { resetModalContent } from './modal';
+import ClipboardJS from "clipboard";
 
 import { version } from '../version';
 window.assetsVersion = version;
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
Looks good. Thanks, mate! :-D

(not sure why you'd get that notice about the lockfile. If yours is recent, why would it try to revert it `lockFileVersion@1`? 🤔 )

assets/js/app/common.js
```diff
@@ -3,6 +3,7 @@ import { DateTime } from 'luxon';
 import { Popover } from 'bootstrap';
 import { Tab } from 'bootstrap';
 import { resetModalContent } from './modal';
+import ClipboardJS from "clipboard";
```
bobdenotter, line 6: Ninja-edit: 

```suggestion
import ClipboardJS from 'clipboard';
```

I-Valchev:
Yeah, strange.. thanks for fixing it here

PR #3265, opened 2022-06-30
Remove collection item if the user accepts

### LINKED ISSUE #3264: Collection items get removed before accepting or denying action
When clicking on the remove button of a collection item, the item gets removed and then we get prompt with a message asking us to accept or deny the removal of the item.
image

Details
-------

Reproduction
------------
### Expected result

First ask the user the accept or deny the item
Second remove or keep the item depending on user answer

### Actual result

Item is removed before user decides what to do with the collection item

nestordedios:
Fixes #3264 

Detect if the save button is clicked and remove the collection item then.


bobdenotter (approved):
Whoop whoop!

PR #3272, opened 2022-07-05
Copy the services.yaml bindings to the services_bolt.yaml config

Spomsoree:
This will enable autowiring for the extensions.

E.g. if you need to use the defaultLocale in a extension, it will now be autowired.

I'm not 100% happy with the way I get the bindings (through the Config-Class), so if there is a better way to get those, let me know.


Spomsoree:
Looks like I need to check which bind will actually be used in the extensions.

Spomsoree:
@bobdenotter are you aware of a class or function which could help me determine what autowires are used in a namespace?

PR #3274, opened 2022-07-07
The services_bolt.yaml file now keeps the binds after regenerating it

Spomsoree:


Spomsoree:
Those failing test aren't my fault, right?

PR #3279, opened 2022-07-20
Made the checkbox template recognise a false string as a value

Spomsoree:
to be able to include it without the value being overwritten by the base template.

Better explanation:

I'm creating my own field template, which includes checkboxes.

I include the checkbox like this:
```
{% include '@bolt/_partials/fields/checkbox.html.twig' with {
    name:  'autoplay-' ~ id,
    label: 'autoplay'|trans,
    value: autoplay,
} %}
```

The value I pass is either `true` or `false`, obviously.
If I pass `false` the value is somehow overwritten, I'm guessing because they both extend the same base template and both use variables named `value`. 

This now opens the option to pass `false` as a string, which stops twig from using the base template `value`.

Edit on the second commit:

I've also added the option to not print the postfix.
If I include the checkbox template I need this option to prevent multiple identical postfixes.

### DIFF
--- a/templates/_partials/fields/_base.html.twig
+++ b/templates/_partials/fields/_base.html.twig
@@ -151,10 +151,12 @@
 
 {% if variant == 'inline' %}</div></div>{% endif %}
 
-{# Print postfix #}
-{% if field.definition['postfix']|default() is not empty %}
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):

bobdenotter:
Ah, this looks good. Thanks, @Spomsoree 👍

PR #3281, opened 2022-07-27
Removed the jsonb option to be able to update the mysql database

Spomsoree:
This fixed the problem described here: https://github.com/bolt/core/issues/1905#issuecomment-1196654934

I'm not able to test if fixtures can be installed with postgres. This may undo the fix made in the Issue above.

### DIFF
--- a/src/Entity/FieldTranslation.php
+++ b/src/Entity/FieldTranslation.php
@@ -21,7 +21,7 @@ class FieldTranslation implements TranslationInterface
      */
     private $id;
 
-    /** @ORM\Column(type="json", options={"jsonb": true}) */
+    /** @ORM\Column(type="json") */
     protected $value = [];
 
     public function getId(): ?int


bobdenotter:
I'm looking into this PR, and i don't know quite enough about Postgres to judge it properly. .

About your comment, for what was breaking: 

| (sfsu is my alias to update the schema: sf doctrine:schema:update)
| I run the following: sfsu --dump-sql && sfsu && sfsu --dump-sql:

To update the schema, I normally use `bin/console doctrine:schema:update --force`. Is that different than `sf` ?

Spomsoree:
Thats just a alias for `php bin/console` from the zsh symfony plugin: https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/symfony2/symfony2.plugin.zsh

It's exactly the same as `php bin/console doctrine:schema:update --force`

Spomsoree:
@bobdenotter As far as I know `jsonb` _only_ improves the speed of the postgres database. 

I just setup a postgres database, updated the schema, removed the `jsonb` option, generated migrations, which did result in an empty migrations file.

I think this should be safe to merge.

PR #3283, opened 2022-08-04
Update HomepageController.php

dotdevpro:
Fixes when i try open homepage which is not singleton (listing) with locale specified in url, ex: default is pt, which is root and i set to /en in url locale is always default.

### DIFF
--- a/src/Controller/Frontend/HomepageController.php
+++ b/src/Controller/Frontend/HomepageController.php
@@ -39,7 +39,7 @@ class HomepageController extends TwigAwareController implements FrontendZoneInte
 
         // Perhaps we need a listing instead. If so, forward the Request there
         if (! $contentType->get('singleton') && ! isset($homepageTokens[1])) {
-            $params = array_merge($this->request->query->all(), ['contentTypeSlug' => $homepage]);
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Nice catch! Thank you, @dotdevpro

src/Controller/Frontend/HomepageController.php
```diff
@@ -39,7 +39,7 @@ public function homepage(ContentRepository $contentRepository): Response
 
         // Perhaps we need a listing instead. If so, forward the Request there
         if (! $contentType->get('singleton') && ! isset($homepageTokens[1])) {
-            $params = array_merge($this->request->query->all(), ['contentTypeSlug' => $homepage]);
+            $params = array_merge($this->request->query->all(), ['contentTypeSlug' => $homepage, '_locale' => $this->request->getLocale()]);
```
bobdenotter, line 42:
```suggestion
            $params = array_merge($this->request->query->all(), [
                'contentTypeSlug' => $homepage,
                '_locale' => $this->request->getLocale()
            ]);
```

bobdenotter (approved):

PR #3284, opened 2022-08-04
Update ContentExtension.php

dotdevpro:
Fixes type error on php8+ due to first argument being null in some cases


dotdevpro:
I found new cases due to this issue will check more and reopen later, problem is that when route is forwarded there is no _route attribute i actually found solution need check little bit.

PR #3292, opened 2022-08-17
Add extra fields option for `image` and `imagelist` fields

xiaohutai:
For both `image`s and `imagelist`s, you can now add an extra property `extra`:
```
extra:
    title:
        label: Titel
        placeholder: Placeholder
    caption:
        label: Caption
    teaser:
        label: Teaser
        placeholder: De teaser in het Nederlands
    teaser_en:
        label: Teaser (EN)
        placeholder: De teaser in het Engels
```

For now I have enabled `label` and `placeholder`, but since it's an key,value-array we can expand it if needed. (I initially had it as: `attrib: [ foo, bar, baz ]` to mimick Bolt 3).

---
This seems to work for me. Somehow the solution looks way simpler than I thought. So please double-check if I haven't missed anything.

### DIFF
--- a/assets/js/app/editor/Components/Image.vue
+++ b/assets/js/app/editor/Components/Image.vue
@@ -34,16 +34,36 @@
                     />
                 </div>
                 <div v-if="includeAlt" class="input-group mb-3">
-                    <input
-                        v-model="altData"
-                        :title="name + ' alt'"
-                        class="form-control"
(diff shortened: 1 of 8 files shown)


bobdenotter (approved):
Did two ninja-edits! Looks good, otherwise! 👍

Issue #3019, opened 2021-12-16
Query on select field with integer values doesn't work anymore


jordanaubert:
I can't fetch Content using Query service on select field with integer values.

Details
-------

| Question                                              | Answer
|------------------------------------------------------ | ------
| Relevant Bolt Version                                 | 5.0.7
| Install type                                          |  composer install
| <abbr title="Backwards Compatibility">BC</abbr> Break | **yes**
| PHP version                                           | 7.4 |
| Web server                                            | Nginx

Reproduction
------------
I have this content type defined in my config file :

```yaml
contenttype:
    name: contenttype
    fields:
        select_field:
            type: select
            values:
                0: value1
                301: value2
                302: value3
```

I want to fetch some data with following request using Query service :

```php
private Bolt\Storage\Query $$query;

public function __construct(Bolt\Storage\Query $query) {
    $this->query = $query;
}

$this->query->getContent('contenttype', ['select_field' => 301]);
```

No result returned while I can see existing value in database :

<img width="667" alt="Capture d’écran 2021-12-16 à 11 28 59" src="https://user-images.githubusercontent.com/16880485/146354794-c6d6462f-ea46-4de6-a2a6-75b2d59b917e.png">

It's possible to fetch data when values are stored as string instead of integer :
<img width="667" alt="Capture d’écran 2021-12-16 à 11 30 59" src="https://user-images.githubusercontent.com/16880485/146355022-53305aaf-5c20-4057-8130-1b98505497f9.png">

It's a BC break because the query worked on Bolt `4.2.5`

jordanaubert:
This is probably why we can't query json field with integer values :

> [JSON_SEARCH(json_doc, one_or_all, search_str[, escape_char[, path] ...])](https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html#function_json-search)
> Returns the path to the _**given string**_ within a JSON document.

It seems that by design, `JSON_SEARCH` works with strings only.

see also https://bugs.mysql.com/bug.php?id=79233

jordanaubert:
Issue fixed by #3302

Issue #3063, opened 2022-01-26
Fetching records based on select field value gives no results


SanderOnline:
When fetching records based on the ID of select field from another contenttype it gives 0 results

Details
-------

| Question                                              | Answer
|------------------------------------------------------ | ------
| Relevant Bolt Version                                 | 5.1.0
| Install type                                          |  Composer install
| PHP version                                           | 7.2
| Web server                                            | Apache

Reproduction
------------

2 contenttypes were used for this:

Contenttype to define different categories:

```
support:
    name: Support Categories
    slug: support
    singular_name: Support
    singular_slug: support
    fields:
        title:
            type: text
            class: large
            group: content
            label: Support category name
        teaser:
            type: html
            height: 150px
            label: Short text below the title                                            
        slug:
            type: slug
            uses: title
            group: Meta         
    recordsperpage: 100
    default_status: published

```

Contenttype for the different articles:

```
articles:
    name: Articles
    slug: articles
    singular_name: Article
    singular_slug: article
    fields:
        title:
            type: text
            class: large
            label: Page title
            group: content
        subtitle:
            type: text
            label: Subtitle below the title, optional
        teaser:
            type: redactor
            label: Introtext below the subtitle, optional
        supportcategory:
            type: select
            values: support/{title}
            sort: title
            autocomplete: true
            label: Support category
        slug:
            type: slug
            uses: title
            group: Meta
    order: title
    records_per_page : 50
    default_status: published
```

Having the following logic:
- There is a support listing page with all the categories shown
- The support record shows all the articles linked to this specific support page through the "supportcategory" field

### Bug summary

When trying to fetch all the articles with a specific "supportcategory" id it gives 0 results when using either
`{% setcontent articles = "articles" where { 'supportcategory': 36 } %}`
or 
`{% setcontent articles = "articles" where { 'supportcategory': record.id } %}`

When printing the query it shows:
```
SELECT content FROM Bolt\Entity\Content content LEFT JOIN content.fields fields_supportcategory LEFT JOIN fields_supportcategory.translations translations_supportcategory WHERE content.contentType = :ct0 AND ((JSON_SEARCH(translations_supportcategory.value, 'one', :supportcategory_1) != '' AND fields_supportcategory.parent IS NULL AND fields_supportcategory.name = :field_supportcategory) AND content.status = :status_1) ORDER BY content.publishedAt DESC

ct0: "articles"
supportcategory_1: 36
status_1: "published"
field_supportcategory: "supportcategory"
```

[...]

PR #3302, opened 2022-08-24
Fix setcontent with `where { }` clause filtering on a foreign `id` in MySQL

bobdenotter:


mcdennem (approved):
Nice! 👍

jordanaubert:
This may fix #3019 and #3063 !

PR #3305, opened 2022-08-29
fix #3304 by injecting a validator on creation of conent as well.

### LINKED ISSUE #3304: Content validation is skipped on the server while creating content in the Bolt Backend
We noticed that while creating our own validation rules they work splendidly when updating entities but were not triggered on content creation. 

This seems to be because of how `\Bolt\Controller\Backend\ContentEditController::new` calls `\Bolt\Controller\Backend\ContentEditController::save`. `save ` gets injected with a validator when called directly from a route match, but without when called from `new`. Due to this, the validation is skipped.

Details
-------

| Relevant Bolt Version                                 | 5.1.7 and higher
| <abbr title="Backwards Compatibility">BC</abbr> Unsure (I don't know if people are now expecting validation to be skipped)
| Web server                                            |  Nginx
| For UX/UI issues                                      | Firefox

Reproduction
------------

### Bug summary

We noticed that while creating our own validation rules they work splendidly when updating entities but were not triggered on content creation. 

This seems to be because of how `\Bolt\Controller\Backend\ContentEditController::new` calls `\Bolt\Controller\Backend\ContentEditController::save`. `save ` gets injected with a validator when called directly from a route match, but without when called from `new`. Due to this, the validation is skipped.
### Steps to reproduce

- Create an entity with validation rules (linke min/max)
- Create a new entity in the back end that breakes that rule

### Expected result

Entity is not created and a validation message is shown.

### Actual result

Entity is created.

mint-jar:
Fixes Issue #3304


bobdenotter (approved):
Looks good. Thank you, @mint-jar 👍

PR #3307, opened 2022-08-30
Issue: #3186. Fix deprecation notice in PHP8.0 and above

### LINKED ISSUE #3186: PHP 8.1 deprecation: `Optional parameter $qb declared before required`
```
$ bin/console c:c
PHP Deprecated:  Optional parameter $qb declared before required parameter $utils is implicitly treated as a required parameter in /usr/local/www/arb/vendor/bolt/core/src/Storage/SelectQuery.php on line 112
```

abdellahrk:
Fixes #3186. Fix deprecation notice in PHP8.0 and above

### DIFF
--- a/src/Storage/SelectQuery.php
+++ b/src/Storage/SelectQuery.php
@@ -110,11 +110,11 @@ class SelectQuery implements QueryInterface
      * Constructor.
      */
     public function __construct(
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Nice catch! Thanks, @abdellahrk 👍

Issue #1975, opened 2020-10-09
Meta snippet insert error


ziselive:
Insert the Meta Generator tag and canonical twice in the front-end head:
```
<link rel="canonical" href="http://www.ts.localhosthttp://ts.localhost/">
<meta name="generator" content="Bolt">
<link rel="canonical" href="http://www.ts.localhosthttp://ts.localhost/">
<meta name="generator" content="Bolt">
```

Details
-------

| Question                      | Answer
|----------------- | ------
| Relevant Bolt Version | 4.1
| Install type                   | Composer install 
| PHP version                 | 7.4 |
| Web server                  | Apache

bobdenotter:
I've seen this happen once as well, a while ago. But, I've not been able to reproduce it consistently since. 

What happens here is that the 'snippets' queue is run twice. We'd need to figure out under which circumstances that happens, and then we can make a good fix for this.

jordanaubert:
Hi, I stumbled on this issue and got the following in the `head` on my page : 
```
<link rel="canonical" href="http://localhost">
<link rel="canonical" href="http://localhost">
<link rel="canonical" href="http://localhost">
<link rel="canonical" href="http://localhost">
```
The way to reproduce it is to reach a singleton page, the `forward` call on `\Bolt\Controller\Frontend\ListingController::listing` creates a sub-request, triggering `WidgetSubscriber` and `ProcessWidgetsQueueSubscriber` resulting in two double insertion. I'll try to make an MR to fix it soon.

PR #3308, opened 2022-08-30
Insert meta snippets on main request only

jordanaubert:
Fix issue #1975 where meta snippets are inserted multiple times, via sub-requests for example.

I added an option in config to give the ability to deactivate default canonical insertion.


bobdenotter (approved):
Hi @jordanaubert, this is an excellent fix! Thanks. 👍

PR #3313, opened 2022-09-13
Add setId() function in User Entity

nestordedios:
To improve import functionality of the Conimex extension this function is necessary to allow the importer to set the ID of the user to whatever the id is in the exported users file. See https://github.com/bobdenotter/conimex/pull/62


bobdenotter (approved):
Cool, merging in! 👍

PR #3314, opened 2022-09-15
Characters escaped when displaying the title bloc in templates

fsidobre:
Hello,

Just noticed that the title in `layout*.html.twig` files is a little bit too much escaped.

There's a `|striptags` filter which is indeed useful because subsequent title blocks may have html elements in it. But the function does too much escaping.

For example, the title of the login page in french is "S'identifier" which is then displayed as : 
wrong-title-escaping

As the content is already stripped from tags, we can just display a `raw` value.


bobdenotter (approved):
Thanks, @fsidobre 👍

PR #3320, opened 2022-09-21
The `|excerpt-filter` now also excerpts fields inside Collections and Sets

bobdenotter:
… Sets


nestordedios (approved):
NIce! Looks good! 👍

PR #3323, opened 2022-09-25
Fix #3112.

### LINKED ISSUE #3112: Error when no user with ID #1 is in the database
Trying to solve another issue (which I don't exactly remember but might have been caused by persistent cookies in my browser actually) I deleted all users and then started fresh, using `bin/console bolt:add-user`. The new users had ID's starting from `5` or so. 

After this, when trying to get a page listing in the Bolt Backend, this error was shown (below). The error was solved by going into the database and setting the ID of one user to `1`  (I picked the user with a DEVELOPER role, not sure if that made any difference).
Screenshot 2022-03-06 at 21-41-22 An exception has been thrown during the rendering of a template ( Entity of type 'Bolt En

abdellahrk:
Add id '1' to the first existing admin or if not available to the current user being created.

### DIFF
--- a/src/Command/AddUserCommand.php
+++ b/src/Command/AddUserCommand.php
@@ -66,7 +66,10 @@ class AddUserCommand extends Command
     /** @var Config */
     private $config;
 
-    public function __construct(EntityManagerInterface $em, UserPasswordHasherInterface $passwordHasher, ValidatorInterface $validator, Config $config)
+    /** @var UserRepository */
+    private $userRepository;
+
+    public function __construct(EntityManagerInterface $em, UserPasswordHasherInterface $passwordHasher, ValidatorInterface $validator, Config $config, UserRepository $userRepository)
     {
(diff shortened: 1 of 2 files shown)


bobdenotter:
Hmm, I'm not sure this is a good fix.. This will effectively "rename" a user to have id `1`. If that specific user already has content, it'll be linked to their id, which would cause that content to be 'orphaned', right?

abdellahrk:
@bobdenotter yes right. Then let us just make the currently created user have the id `1`?

bobdenotter:
> @bobdenotter yes right. Then let us just make the currently created user have the id `1`?

Yes, you mean whenever we create a user, to ensure there's one with `id` `1`?

abdellahrk:
> Yes, you mean whenever we create a user, to ensure there's one with `id` `1`?

Yes @bobdenotter. Or is there a better solution than this?

Issue #3330, opened 2022-10-04
postUpdate fails while composer update


samarija:
"composer update" fails to complete the postUpdate - instead the "vendor/bolt/core/bin/fix-bundles.php" file is opened in the IDE.

Details
-------

| Question                                              | Answer
|------------------------------------------------------ | ------
| Relevant Bolt Version                                 | 5.1.16
| Install type                                          |  Composer install
| PHP version                                           | 8.1.10
| Web server                                            | Apache  [ 2.4.47 ]
| OS | WIndows 10

Reproduction
------------

### Bug summary

After running the composer update command the postUpdate part of the process fails (timeout). At the same time "vendor/bolt/core/bin/fix-bundles.php" is opened in the IDE assigned to .php-files.

### Steps to reproduce

run composer update

### Expected result

finish the update succesfully 

### Actual result

error while runninig postUpdate

### Solution?

remove the .php ending for "vendor/bolt/core/bin/fix-bundles.php" in bin/composer-script/PostUpdateScript.php ( line 13 )

bobdenotter:
Closed in #3331

PR #3331, opened 2022-10-04
fix "fix-bundles" in PostUpdateScript.php

samarija:
Remove .php ending of fix-bundles script #3330


bobdenotter (approved):
Thanks for the fix, @samarija 👍

PR #3333, opened 2022-10-05
Ensuere we return the found ImageField, and not the surrounding Set or Collection

bobdenotter:


nestordedios (approved):
Good catch!! This looks good! 👍

PR #3334, opened 2022-10-10
Check for empty request before using it

bobvandevijver:
Crashed today in one of my production environments:

image

Unfortunately I couldn't use the null-safe operator here :(


bobdenotter (approved):
Thanks, @bobvandevijver

PR #3338, opened 2022-10-22
Fixed some typos and made some phrases more natural/native sounding

AndrewDawes:
I went through the translations and fixed any typos I found. I noticed some of the phrases didn't sound that natural or native, so I adjusted those, I understand if you don't like the new wording :-)


bobdenotter (approved):
Thanks, @AndrewDawes 👍

PR #3350, opened 2022-11-08
Added the thumbnail path prefix config option

Spomsoree:
to be able to use a different thumbnail route.

I'll use this to create my own thumbnails route and convert images to webp.
If there is a better way to achieve this, let me know.

### DIFF
--- a/config/bolt/config.yaml
+++ b/config/bolt/config.yaml
@@ -28,7 +28,7 @@ caching:
     frontend_menu: ~
     backend_menu: ~
     files_index: ~
-    
+
 # The theme to use.
 #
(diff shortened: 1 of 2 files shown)


bobdenotter:
If we do this, the `/thumbs/` route will break, right? Maybe we could make that route work with the configuration as well?

bobdenotter:
I meant, if you change the prefix to something else than `/thumbs`, this will surely break: 

https://github.com/bolt/core/blob/master/src/Controller/ImageController.php#L39

Or at least, bolt will generate different routes, that won't work. Shouldn't we make it so that those keep working? Kinda like if we change the "backend_url" in `services.yaml`, it'll 'just work'. :-)

PR #3358, opened 2022-11-21
Make guesstimateAuthor return type nullable

clem:
As seen with @bobdenotter, author can be `null` sometimes, especially on ContentType creation from forms with `bolt/forms`.


bobdenotter (approved):
Cool, thanks @clem 👍

PR #3365, opened 2022-11-25
Un bidiretionalize related content

doenietzomoeilijk:
This switches the boolean `$bidirectional` parameter on the related-methods (which defaulted to `true` and never actually did anything) to a string, either "to", "from" or "both" (the latter being the default). Choosing "to" or "from" limits the related content to content that either relates _to_ the current content, or _from_ it. Choosing "both" gets you either.

For backwards compatibility purposes, you can also supply a boolean, where `true` will map to "both" and`false` will map to "to".

Touches the `related_content_by_type` and `related_content` Twig filters, as documented [here](https://docs.boltcms.io/5.0/twig-components/filters#related) - this would need an update in the documentation as well.


bobdenotter:
Hi @doenietzomoeilijk! Looks like ECS and Phphstan are complaining. ECS can be fixed automatically: `vendor/bin/ecs check src --fix` 

The other one might be a bit trickier to fix: https://github.com/bolt/core/actions/runs/3550478854/jobs/5963901751

bobdenotter:
> The "fix" that ecs proposes means making the docblock... incomplete? 

It's mostly that it's redundant. If the argument is already typed in the function, it feels there's no need in repeating it.

bobdenotter (approved):
Cool, looks good! 👍 

> Touches the related_content_by_type and related_content Twig filters … - this would need an update in the documentation as well.

Yes, please. ;-)

PR #3367, opened 2022-12-02
Fix `excerpt` when content would be too short _after_ trimming

xiaohutai:
but would be long enough _before_ trimming in order to skip returning the `fulltext` instead.

The issue lies in a piece of small text that would be longer than the excerpt length `relLength`, but when trimmed at https://github.com/bolt/core/blob/master/src/Utils/Excerpt.php#L110-L112 would overflow resulting in a negative starting position. This gives an undesired result (and potenttially no highlight).


bobdenotter (approved):
Nice catch, thanks! 👍

PR #3368, opened 2022-12-02
Made the is current filter remove the current locale 

Spomsoree:
before comparing, to the current route to get continuous results across all languages.

/en/products becomes /products, which we then compare to the path info.

### DIFF
--- a/src/Twig/FrontendMenuExtension.php
+++ b/src/Twig/FrontendMenuExtension.php
@@ -58,8 +58,11 @@ class FrontendMenuExtension extends AbstractExtension
 
     public function isCurrent($item): bool
     {
+        $currentRequest = $this->requestStack->getCurrentRequest();
+        $currentLocale = $currentRequest->getLocale();
         $uri = $item['uri'] ?? '';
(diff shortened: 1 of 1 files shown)


src/Twig/FrontendMenuExtension.php
```diff
@@ -58,8 +58,11 @@ public function renderMenu(Environment $twig, ?string $name = null, string $temp
 
     public function isCurrent($item): bool
     {
+        $currentRequest = $this->requestStack->getCurrentRequest();
+        $currentLocale = $currentRequest->getLocale();
         $uri = $item['uri'] ?? '';
+        $currentUrl = str_replace('/' . $currentLocale, '', $currentRequest->getPathInfo());
```
bobdenotter, line 64: I think this is a good fix in general, just not sure if this won't cause an edgecase.. Would it be better to include the the slash _after_ the locale as well? 

```suggestion
        $currentUrl = str_replace('/' . $currentLocale, '/', '/', $currentRequest->getPathInfo());
```

Spomsoree: You are right

bobdenotter (approved):
Thanks, @Spomsoree 👍

PR #3371, opened 2022-12-05
Don't blindly pre-select the first option in a filter.

### LINKED ISSUE #3370: Relations get "pre-filled" with the first option when empty
When creating a new content record with a relation, or when editing one that doesn't yet have that relation filled, the first option of the selectable related records is selected. You'd expect that field to remain empty until you actually select something.

Details
-------

| Relevant Bolt Version                                 | 5.2.0-beta / master
| Install type                                          | irrelevant, either composer or direct checkout

Reproduction
------------

The default setup in the bolt/core repo exhibits the behaviour, so clone bolt/core, checkout master or 5.2.0-beta9, navigate to the backend and create a new item for any contenttype with relations (which is most of them). Instead of an empty field (this is a new content record, it _can't have_ relations), you see the first item in the dropdown already selected. Change the order of those items in contenttypes.yaml, and the pre-filled value will change accordingly.

Cause(s)
-----------

The actual cause is in Select.vue, and was introduced [here](https://github.com/bolt/core/commit/26b94354e32c3848299f42e083fc4c9050e404a6#diff-fb54298fece426a644e9c47df4a2ed9f49b86e494f27cdc1edcb4c8c2996927aR83). "If the list of selected items is empty, select the first one", yep, that sounds about right. It's part of [this PR](https://github.com/bolt/core/pull/994), which introduces required fields.

However, during a `git bisect` session, the issue somehow first showed [at this commit](https://github.com/bolt/core/commit/f861dee6e52a5c386a3205df763ceb24e4031ea3), which does somewhat related things, but doesn't _directly_ cause the issue. 

So I'm kinda... confused. fixing the actual issue is easy enough (take out those three lines , done), but I'm not sure if those three lines were, as the commit indicated, part of a WIP and meant to be changed later, simply forgotten, or if there's some actual need for them. In case of the latter, I'm not sure what that need _is_. Hoping @I-Valchev and/or @bobdenotter can weigh in here. 

Apart from that, and if it wouldn't break anything else, a fix would of course be appalingly simple. ;)

doenietzomoeilijk:
This eventually caused Relations selects for new content to have the first Related option pre-selected. Not entirely sure where this change came from and what the thought behind it was, so I cannot in full confidence say that this one won't break anything. I'd like @bobdenotter and @I-Valchev (and possibly @nestordedios ?) to weigh in.

Fixes #3370

### DIFF
--- a/assets/js/app/editor/Components/Select.vue
+++ b/assets/js/app/editor/Components/Select.vue
@@ -135,10 +135,6 @@ export default {
             })
             .filter(item => undefined !== item);
 
-        if (filterSelectedItems.length === 0) {
-            filterSelectedItems = [_options[0]];
-        }
-
         this.selected = filterSelectedItems;
     },
     methods: {
--- a/assets/js/app/editor/Components/Select.vue
+++ b/assets/js/app/editor/Components/Select.vue
@@ -94,6 +94,7 @@ export default {
         classname: String,
         autocomplete: Boolean,
         errormessage: String | Boolean, //string if errormessage is set, and false otherwise
+        required: String | Boolean,
     },
     data: () => {
         return {
@@ -134,6 +135,10 @@ export default {
                 }
             })
             .filter(item => undefined !== item);
+
+        if (!!this._props.required && filterSelectedItems.length === 0) {
+            filterSelectedItems = [_options[0]];
+        }
 
         this.selected = filterSelectedItems;
     },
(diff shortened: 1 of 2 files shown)


bobdenotter (approved):
This looks good! I'll merge it in, and hopefully I can package a new beta, so we test it out in the field. 👍

PR #3383, opened 2023-01-07
3115-thumbnails-parameters

viktor-wolf:
The thumbnail quality setting works now. I have also added a `$quality` parameter as the last parameter of the `thumbnail` filter. I have also brought the quality setting in `GeneralParser`'s default config in line with the default in `config.yaml`. As a result of my changes, quality is now the 3rd element of generated thumbnail paths. 

I have removed superfluous settings from `config.yaml` (`cropping` and `allow_upscale` – both seem to be superseded by the `$fit` parameter of the `thumbnail` filter).

I have tried to make my changes as undisruptive as possible, but I still recommend giving them a good look over.

### DIFF
--- a/config/bolt/config.yaml
+++ b/config/bolt/config.yaml
@@ -166,9 +166,6 @@ records_per_page: 8
 # Default settings for thumbnails.
 #
 # quality:       Quality should be between 0 (horrible, small file) and 100 (best, huge file).
-# fit:           One of either none, crop (= crop-center), crop-top-left, crop-top, crop-top-right, crop-left, crop-right, crop-bottom-left, crop-bottom or crop-bottom-right.
-# allow_upscale: Determines whether small images will be enlarged to fit
-#                the requested dimensions.
 # save_files:    Save files in the thumbs/ folder, so subsequent requests will serve file directly. Great for performance
 #
 # Note: If you change these values, you might need to clear the cache before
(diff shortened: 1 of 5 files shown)


bobdenotter (commented):
Hi @viktor-wolf. Thanks or the PR. 

Why did you choose to make quality the third param? Doesn't that break existing thumbs that have a `fit`? like: 

```
https://playground.wip/thumbs/600×361×f/stock/image_19493.jpg
https://playground.wip/thumbs/600×361×s/stock/image_19493.jpg
```

viktor-wolf:
Hi,

my thought process went like this:

1. I thought it would be nice to be able to pass quality as an argument to the `thumbnail` function.
2. I have added `$quality` as the last parameter of the function, in order to preserve backwards compatibility.
3. In `ThumbnailHelper` I have added it as the third argument when `paramString` is being constructed, because it seemed like the most straightforward way of passing it to `ImageController` without needing to completely overhaul the parsing of `fit` and `location` there.
4. I have incremented the indices that are being checked/used as values while parsing the `paramString` by 1. Since quality is always included in the `paramString` (by setting it in a template, or setting it in the config, or from `GeneralParser` default settings), this should always work and preserve backwards compatibility.

Since then, we have talked on Slack, and you said the quality parameter on the `thumbnail` function probably shouldn't be included. If we decide to remove it, we can remove all the changes to the function and `ThumbnailHelper`, and simply pull the quality from configs directly in `ImageController`.

bobdenotter (approved):
Ok, cool! Let's merge! 👍

viktor-wolf:
Hi, thanks for approving the merge request. However, there a couple of loose ends still remain:

1. I was waiting for you to confirm that I should remove the thumbnail parameter from the thumbnail filter/function. Have you decided that it should be left in?
2. Are there any post-merge tasks I should take care of? Updating the docs and such?

PR #3390, opened 2023-01-19
Fix #3389 : Do not display twig value for SetField

### LINKED ISSUE #3389: [BC Break - 5.1.16] Get CheckboxField from SetField in twig doesn't parse his value
Bug introduced in `5.1.16` by https://github.com/bolt/core/pull/3320

Details
-------

| Relevant Bolt Version                                 | 5.1.19
| <abbr title="Backwards Compatibility">BC</abbr> Break | yes
| Web server                                            | Nginx

Reproduction
------------

### Bug summary

It seems that since `SetField` is `Excerptable`, displaying the value of a checkbox field doesn't work as intended. I can't see why yet but removing the interface fix the issue.

### Steps to reproduce

Have a content type with a `SetField` with a `CheckboxField` in it. If you dump the value of `record.setField.checkboxField`, it will return the `CheckboxField` object instead of `true` or `false`.

### Expected result

use `record.setField.checkboxField` to display the value instead of using `record.setField.checkboxField.value[0]`

-----------

## EDIT

The condition in `Content` `__call` method converts `SetField` into an array of fields since it's now `Excerptable` !

jordanaubert:
Fix issue #3389 where SetField is converted into an array in twig instead of keeping the object, resulting on some BC breaks.

### DIFF
--- a/src/Entity/Content.php
+++ b/src/Entity/Content.php
@@ -11,6 +11,7 @@ use ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\SearchFilter;
 use Bolt\Configuration\Content\ContentType;
 use Bolt\Entity\Field\Excerptable;
 use Bolt\Entity\Field\ScalarCastable;
+use Bolt\Entity\Field\SetField;
 use Bolt\Enum\Statuses;
 use Bolt\Repository\FieldRepository;
 use DateTimeZone;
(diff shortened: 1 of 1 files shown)


bobdenotter:
Nice catch! I'll merge it in, and release a `5.1.20` version! 👍

bobdenotter (approved):

PR #3395, opened 2023-01-24
Fix phpstan errors

enrise-mbraam:
Fix deprecation errors that appear when using Bolt on PHP 8.1 (and I guess 8.2)

Implementing `\Iterator` is not needed, I believe.

Also fixed another error Phpstan reported in the same way it was dealt with in `src/Controller/ErrorController.php`

See image for errors that were occurring. 

deprecation notices bolt


bobdenotter (approved):
Thanks, @enrise-mbraam. Looks good! 👍

PR #3399, opened 2023-01-31
Fix excerpt of search result when there is no whitespace in text

mcdennem:
… term in the result


bobdenotter (approved):
Thanks, @mcdennem 👍

PR #3400, opened 2023-02-03
Partial reverse of #3395, we really need those `\Iterator`s

bobdenotter:
Pinging @enrise-mbraam , I had to partially revert your recent PR, because the removal of `Iterator` causes things like this to stop functioning: 

```
{% for image in record.gallery %}
    {{ image|popup() }}
{% endfor %}
```

Could you take another look at what was breaking on your end, and if we can do an alternative fix for that? :-)

### DIFF
--- a/src/Entity/Field/CollectionField.php
+++ b/src/Entity/Field/CollectionField.php
@@ -17,7 +17,7 @@ use Doctrine\ORM\Mapping as ORM;
 /**
  * @ORM\Entity
  */
-class CollectionField extends Field implements Excerptable, FieldInterface, FieldParentInterface, ListFieldInterface, RawPersistable
+class CollectionField extends Field implements Excerptable, FieldInterface, FieldParentInterface, ListFieldInterface, RawPersistable, \Iterator
 {
     use FieldParentTrait;
     use IterableFieldTrait;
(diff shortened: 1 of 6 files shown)


enrise-mbraam:
Too bad! 
To answer your question: We get deprecation notices; see screenshot of composer install. But we also see these messages on the site (on first page load, after that they're gone). I'm sure we can find a way to suppress them on our end.

Given that my proposed fix is not an option, you might be able to look at other options

- Removing the Trait, as it's conflicting with the Iterator class
and/or
- Dropping support for 7.2 (or 7.x) 

We'd be happy to discuss this. 

Schermafbeelding 2023-02-06 om 13 09 10

bobdenotter:
I think the ideal solution would be to improve the code, so it doesn't throw the deprecations, but it also doesn't introduce any backwards incompatible issues. 

Working on the deprecations is a boon, regardless. We'd need to resolve them anyways if we want to bump to Symfony 6.x

Speaking of which: 

> Dropping support for 7.2 (or 7.x)

Yes, that'd be really nice to do, but it needs to go in hand with bumping Symfony to 6.x, because then we can require `^8.1`

PR #3402, opened 2023-02-07
Fixed thumbnail sizes

Spomsoree:
which did not fit the size because of a incomplete regex replace.

Further reference: 
https://boltcms.slack.com/archives/C094GL7ME/p1675776303978339?thread_ts=1675769854.480089&cid=C094GL7ME


bobdenotter (approved):
Apologies for this silly bug, and thanks for helping solve it! 👍

PR #3417, opened 2023-02-21
allow_twig : prevent saving error

Filoz:
allow_twig : prevent saving error (translations getting cleared if Twig code contains errors)

When inserting twig code (for example in an Article field), if the code contains one or more error, the content is saved in the current locale... but others translations are cleared.


bobdenotter:
Nice catch! Thanks, @Filoz 👍

PR #3418, opened 2023-02-27
Fix #3375 REST API DELETE method not working

### LINKED ISSUE #3375: REST API DELETE method not working
When attempting to delete a record using the REST API, a status "204 No Content" is returned (which indicates that the request was successful) but the record still remains unchanged.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 5.1.18
| Install type                   | Composer install
| PHP version                 | 7.2
| Web server                  | Apache / Built-in

Reproduction
------------

### Bug summary

    * Most likely an APIPlatform implementation fault
    * Annotations for deleting content using the API are present in Bolt\Entity\Content

### Specifics

    * API route is `/api/contents/{id}`, method is DELETE
    * API definition can be found by default under `/bolt/api`
    * Bolt was installed using Composer CLI
    * Bolt version is 5.1.28
    * Clients used to verify the problems are Postman and PHP/Curl

### Steps to reproduce

    * Create a User with ROLE_WEBSERVICE
    * Create an example record of any content type
    * Make an HTTP call with DELETE method to `/api/contents/{id}` -> should return status 204
    * Verify that the record has not been deleted

### Expected result

    * The record is deleted from the database completely

### Actual result

    * The record remains unchanged

jeremycr:


bobdenotter (approved):
Thanks, @jeremycr 👍

PR #3420, opened 2023-03-07
Update ImageExtension.php

jpshdev:
Added class parameter with default value so a class can be assigned when calling the function from Twig.


bobdenotter:
Ping @jpshdev ;-)

jpshdev:
Sorry about that. Merged the suggestion.

bobdenotter:
@jpshdev I think you've merged in only one of the two suggestions ;-)

bobdenotter (approved):

PR #3421, opened 2023-03-08
try to fix failing composer script on windows

samarija:
commands executed from the post-update-cmd are not executed on windows. unshifting the executable as parameter seems to the fix the problem.


bobdenotter (approved):
Neat little fix! Thanks, @samarija 👍

Issue #3426, opened 2023-03-15
French status translation to improve


jodou92:
I find that some french status translation are confusing
```
status.held: Tenu
status.timed: Chronométré
```
In my opinion theses translations are literally correct but not really understandable for a French speaking contributor. 
I suggest to use instead :
```
status.held: Non publié
status.timed: Publication planifiée
```

I am going to make a PR for this

clem:
@jodou92 As the PR has been merged, I think you can close your issue :) Thanks for the french translation enhancement 🙌

PR #3427, opened 2023-03-15
suggests better french translations for statuses #3426

jodou92:
Give more understandable translations for french statuses. Refers to the issue : https://github.com/bolt/core/issues/3426

Thanks Romich' to have found the issue !


bobdenotter (approved):
Thanks, @jodou92 👍

PR #3445, opened 2023-04-04
Only attempt to get a link to content with a slug for the locale.

### LINKED ISSUE #3442: Content with `null` slug breaks frontend display
I have a record with two locales (nl/en). The nl content is filled, but the en content is not, and the record has an empty slug (as in, `null`) for en. In my config, `fallback_when_missing` is set to `false`.

In my template, I call `htmllang()`, which in turn calls `getLocales()`, which tries to generate links for every locale for that contentitem, which breaks. `Content::getSlug()` does a fallback for a slug that is an empty string, but not for one that is null. This results in an attempt to generate an URL for the `record_locale` route, but without the required `slugOrId` parameter.

I propose to explicitly cast the found slug in `Content::getSlug()` to string, so that a null value gets caught correctly. Either that, or `LocaleHelper::getLocales` should not try to generate a URL for that record/locale combo. Or, third option, we _should_ let things fall over, but in more troubleshootable way.

~~I'll provide a PR.~~ This needs a bit of discussion before I start shooting off PRs, I think.

doenietzomoeilijk:
This prevents us from attempting to create a route without a slugOrId, which would only make the Router sad.

Fixes #3442

### DIFF
--- a/src/Utils/LocaleHelper.php
+++ b/src/Utils/LocaleHelper.php
@@ -110,7 +110,7 @@ class LocaleHelper
                 $routeParams['slugOrId'] = $slug;
             }
 
-            if ($route && (!isset($content) || isset($routeParams['slugOrId']))) {
+            if ($route && (stripos($route, "bolt_") === 0 || (!isset($content) || isset($routeParams['slugOrId'])))) {
                 $locale->put('link', $this->getLink($route, $routeParams, $locale));
             } else {
                 // For edge-cases like '404', the `_route` is null.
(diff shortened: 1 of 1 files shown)


doenietzomoeilijk:
Current fix breaks switching the locale in the backend edit screen.

doenietzomoeilijk:
With the latest commit, things work. I'm not 100% confident that we'd only need to filter out the bolt_* routes, though. I *think* it should be enough (you want to be able to edit all content, but not review it through frontend or API, and editing stuff through the API uses IDs anyway). @bobdenotter care to comment on that?

bobdenotter:
> I'm not 100% confident that we'd only need to filter out the bolt_* routes

Theoretically, no.. But, since it also checks for a `slugOrId` parameter, I'd say it's fairly safe to assume.

Let's roll with this! 👍

bobdenotter (approved):

PR #3455, opened 2023-05-15
4.2.8

neohades:
Update for nexa/zip package to be compatible with PHP 8


bobvandevijver:
I'm going to close this PR; I recommend upgrading to 5.2 which has support for PHP 8.

Issue #3471, opened 2023-08-03
The FrontendMenuBuilder doesn't check if the contentType exists


clem:
When menus on the frontend are built, the builder does not check if the related contentType exists, it assumes that it does.

Details
-------

| Question                                              | Answer
|------------------------------------------------------ | ------
| Relevant Bolt Version                                 | master
| Install type                                          |  GitHub checkout
| <abbr title="Backwards Compatibility">BC</abbr> Break | no
| PHP version                                           | 8.2 |
| Web server                                            | Built-in

Reproduction
------------

### Bug summary

* Create a menu item with a link like `whatever/an_existing_slug` (with `an_existing_slug` that really exists as a ContentType slug)
* Display the menu on the website
* The url of the menu item does not match the provided link

### What's next?

I will create a PR to fix this issue 😊

bobdenotter:
Fixed by #3472

PR #3472, opened 2023-08-03
Fix FrontendMenuBuilder.php when content type does not exists

clem:
A fix for the #3471 issue.


src/Menu/FrontendMenuBuilder.php
```diff
@@ -131,6 +131,9 @@ private function getContent(string $link): ?Content
 
         /** @var ContentType $contentType */
         $contentType = $this->config->getContentType($contentTypeSlug);
+        if ($contentType === null) {
```
bobdenotter, line 134:
```suggestion
        if (! $contentType instanceof ContentType) {
```

Keep PHPStan happy.

bobdenotter (approved):
Thanks, @clem

PR #3476, opened 2023-09-10
Full translation into Russian

hide-me:
1. Finished full translation into Russian.
2. Many strings was edited and fixed;
3. As well as many strings was translated first time.
4. Updated with new strings by messages.en.xlf and also transtated.

A complete and much more holistic translation made in the same style.


bobdenotter (approved):
Cool! 

Thank you, @hide-me 👍

PR #3485, opened 2023-10-05
Allow `lazy` as an option for select fields using contenttype-queries

xiaohutai:
### What this does

Allows the use of `lazy: true` for contenttype-queries on select fields. __Potentially__ useful on sites that have a lot of records that need to be processed before rendering the template/response. This is a proof-of-concept for now. I've chosen an opt-in method since this will be very niche, and we will have to see whether we get the gains, or not.

```diff
selected_pages:
    type: select
    label: "Select Pages"
    values: "pages/{contenttype} Nº{id}: {title} - {status}"
    autocomplete: true
    multiple: true
    limit: 1000
    variant: inline
+   lazy: true
```

### Issues

When components get initialized (mounted), they will each query the server for the list of options. This is to ensure that the selected values are properly selected and shown in the UI. The problem stems that they will do so simultaneously. After that the response will be stored in the cache.

- How to remake a proper Field instance from a bunch of strings?
- How to cache both client-side and server-side?
- What is the proper permission for the Async route?

### DIFF
--- a/assets/js/app/editor/Components/Select.vue
+++ b/assets/js/app/editor/Components/Select.vue
@@ -16,6 +16,7 @@
             tag-placeholder="Add this as new tag"
             tag-position="bottom"
             track-by="key"
+            :loading="isLoading"
             @tag="addTag"
         >
             <template v-if="name === 'status'" slot="singleLabel" slot-scope="props">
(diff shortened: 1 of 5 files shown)


assets/js/app/editor/Components/Select.vue
```diff
@@ -77,6 +78,7 @@
 <script>
 import Multiselect from 'vue-multiselect';
 import $ from 'jquery';
+//import axios from 'axios';
```
bobdenotter, line 81:
```suggestion
```

bobdenotter (approved):
Whoop, whoop, let's roll!

PR #3487, opened 2023-10-09
 Explicitly remove a removed Relation from the newRelations collection.

### LINKED ISSUE #3486: Removing a relation from content throws error on save
When editing Content and removing a Relation between this content and another contenttype, storing the Content will throw an error:

```
Bolt\Entity\Relation::getFromContent(): Return value must be of type Bolt\Entity\Content, null returned
```

Details
-------

| Install type                                          | Composer install
| <abbr title="Backwards Compatibility">BC</abbr> Break | no
| PHP version                                           | 8.1
| Web server                                            | Apache

Reproduction
------------

- Have two contenttypes with a relation between the two
- Have content for both types, with said relation used
- Remove the related content from one of the content entries, and attempt to save
- Receive an error

### Bug summary

The problem lies in the `$newRelations` collection still having the removed relation in its items; when doing the "re-add new relations" bit after the remove, it'll try to look up the related Content items for that relations, which will fail because they've already been nulled. The fix is to explicitly remove the relation from the `$newRelations` collection.

doenietzomoeilijk:
This fixes a bug where removing a relation between two Content entities
would result in an error.

Fixes #3486


bobdenotter:
Thanks for the fix, @doenietzomoeilijk 👍

bobdenotter (approved):

PR #3491, opened 2023-10-16
fix(editor): check only editor input:invalid

napley:
The fix is for the issue https://github.com/bolt/core/issues/3488

The solution is to check only input in #editor, otherwise the global search field is find first


bobdenotter (approved):
Thanks, @napley

PR #3492, opened 2023-10-16
fix(form): allow p and span in first character

napley:
The fix is for the issue https://github.com/bolt/core/issues/3489

I don't understand why is not authorized 🤔

### DIFF
--- a/templates/_macro/_macro.html.twig
+++ b/templates/_macro/_macro.html.twig
@@ -105,7 +105,7 @@
     'id':     Is a string
 #}
 {% macro generatePrefix(prefix, id) %}{% apply spaceless %}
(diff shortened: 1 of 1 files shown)


napley:
Is not working

bobdenotter:
Hi @napley ,

Since it's being filtered by `|markdown`, wouldn't this change mean you'd get `<div><p>my postfix here</p></div>`. In which cases does this solve an existing problem? :-)

napley:
It is for this issue https://github.com/bolt/core/issues/3489, when I have upgrade my project from Bolt 4.2, the test : 
"if prefix and not (prefix starts with '<p' or prefix starts with '<span' or prefix starts with '<div'" 
doesn't exist before, so all my prefix and postfix were not displayed anymore

PR #3497, opened 2023-10-26
Add YAML Migrations

xiaohutai:
See for the diff: https://gist.github.com/xiaohutai/b5f44fec671cfff6691255354f6f769b

Not sure if the uncommented `canonical: https://example.org` was intentional.

Need to test the integrations before merging in.


bobdenotter:
I think `JSON_VALUE` needs to be added: 

https://github.com/bolt/project/blob/5.2-php-8/config/packages/doctrine.yaml#L37

xiaohutai:
I found the one for for `doctrine.yaml`: https://github.com/bolt/core/issues/2088

Not sure if `5.1.6` is still the correct version here, but it seems the existing config in `5.1` already has it added: https://github.com/bolt/core/blob/5.1/config/packages/doctrine.yaml#L35C2-L35C21 (Ideally this migration should've been added with the config change 😋)

I've added those, but the bolt/core doesn't have the `cache.app` so I changed it in this commit as well.

bobdenotter (approved):
Whoop! 👍

PR #3509, opened 2023-12-13
Set limit per relation and select field option

nestordedios:
This PR adds a fix for a bug introduced on PR #3477(Allow setting a limit for the options displayed in relationships selects)

The limit option is now set to the general config option `maximum_listing_select` or 2000 by default.

If a relation or select field has the `limit` option defined, the limit will be set to the value of that option.


bobdenotter (approved):
Looks good! 👍

PR #3510, opened 2023-12-17
Handle incorrect parameters gracefully

bobvandevijver:
Solves a `Warning: Undefined array key 1` notice.

Some clients do not understand how to use UTF-8, and execute a request to for example `/thumbs/400%C3%83%C2%97400%C3%83%C2%97contain/news`, which contains `Ã` instead of the expected `×`. I do not know why `%83%C2` is inserted in the URL (it is some sort of Python based external scanner...), but I do believe this code should be robust enough to not throw a warning when this happens. With this change the raw parameters are checked for existence before being used, allowing for a sane fallback.

I do not believe the CI failures are related to this change.

### DIFF
--- a/src/Controller/ImageController.php
+++ b/src/Controller/ImageController.php
@@ -143,7 +143,7 @@ class ImageController
         $raw = explode('×', preg_replace('/([0-9])(x)([0-9a-z])/i', '\1×\3', $paramString));
 
         $this->parameters = [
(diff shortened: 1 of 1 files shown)


bobdenotter:
Neat, thanks! 👍 

> I do not believe the CI failures are related to this change.

No, they aren't. I just haven't been able to make time to fix those. :-/

bobdenotter (approved):

PR #3526, opened 2024-10-24
Set explicit doctrine/lexer version dependency

zanardigit:
Bolt v5 is not compatible with `doctrine/lexer` v3 which introduces a breaking change: the `token` is now an object of class `Doctrine\Common\Lexer\Token` and not an array anymore. A normal `composer update` in a Bolt CMS based project may now generate an error: `Cannot use object of type Doctrine\Common\Lexer\Token as array`. The error goes away if `doctrine/lexer` v2 is explicitly required.

On a separate PR I will propose an actual fix. This is about setting the dependency in Bolt CMS own `composer.json`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "doctrine/doctrine-bundle": "^2.4",
         "doctrine/doctrine-fixtures-bundle": "^3.4",
         "doctrine/doctrine-migrations-bundle": "^3.2",
+        "doctrine/lexer": "^2",
         "doctrine/orm": "^2.10",
         "drupol/composer-packages": "^2.0",
         "embed/embed": "^3.4",


bobvandevijver:
😞 I wanted to merge this, but one of the dev dependencies (`coduo/php-matcher`) requries `doctrine/lexer` v1. Would you be open to look into this?

zanardigit:
Let's see. I am already using the modified version successfully in PROD, as that conflict is on a DEV requirement, so I'd rather not revert back to `doctrine/lexer` v1. We could bump `php-matcher` to version 6 but is only compatible with PHP 8.1+, would that be ok?

bobvandevijver:
That depends on the outcome of this vote: https://github.com/bolt/core/discussions/3538. It has been opened just yesterday (so I want to give it some time), but I expect we will be dropping 8.0 and older soon.

bobvandevijver:
This has been superseded by #3558, which has been merged.

PR #3527, opened 2024-10-24
Add doctrine/lexer v3 compatibility

zanardigit:
Using the token as Token object is compatible with both v2 and v3 of doctrine/lexer

### DIFF
--- a/src/Doctrine/Query/Cast.php
+++ b/src/Doctrine/Query/Cast.php
@@ -52,7 +52,7 @@ class Cast extends FunctionNode
         $this->first = $parser->ArithmeticPrimary();
         $parser->match(Lexer::T_AS);
         $parser->match(Lexer::T_IDENTIFIER);
-        $this->second = $parser->getLexer()->token['value'];
+        $this->second = $parser->getLexer()->token->value;
         $parser->match(Lexer::T_CLOSE_PARENTHESIS);
     }
 }


bobvandevijver:
@zanardigit Can you rebase this and solve the PHPStan issue?

src/Doctrine/Query/Cast.php
```diff
@@ -52,7 +52,7 @@ public function parse(Parser $parser): void
         $this->first = $parser->ArithmeticPrimary();
         $parser->match(Lexer::T_AS);
         $parser->match(Lexer::T_IDENTIFIER);
-        $this->second = $parser->getLexer()->token['value'];
+        $this->second = $parser->getLexer()->token->value;
```
macintoshplus, line 55: @bobvandevijver I have written this code to fix it because the Lexer can be installed in version 2 or 3!

```suggestion
        // doctrine/lexer v2
        if (is_array($parser->getLexer()->token)) {
            $this->second = $parser->getLexer()->token['value'];
        /** @phpstan-ignore-next-line  */
        } elseif (is_object($parser->getLexer()->token)) { // doctrine/lexer v3
            $this->second = $parser->getLexer()->token->value;
        } else {
            throw new \LogicException('Unable to access parser token');
        }
```

zanardigit: @macintoshplus it's been a while since I tested this, but as far as i remember using the token as a `Token` object should be compatible with both v2 and v3 of `doctrine/lexer`. So, that is probably not required. Feel free to correct me if you found different results.

macintoshplus: @zanardigit, I have fixed this error today by requiring `doctrine/lexer` v2 in my project.
After the library downgrade the Bolt code works fine but, I don't want to keep this requirement in my project.

We don't know the version used in the project. We must assume works with version 2 or 3.

zanardigit: @macintoshplus cannot test right now, I trust you. I
n that case, my PR should be rejected; you should issue a new one with the proposed changes, which are safer and will work in any case, so they are definitely better.

PR #3529, opened 2024-11-20
Fix bug on remove data

jbcr:
Unable to remove data from persister.


jbcr:
Solved by this fork : https://github.com/bolt-oss/core/blob/master/UPGRADE-TO-OSS.md

PR #3532, opened 2025-03-06
:bug: Fix the pregmatchall count check

macintoshplus:
This PR fixes the way to check if the regex had a result.

`preg_match_all` returns the number of matches, 0 if no match or `false` if the regex throws an error.


src/Twig/CommonExtension.php
```diff
@@ -73,9 +77,9 @@ private function getLocale($item): ?string
     {
         if (is_string($item)) {
             $localepattern = '/^[a-z]{2}((-|_)[a-z]{2})?$/m';
-            preg_match_all($localepattern, $item, $matches);
+            $matchCount = preg_match_all($localepattern, $item);
 
-            return ! empty($matches) ? $item : null;
+            return $matchCount > 0 ? $item : null;
```
bobvandevijver, line 82: Would the usage of just `preg_match` not be enough here?

macintoshplus: Sure, I fix it

src/Twig/CommonExtension.php
```diff
@@ -72,11 +76,9 @@ public function generateSecret(string $slug): string
     private function getLocale($item): ?string
     {
         if (is_string($item)) {
-            $localepattern = '/^[a-z]{2}((-|_)[a-z]{2})?$/m';
-            preg_match_all($localepattern, $item, $matches);
-
-            return ! empty($matches) ? $item : null;
-        } elseif ($item instanceof Collection) {
+            return $preg_match('/^[a-z]{2}((-|_)[a-z]{2})?$/m', $item) === 1 ? $item : null;
```
bobvandevijver, line 79:
```suggestion
            return preg_match('/^[a-z]{2}((-|_)[a-z]{2})?$/m', $item) ? $item : null;
```

src/Twig/CommonExtension.php
```diff
@@ -72,11 +76,9 @@ public function generateSecret(string $slug): string
...
     {
         if (is_string($item)) {
-            $localepattern = '/^[a-z]{2}((-|_)[a-z]{2})?$/m';
-            preg_match_all($localepattern, $item, $matches);
-
-            return ! empty($matches) ? $item : null;
-        } elseif ($item instanceof Collection) {
+            return preg_match('/^[a-z]{2}((-|_)[a-z]{2})?$/m', $item) === 1 ? $item : null;
+        }
+        if ($item instanceof Collection) {
```
bobvandevijver, line 81:
```suggestion

        if ($item instanceof Collection) {
```

bobvandevijver (approved):

PR #3530, opened 2025-03-06
Release updates

bobvandevijver:
Small updates to reflect the current repository state and fix all actions (except for Cypress and accessibility)

PR #3533, opened 2025-03-06
:white_check_mark: Update Github action workflow

macintoshplus:
This PR updates the GitHub Action workflows to use the latest versions of the actions and their specific configuration.

`shivammathur/setup-php@v2` installs the latest version of composer which removes the update actions.

It also installs the Symfony client which can be used as a web server for testing.

This removes the `symfony/web-server-bundle` dependency.

### DIFF
--- a/.github/workflows/code_analysis.yaml
+++ b/.github/workflows/code_analysis.yaml
@@ -54,6 +54,7 @@ jobs:
                     php-version: ${{ matrix.php-version }}
                     extensions: json, mbstring, pdo, curl, pdo_sqlite
                     coverage: none
+                    tools: composer, symfony-cli
 
             -   run: composer install --no-progress --ansi
 
--- a/.github/workflows/code_analysis.yaml
+++ b/.github/workflows/code_analysis.yaml
@@ -54,7 +54,9 @@ jobs:
                     php-version: ${{ matrix.php-version }}
                     extensions: json, mbstring, pdo, curl, pdo_sqlite
                     coverage: none
-                    tools: composer, symfony-cli
+                    tools: symfony-cli:5
+                env:
+                    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
             -   run: composer install --no-progress --ansi
 
--- a/.github/workflows/cypress_tests.yaml
+++ b/.github/workflows/cypress_tests.yaml
@@ -25,7 +25,11 @@ jobs:
                     php-version: 8.2
                     extensions: json, mbstring, pdo, curl, pdo_sqlite
                     coverage: none
-                    tools: symfony-cli
+                    tools: symfony-cli:5
+                    # Configure PHP to disable Session garbage collector and fix the timezone to UTC like Bolt use to compare date.
+                    ini-values: session.gc_probability=0, date.timezone=UTC
+                env:
+                    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
             -   uses: actions/setup-node@v4
                 with:
                     node-version: 20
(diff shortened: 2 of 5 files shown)


bobvandevijver:
This will conflict with the work done in #3530, so it will need to be rebased once that has been merged (which I cannot do yet due to a missing review).

macintoshplus:
No problem, I have read the #3530 PR.

.github/workflows/cypress_tests.yaml
```diff
@@ -50,25 +51,37 @@ jobs:
                     # build assets
                     npm run build
                     sudo chmod -R 777 config/ public/files/ public/theme/ public/thumbs/ var/
+                    
+                    #Configure PHP to disable Session garbage collector and fix the timezone to UTC like Bolt use to compare date.
```
bobvandevijver, line 55:
```suggestion
                    # Configure PHP to disable Session garbage collector and fix the timezone to UTC like Bolt use to compare date.
```

.github/workflows/code_analysis.yaml
```diff
@@ -54,6 +54,7 @@ jobs:
                     php-version: ${{ matrix.php-version }}
                     extensions: json, mbstring, pdo, curl, pdo_sqlite
                     coverage: none
+                    tools: composer, symfony-cli
```
bobvandevijver, line 57: Maybe we should specify the major here to prevent surprises down the line?

```suggestion
                    tools: composer:v2, symfony-cli:5
```

I do see that the documentation of the actions mentions:

> With the exception of major versions of composer, if you specify only the major version or the version in major.minor format for a tool you can get rate limited by GitHub's API. To avoid this, it is recommended to provide a [GitHub OAuth token](https://github.com/shivammathur/setup-php#composer-github-oauth).

I cannot access the repository secrets just yet, but I'm still tempted to add this specification already. WDYT?

We will want to update this at all usages of the `shivammathur/setup-php@v2` task.

.github/workflows/cypress_tests.yaml
```diff
@@ -50,25 +51,37 @@ jobs:
...
+                    echo "date.timezone=UTC" >> php.ini
+                    echo "session.gc_probability=0" >> php.ini
+                    echo "session.gc_divisor=100000" >> php.ini
+                    echo "session.gc_maxlifetime=1440000" >> php.ini
```
bobvandevijver, line 59:
```suggestion
```

`session.gc_probability=0` should be enough to disable GC

macintoshplus:
@bobvandevijver This PR is rebased. But, for testing, it's more than using the lower-supported version.

IMHO, PHP 8.0 because nobody uses Bolt with PHP 7.x. The CI increased PHP 7.x usage in the last day.

image


Stats source: https://packagist.org/packages/bolt/core/php-stats#5

.github/workflows/cypress_tests.yaml
```diff
@@ -50,25 +51,37 @@ jobs:
                     # build assets
                     npm run build
                     sudo chmod -R 777 config/ public/files/ public/theme/ public/thumbs/ var/
+                    
+                    #Configure PHP to disable Session garbage collector and fix the timezone to UTC like Bolt use to compare date.
+                    echo "date.timezone=UTC" >> php.ini
+                    echo "session.gc_probability=0" >> php.ini
```
bobvandevijver, line 57: Maybe better to use https://github.com/shivammathur/setup-php?tab=readme-ov-file#ini-values-optional instead?

bobvandevijver (changes requested):

bobvandevijver:
For PHP-version I'm all for dropping all unsupported PHP versions with 5.3.0, making the project >=8.1.

bobvandevijver (approved):

macintoshplus:
> For PHP version I'm for dropping all unsupported PHP versions with 5.3.0, making the project >=8.1.

Bolt follows [semantic versioning](https://semver.org/) and this change in a minor version is not allowed, only in a major.

Symfony has changed the minimum PHP version required to 8.1 when releasing Symfony 6.1. The Symfony community has been angry.

I prefer to keep PHP 7.2 for the 5.x branch and bump to PHP 8.1 with Bolt 6.x and Symfony 7.4. The change will be required by Symfony.

Where can you discuss the roadmap?

PR #3534, opened 2025-03-06
:bug: Change the check way to check if the user have right to change some property

macintoshplus:
This PR fixes one of the Cypress tests for users with the ROLE_EDITOR role.

If the user does not have permission to change the properties, the original values ​​are reapplied.

With the old behavior, a user with the ROLE_EDITOR role (john_editor in the Cypress tests) cannot add or modify content.


macintoshplus:
Ping me if you want more explanation for this change.

PR #3541, opened 2025-03-07
feat: added or improved czech translation files

Vondry:
I have tried to update all 3 translation files for Czech language 🇨🇿 . Hope it will be useful.

### DIFF
--- a/translations/messages.cs.xlf
+++ b/translations/messages.cs.xlf
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:2.0" version="2.0" srcLang="en" trgLang="cs">
-  <file id="messages.en">
+  <file id="messages.cs">
     <unit id="tbB2gib" name="not_available">
       <notes>
         <note>templates/debug/source_code.twig:26</note>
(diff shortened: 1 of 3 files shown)


translations/messages.cs.xlf
```diff
@@ -1205,14 +1205,14 @@
     </unit>
     <unit id="eWLSFJs" name="editor_embed.label_pixel">
       <segment state="translated">
-        <source>editor_embed.label_pixel</source>
+        <source>pixel</source>
```
bobvandevijver, line 1208: I believe this should remain `editor_embed.label_pixel`, and there are multiple similar changes below. Can you check and revert them?

bobvandevijver (changes requested):

translations/messages.cs.xlf
Vondry: Based on my manual tests, Bolt CMS or more likely Symfony uses the value of the `name` attribute to find the translated text. 

- So I tried updating the `source` tag to contain the original English text, which is matched by the translated text in `target`, so that anyone knows where the Czech (translated) text comes from - I considered this actually as a good practice to keep it there instead of auto-generated value that is same as in `name` attribute.

There could be done some general cleanup of all translations, because it is sometimes a bit messy (obsolete, weird names, ...), but this for some future refactor though.

bobvandevijver: According to the [format](https://symfony.com/doc/current/reference/formats/xliff.html) as published by Symfony, the source tag and name attribute need to be the same. I'm not sure as to why, but I do want to follow that also because that is used almost everywhere. Even in this PR you have not updated it consistently.

So, please the source tag content as it was.

Vondry: Allright, updated and verified using this XPath: `//file/unit[@name != segment/source]` on http://xpather.com. 

Should be good now ✅

bobvandevijver (approved):

bobvandevijver:
Love it, thank you!

PR #3544, opened 2025-03-10
:construction_worker: add build Assets package

macintoshplus:


bobvandevijver (approved):

bobvandevijver:
Thanks! I dropped the matrix as we will only ever need one build.

Issue #3551, opened 2025-03-11
Upgrade PHP 8.1 minimum and dependencies


macintoshplus:
This issue is a meta issue to follow the dependencies update for next version of Bolt.

The list of outdated direct dependencies:

```
Direct dependencies required in composer.json:
api-platform/core                  v2.7.18   v3.4.16  Build a fully-featured hypermedia or GraphQL API in minutes!
babdev/pagerfanta-bundle           v2.11.0   v4.5.0   Bundle integrating Pagerfanta with Symfony
beberlei/doctrineextensions        v1.3.0    v1.5.0   A set of extensions to Doctrine 2 that add support for additional query functions available in MySQL, Oracle, PostgreSQL and SQLite.
coduo/php-matcher                  5.0.1     6.0.17   PHP Matcher enables you to match values with patterns
dama/doctrine-test-bundle          v6.7.5    v8.3.0   Symfony bundle to isolate doctrine database tests and improve test performance
doctrine/doctrine-bundle           2.7.2     2.13.2   Symfony DoctrineBundle
doctrine/doctrine-fixtures-bundle  3.4.5     4.0.0    Symfony DoctrineFixturesBundle
doctrine/orm                       2.14.3    3.3.2    Object-Relational-Mapper for PHP
embed/embed                        v3.4.18   v4.4.15  PHP library to retrieve page info using oembed, opengraph, etc
knplabs/doctrine-behaviors         2.5.0     2.6.2    Doctrine Behavior Traits
league/glide-symfony               1.1.1     2.0.1    Glide adapter for Symfony
nesbot/carbon                      2.73.0    3.8.6    An API extension for DateTime that supports 281 different languages.
php-http/httplug-bundle            1.34.3    2.1.0    Symfony integration for HTTPlug
php-translation/symfony-bundle     0.12.8    0.15.0  
phpstan/phpstan                    1.12.21   2.1.8    PHPStan - PHP Static Analysis Tool
phpstan/phpstan-doctrine           1.5.7     2.0.2    Doctrine extensions for PHPStan
phpstan/phpstan-symfony            1.4.13    2.0.2    Symfony Framework extensions and rules for PHPStan
phpunit/phpunit                    8.5.41    10.5.45  The PHP Unit Testing framework.
psr/simple-cache                   1.0.1     3.0.0    Common interfaces for simple caching
scienta/doctrine-json-functions    4.5.0     6.3.0    A set of extensions to Doctrine 2 that add support for json query functions.
[...]

PR #3555, opened 2025-03-11
:arrow_up: Replace dependency tightenco/collect by illuminate/collect…

macintoshplus:
Replace dependency tightenco/collect by illuminate/collection

See issue #3551


bobvandevijver (changes requested):
This is a BC break due to the namespace change. We need at the very least a note in the release document, as to what has changed, and I am at doubt whether we can merge this for 5.3.

bobvandevijver:
5.3 won't be getting a Symfony upgrade for the same reason.

macintoshplus:
Ok, this is why Symfony dependencies were removed from ticket 3551.

It's Okay; all PR with BC Break can wait until after Bolt 5.3 to go to Bolt 6. Or a branch for Bolt 6 can be opened before Bolt 5.3 is released?

bobvandevijver:
The first focus is 5.3, and we can do for that without breaking BC.
The path for 6.0 is still to be discussed, and as I will be on holiday a final decision will be postponed until May.

bobvandevijver (approved):

PR #3557, opened 2025-03-11
:arrow_up: Add dependencies replace - issue #3551

macintoshplus:
The replace section allows the removal of unnecessary dependencies like polyfill.

All polyfill for PHP before the current PHP minimal version must be added in the `replace` section.

See issue #3551

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -146,6 +146,17 @@
             "spec\\Bolt\\": "tests/spec/Bolt/"
         }
     },
+    "replace": {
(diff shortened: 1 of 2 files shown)


bobvandevijver:
You have misunderstood how replace works, you have configured that when you include bolt/core in your project that bolt/core itself replaces those polyfills. This is not the responsibility of a library.

[docs](https://getcomposer.org/doc/04-schema.md#replace) - "Map of packages that are replaced by this package."

PR #3558, opened 2025-03-11
:arrow_up: remove unused coduo/php-matcher dependency - issue #3551

macintoshplus:
The development dependency `coduo/php-matcher` is not used in code and lock the `doctrine/orm` update.

See issue #3551

### DIFF
--- a/src/Doctrine/Functions/Rand.php
+++ b/src/Doctrine/Functions/Rand.php
@@ -33,7 +33,7 @@ class Rand extends FunctionNode
         $parser->match(Lexer::T_IDENTIFIER);
         $parser->match(Lexer::T_OPEN_PARENTHESIS);
 
-        if ($lexer->lookahead['type'] !== Lexer::T_CLOSE_PARENTHESIS) {
+        if ($lexer->lookahead->type !== Lexer::T_CLOSE_PARENTHESIS) {
             $this->expression = $parser->SimpleArithmeticExpression();
         }
 
--- a/src/Entity/Field.php
+++ b/src/Entity/Field.php
@@ -80,7 +80,7 @@ class Field implements FieldInterface, TranslatableInterface
     private $version;
 
     /**
-     * @ORM\ManyToOne(targetEntity="Bolt\Entity\Content", inversedBy="fields", fetch="EAGER")
+     * @ORM\ManyToOne(targetEntity="Bolt\Entity\Content", inversedBy="fields")
      * @ORM\JoinColumn(nullable=false)
      * @Groups("api_write")
      */
--- a/src/Entity/Field.php
+++ b/src/Entity/Field.php
@@ -80,7 +80,7 @@ class Field implements FieldInterface, TranslatableInterface
     private $version;
 
     /**
-     * @ORM\ManyToOne(targetEntity="Bolt\Entity\Content", inversedBy="fields")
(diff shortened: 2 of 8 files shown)


bobvandevijver (changes requested):
It's still mentioned in the about

macintoshplus:
About removed

bobvandevijver (changes requested):
Tests are still failing.

macintoshplus:
> Tests are still failing.

The API test faill with this error `Associations with fetch-mode=EAGER may not be using WITH conditions in`.

I have this error only with SQLite.

src/Entity/Field.php
```diff
@@ -80,7 +80,7 @@ class Field implements FieldInterface, TranslatableInterface
     private $version;
 
     /**
-     * @ORM\ManyToOne(targetEntity="Bolt\Entity\Content", inversedBy="fields", fetch="EAGER")
+     * @ORM\ManyToOne(targetEntity="Bolt\Entity\Content", inversedBy="fields")
```
bobvandevijver, line 83: The WITH query should be adjusted instead.

src/Doctrine/Functions/Rand.php
```diff
@@ -33,7 +33,7 @@ public function parse(\Doctrine\ORM\Query\Parser $parser): void
         $parser->match(Lexer::T_IDENTIFIER);
         $parser->match(Lexer::T_OPEN_PARENTHESIS);
 
-        if ($lexer->lookahead['type'] !== Lexer::T_CLOSE_PARENTHESIS) {
+        if ($lexer->lookahead->type !== Lexer::T_CLOSE_PARENTHESIS) {
```
bobvandevijver, line 36: This is because in the CI it migrated silently to lexer v3? What if another dependency is still requiring v2? It was an array before in v2, so we need to have something to guard against that.

bobvandevijver (changes requested):

src/Doctrine/Functions/Rand.php
macintoshplus: The dependency removal has unlocked some dependency upgrades like Lexer and Doctrine.

All libraries who need Lexer to accept the v2 or v3:
```
$ symfony composer why doctrine/lexer
doctrine/annotations    2.0.2   requires  doctrine/lexer (^2 || ^3)     
doctrine/orm            2.20.2  requires  doctrine/lexer (^2 || ^3)     
egulias/email-validator 4.0.3   requires  doctrine/lexer (^2.0 || ^3.0) 
symfony/doctrine-bridge v5.4.48 conflicts doctrine/lexer (<1.1)         
symfony/validator       v5.4.48 conflicts doctrine/lexer (<1.1)   
```

src/Entity/Field.php
macintoshplus: Changed.

src/Doctrine/Functions/Rand.php
bobvandevijver: Let's go for `doctrine/lexer: ^2 || ^3`, as we use the lexer internally in code we need to mark is as own dependency anyway. Looks like 2 added object access and contained both array and object access options, so the code should be fine.

macintoshplus: Dependency added. Ready to merge.

bobvandevijver (approved):

PR #3561, opened 2025-03-13
:recycle: Integrate translation behavior

macintoshplus:


bobvandevijver (changes requested):
Breaking change, needs a note in the upgrade docs.

config/services.yaml
bobvandevijver, line 1: Changes to the `services.yaml` will need to have a yaml migration defined as well.

config/services.yaml
```diff
@@ -31,6 +33,8 @@ services:
             $tablePrefix: '%bolt.table_prefix%'
             $backendUrl: '%bolt.backend_url%'
             $rememberLifetime: '%bolt.remember_lifetime%'
+            $translatableFetchMode: '%bolt.doctrine_behaviors_translatable_fetch_mode%'
+            $translationFetchMode: '%bolt.doctrine_behaviors_translation_fetch_mode%'
```
bobvandevijver, line 37:
```suggestion
```

I do not think these two need a default bind. They are only used in the listeners, and should be explicitly bound to that service.

config/services.yaml
```diff
@@ -78,6 +82,12 @@ services:
     Bolt\Event\Listener\UserAvatarLoadListener:
         tags:
             - { name: doctrine.event_listener, event: postLoad }
+    
+    Bolt\Event\Listener\TranslatableListener:
+        tags:
```
bobvandevijver, line 87:
```suggestion
    Bolt\Event\Listener\TranslatableListener:
        arguments:
            $translatableFetchMode: '%bolt.doctrine_behaviors_translatable_fetch_mode%'
            $translationFetchMode: '%bolt.doctrine_behaviors_translation_fetch_mode%'
        tags:
```

Once we have upgraded to Symfony 6 we can probably use the `#[Autowire]` on the class instead, but for now we should do it like this.

bobvandevijver (changes requested):
Still need to check the new classes but I'm out of time for today, but here are already some comments that need to adressed anyways

Issue #3504, opened 2023-11-10
Any URI starting with bolt.backend_url is treated as restricted


lafor:
Bolt treats any URI starting with `bolt.backend_url` string as restricted, even if not immediately followed by a slash.

Details
-------

| Question                                              | Answer
|------------------------------------------------------ | ------
| Relevant Bolt Version                                 | 5.1.24
| Install type                                          |  Composer install
| PHP version                                           | 8.2

Reproduction
------------

### Steps to reproduce

Log out from the backend if logged in. Assuming your `bolt.backend_url` (as defined in `config/services.yaml`) is the default `/bolt`, try accessing an URI like `/bolts-and-nuts`.

### Expected result

You should be served `/bolts-and-nuts` resource if it exists, or 404 if it doesn't.

### Actual result

You're redirected to `/bolt/login`.

### Fix

Paths ending with `bolt.backend_url` in `access_control` definitions of `config/packages/security.yaml` should include trailing slashes, e.g.:

`- { path: '^%bolt.backend_url%/', roles: IS_AUTHENTICATED_REMEMBERED }`
`- { path: '^/(%app_locales%)%bolt.backend_url%/', roles: IS_AUTHENTICATED_REMEMBERED }`

macintoshplus:
This issue fix must be applied in the repository and in project repository.

The route '/bolt$' must be protected.

macintoshplus:
Hi @lafor,

The route `/bolts-and-nuts` is defined in controller or by content slug?

bobvandevijver:
This has actually been fixed in Bolt 6., dbeab8bde450ebf927902f071c52c164c7b66189.

PR #3573, opened 2025-03-17
:bug: Does not redirect ot login if url start by /bolt...

macintoshplus:
Issue #3504 Fix the security side effect on URL start with `/bolt`.

* `/bolt$` is secured and redirect on login if need. :lock:
* `/bolt/new/page` is secured and redirect on login if need. :lock: 
* `/bolt-and-nuts` Not secured. :unlock:

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -58,5 +58,5 @@ security:
         - { path: '^/(%app_locales%)%bolt.backend_url%/api', roles: ADMIN_API_ACCESS } # handled by voter
         - { path: '^%bolt.backend_url%/_trans', roles: ADMIN_TRANSLATE_ACCESS } # handled by voter
         - { path: '^/(%app_locales%)%bolt.backend_url%/_trans', roles: ADMIN_TRANSLATE_ACCESS } # handled by voter
-        - { path: '^%bolt.backend_url%', roles: IS_AUTHENTICATED_REMEMBERED }
-        - { path: '^/(%app_locales%)%bolt.backend_url%', roles: IS_AUTHENTICATED_REMEMBERED }
+        - { path: '^%bolt.backend_url%($|/)', roles: IS_AUTHENTICATED_REMEMBERED }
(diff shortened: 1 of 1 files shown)


bobvandevijver (changes requested):
This looks good, but it will require a yaml migration as well so it is to be applied to existing installations. See for example https://github.com/bolt/core/blob/master/yaml-migrations/m_2022-02-16-security_1.yaml.

macintoshplus:
> This looks good, but it will require a yaml migration as well so it is to be applied to existing installations. See for example https://github.com/bolt/core/blob/master/yaml-migrations/m_2022-02-16-security_1.yaml.

It's not possible. The YAML migration tool doesn't work fine with an array. This change requires two migrations. One for removing older configuration and one for adding new configuration.

The remove migration like 

```
# See: https://github.com/bolt/core/issues/3504

file: packages/security.yaml
since: 5.2.1
remove:
    security:
        access_control:
            - { path: '^%bolt.backend_url%', roles: IS_AUTHENTICATED_REMEMBERED }
            - { path: '^/(%app_locales%)%bolt.backend_url%', roles: IS_AUTHENTICATED_REMEMBERED }
```

Apply the removing migration generate this error:

```
PHP Fatal error:  Uncaught TypeError: YamlMigrate\ArrayMerge::removeArrayRecursively(): Argument #1 ($origin) must be of type array, string given, called in /home/user/dev/bolt/bolt-5.2/vendor/bobdenotter/yaml-migrations/src/ArrayMerge.php on line 62 and defined in /home/user/dev/bolt/bolt-5.2/vendor/bobdenotter/yaml-migrations/src/ArrayMerge.php:54
```

To apply this change, it's necessary to upgrade the YAML migration tool to add the updating action like `kaliop/ezmigrationbundle`.

macintoshplus:
I've added the upgrade file and updated the changelog.

This patch may have already been applied to users' Bolt projects.

PR #3587, opened 2025-07-16
:arrow_up: Replace bobdenotter/weatherwidget by bolt/weatherwidget 

macintoshplus:
Use Bolt fork of weatherwidget


macintoshplus:
Why?

bobvandevijver:
See Slack, but also because the namespace must be updated (we do not want to continue to use the `bobdenotter` namespace). Namespace clashes will become a thing (which is also probably why Bob made used his own namespace), so we need to decide/register which namespaces we use for what.

PR #3608, opened 2025-10-18
Migrate to attributes

bobvandevijver:
Migrates Bolt code base to attributes

### DIFF
--- a/src/Entity/Field.php
+++ b/src/Entity/Field.php
@@ -12,6 +12,7 @@ use Bolt\Configuration\Content\FieldType;
 use Bolt\Event\Listener\FieldFillListener;
 use Bolt\Repository\FieldRepository;
 use Bolt\Utils\Sanitiser;
+use Doctrine\DBAL\Types\Types;
 use Doctrine\ORM\Mapping as ORM;
 use Illuminate\Support\Collection;
 use Knp\DoctrineBehaviors\Contract\Entity\TranslatableInterface;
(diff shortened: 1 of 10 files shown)


src/Entity/Field.php
```diff
@@ -24,75 +25,67 @@
...
+    #[ORM\Column(type: 'integer')]
     private ?int $id = 0;
 
-    /**
-     * @ORM\Column(type="string", length=191)
-     * @Groups({"get_field","api_write"})
-     */
+    #[Groups(['get_field', 'api_write'])]
+    #[ORM\Column(type: 'string', length: 191)]
     public $name;
```
tobias-93: No type, do you want to add `?string` as a type?

src/Entity/Media.php
```diff
@@ -4,88 +4,84 @@
...
-     */
-    private $author;
+    #[ORM\ManyToOne(targetEntity: User::class)]
+    #[ORM\JoinColumn(nullable: true)]
+    private ?User $author = null;
 
-    /** @ORM\Column(type="datetime") */
-    private $createdAt;
+    #[ORM\Column(type: 'datetime')]
+    private DateTimeInterface $createdAt;
```
tobias-93:
```suggestion
    private DateTimeInterface $createdAt; // Default in constructor
```

src/Entity/Media.php
```diff
@@ -4,88 +4,84 @@
...
 
-    /** @ORM\Column(type="datetime") */
-    private $modifiedAt;
+    #[ORM\Column(type: 'datetime')]
+    private DateTimeInterface $modifiedAt;
```
tobias-93:
```suggestion
    private DateTimeInterface $modifiedAt; // Default in constructor
```

PR #3617, opened 2025-11-04
Integrate translation behaviour

bobvandevijver:
Rebased version of #3561.

### DIFF
--- a/src/Entity/Translatable/BoltTranslatableInterface.php
+++ b/src/Entity/Translatable/BoltTranslatableInterface.php
@@ -20,12 +20,12 @@ interface BoltTranslatableInterface
     public function getNewTranslations(): Collection;
 
     /**
-     * @phpstan-param T $translation
+     * @param T $translation
      */
     public function addTranslation(BoltTranslationInterface $translation): void;
 
     /**
-     * @phpstan-param T $translation
+     * @param T $translation
      */
     public function removeTranslation(BoltTranslationInterface $translation): void;
 
@@ -36,7 +36,7 @@ interface BoltTranslatableInterface
      *
      * @param string|null $locale The locale (en, ru, fr) | null If null, will try with current locale
      *
-     * @phpstan-return T
+     * @return T
      */
     public function translate(?string $locale = null, bool $fallbackToDefault = true): BoltTranslationInterface;
 
@@ -54,7 +54,7 @@ interface BoltTranslatableInterface
     public function getDefaultLocale(): string;
 
     /**
-     * @phpstan-return class-string<T>
+     * @return class-string<T>
      */
     public static function getTranslationEntityClass(): string;
 }
--- a/src/Entity/Translatable/BoltTranslatableTrait.php
+++ b/src/Entity/Translatable/BoltTranslatableTrait.php
@@ -51,7 +51,7 @@ trait BoltTranslatableTrait
     }
 
     /**
-     * @phpstan-param T $translation
+     * @param T $translation
      */
     public function addTranslation(BoltTranslationInterface $translation): void
     {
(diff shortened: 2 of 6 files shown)


src/Entity/Translatable/BoltTranslatableInterface.php
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public function getNewTranslations(): Collection;
+
+    /**
+     * @phpstan-param T $translation
+     */
+    public function addTranslation(BoltTranslationInterface $translation): void;
+
+    /**
+     * @phpstan-param T $translation
```
tobias-93:
```suggestion
     * @param T $translation
     */
    public function addTranslation(BoltTranslationInterface $translation): void;

    /**
     * @param T $translation
```

src/Entity/Translatable/BoltTranslatableTrait.php
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /**
+     * Returns translation for specific locale (creates new one if doesn't exists). If requested translation doesn't
+     * exist, it will first try to fallback default locale If any translation doesn't exist, it will be added to
+     * newTranslations collection. In order to persist new translations, call mergeNewTranslations method, before flush
+     *
+     * @param string|null $locale The locale (en, ru, fr) | null If null, will try with current locale
+     *
+     * @phpstan-return T
```
tobias-93:
```suggestion
     * @param T $translation
     */
    public function addTranslation(BoltTranslationInterface $translation): void
    {
        $this->getTranslations()
            ->set($translation->getLocale(), $translation);
        $translation->setTranslatable($this);
    }

    /**
     * @param T $translation
     */
    public function removeTranslation(BoltTranslationInterface $translation): void
    {
        $this->getTranslations()
            ->removeElement($translation);
    }

    /**
     * Returns translation for specific locale (creates new one if doesn't exists). If requested translation doesn't
     * exist, it will first try to fallback default locale If any translation doesn't exist, it will be added to
     * newTranslations collection. In order to persist new translations, call mergeNewTranslations method, before flush
     *
     * @param string|null $locale The locale (en, ru, fr) | null If null, will try with current locale
     *
     * @return T
```

src/Entity/Translatable/BoltTranslatableTrait.php
```diff
@@ -0,0 +1,275 @@
...
+            if ($translation && ! $translation->isEmpty()) {
+                return $translation;
+            }
+        }
+
+        return $this->findTranslationByLocale($this->getDefaultLocale(), false);
+    }
+
+    /**
+     * @phpstan-return class-string<T>
```
tobias-93:
```suggestion
     * @return class-string<T>
```

src/Entity/Translatable/BoltTranslationTrait.php
```diff
@@ -0,0 +1,76 @@
...
+        // By default, the translatable class has the same name but without the "Translation" suffix
+        return mb_substr(static::class, 0, -11, 'UTF-8');
+    }
+
+    protected string $locale = '';
+
+    /**
+     * Will be mapped to translatable entity by TranslatableSubscriber.
+     *
+     * @phpstan-var T|null
```
tobias-93:
```suggestion
     * @return class-string<T>
     */
    public static function getTranslatableEntityClass(): string
    {
        // By default, the translatable class has the same name but without the "Translation" suffix
        return mb_substr(static::class, 0, -11, 'UTF-8');
    }

    protected string $locale = '';

    /**
     * Will be mapped to translatable entity by TranslatableSubscriber.
     *
     * @var T|null
```

PR #3618, opened 2025-11-05
Remove migrations

### LINKED ISSUE #3613: Remove migrations
We will be dropping migrations from bolt/core as migrations are highly database specific, so we will need something to tell the users to migrate to.

bobvandevijver:
Resolves #3613.

Depends on #3617.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -2,7 +2,7 @@
 
 ## Dropped Bolt provided migrations
 
-As migrations heavily depend on the used database and were incomplete to begin with, we have dropped database migrations from this bundle. Instead, we will be offering a migration path or example migrations in these notes.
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -0,0 +1,53 @@
+# From Bolt 5.2 to 6.0
+
+## Dropped Bolt provided migrations
+
+As migrations heavily depend on the used database and were incomplete to begin with, we have dropped database migrations from this bundle. Instead, we will be offering a migration path or example migrations in these notes.
```
tobias-93:
```suggestion
As migrations heavily depend on the used database and were incomplete to begin with, we have dropped database migrations from this application. Instead, we will be offering a migration path or example migrations in these notes.
```

tobias-93 (changes requested):
Some textual updates, this can be merged when the translation PR is merged

PR #3662, opened 2025-12-30
Register controller services

bobvandevijver:
Extension controller should use the `ExtensionController` base class, which uses the `ServicesTrait`. However, those services were not populated correctly. This PR fixes that.


tobias-93 (approved):

tobias-93:
I see a failed test, can you check?

PR #3664, opened 2026-01-09
Fix ImageController filename

clem:
The filename here does not need to be updated as it will be the full path filename, instead of the relative path filename.

For more information, this bug was explained on Bolt's Slack.


bobvandevijver:
Closing this as incorrect fix, see #3665 for the 5.2 branch.

PR #3665, opened 2026-01-10
Fix thumb save location

bobvandevijver:
Path calculation was broken after d007cf8efdf09f28a3ec7cfee72c1afa57f1b353. Instead, calculate the correct path earlier, and ensure path traversal is not possible.

### DIFF
--- a/src/Controller/ImageController.php
+++ b/src/Controller/ImageController.php
@@ -12,6 +12,7 @@ use League\Glide\Responses\SymfonyResponseFactory;
 use League\Glide\Server;
 use League\Glide\ServerFactory;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Path;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


clem:
Just needed to update the top of the class with this code for Bolt 6.0:

```php
class ImageController
{
    private Server $server;

    /**
     * @var array{
     *     w?: int,
     *     h?: int,
     *     fit?: string,
     *     location?: string,
     *     q?: int
     * }
     */
    private array $parameters = [];

    private Request $request;

    public function __construct(
        private readonly Config $config
    ) {
    }

    #[Route(path: '/thumbs/{paramString}/{filename}', name: 'thumbnail', requirements: ['filename' => '.+'], methods: [Request::METHOD_GET])]
    public function thumbnail(Request $request, string $paramString, string $filename): Response
    {
        $this->request = $request;

        if (!$this->isImage($filename)) {
            return $this->sendErrorImage();
        }
        // ...
```

And thanks a lot because it works!

PR #3704, opened 2026-04-13
Disable save button during form submission

### LINKED ISSUE #2837: Disappearing content in localized collections/sets
Download this Bolt instance with only the affected contenttype to reproduce the bug: [Wetransfer](https://we.tl/t-ovMFVe7v0g)

In a localized Bolt 5.0.2 with 
```
pages:
    name: Pages
    singular_name: Page
    fields:
(...)
        contentblocks:
            type: collection
            label: Content Blocks
            group: Extra Content Blocks
            localize: true
            fields:
                partnerlistitem:
                    label: Partner list item
                    icon: "fa-coffee"
                    type: set
                    localize: true
                    fields: 
                        partnertitle: 
                            type: text
                            label: "Partner title"
                            postfix: "<h4>Only shown in the backend to keep things orderly</h4>"
                        partnerlogo:
                            type: image
                            upload: "logos"
                            # localize: true
                        partnertext: 
                            type: redactor
                            localize: true
```
When I fill in a second or third `partnerlistitem`, content in the preceeding `partnerlistitem` is deleted. 
I cannot put my finger on what gets deleted when, but I _think_ that at least `partnerlogo` fields in the existing `partnerlistitem`s are deleted/emptied whenever a new `partnerlistitem` is added - even before you add content to the new item.

In this usecase, ideally the same logo would be shown for all languages, so you don't have to fill that in twice. 
As you can see I tried some variations with the places where  `localize: true` is applied, I don't think I saw any difference.

As mentioned in line 1, you can use this zipped Bolt instance (or set up a fresh one using its config files, and the theme directory) to reproduce the bug [Wetransfer](https://we.tl/t-ovMFVe7v0g). 

Start with an empty page and (in the #extra-content-blocks tab) add 2 or 3 `partnerlistitem`s, saving  in one or both languages and refreshing at the frontend after adding each new item. You should see some of your previously entered content disappear.

When trying this out, please note that `partnertitle` is not supposed to be shown in the frontend, only the `partnertext`.

kouz75:
link with  #2837 

                                                                                                                                                                                                                                                                                                                           
Prevent double-click and duplicate submissions on the edit content form by disabling the save button as soon as the user triggers a save.

### DIFF
--- a/assets/js/app/ajax-save.js
+++ b/assets/js/app/ajax-save.js
@@ -4,7 +4,8 @@ import { Toast } from 'bootstrap';
 
 let form = $('#editcontent');
 let record_id = form.data('record');
-let element = $('button[name="save"][type="button"]'); //type="button" because it should only work when ajaxy is enabled
+let elementButton = $('button[name="save"][type="button"]'); //type="button" because it should only work when ajaxy is enabled
+let elementSubmit = $('button[name="save"][type="submit"]'); //type="submit" for classic form POST
 
 let dom_element =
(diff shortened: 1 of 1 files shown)


kouz75:
@bobvandevijver Hi, have you been able to check the PR ?

bobvandevijver (approved):
I did not, but I did now. Many thanks @kouz75!

PR #3710, opened 2026-04-25
 Fix MySQL CAST AS TEXT regression

kouz75:
With MySQL/MariaDB orderby failed.

### A test have been replaced for bolt 6 :
```php        
  // Before (Bolt 5 / pre-DBAL v4) — correct:
  // TEXT→CHAR for everything EXCEPT SQLite → MySQL gets CHAR ✓                                                                                 
  if (! mb_strpos($backend_driver, 'sqlite') && $this->second === 'TEXT') {
      $this->second = 'CHAR';                                                                                                                   
  }                                                                                                                                           
                                                                                                                                                
  // After (Bolt 6.1.x) — regression:                                                                                                         
  // TEXT→CHAR ONLY for SQLite → MySQL gets TEXT ✗
  if ($platform instanceof SQLitePlatform && $this->second === 'TEXT') {                                                                        
      $this->second = 'CHAR';
  }                                                                                                                                             
  ```        

### to reproduce the bug:
**Create the content type**
```yaml   
  products:
      name: Products                                                                                                                            
      singular_name: Product
      title_format: "{title}-{subtitle}"                                                                                                        
      fields:                                                                                                                                 
          title:
              type: text
          subtitle:                                                                                                                             
              type: text
  ```                 

**in a twig file** 
 ```twig                                                                                                                                       
  {% setcontent allProducts = 'products' limit 200 orderby 'title' %}                                                                         
  {{ allProducts|length }}                                                                                                                      
  ```

**you should have the error** 
Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'TEXT)), LOWER(value_21) FROM (SELECT b0_.id AS id_0, b0_.content_type AS content' at line 1")

### DIFF
--- a/src/Doctrine/Query/Cast.php
+++ b/src/Doctrine/Query/Cast.php
@@ -34,7 +34,7 @@ class Cast extends FunctionNode
             }
         }
 
-        if ($platform instanceof AbstractMySQLPlatform && $this->second === 'TEXT') {
+        if (! $platform instanceof SQLitePlatform && $this->second === 'TEXT') {
             $this->second = 'CHAR';
         }
 
--- a/src/Doctrine/Query/Cast.php
+++ b/src/Doctrine/Query/Cast.php
@@ -35,7 +35,7 @@ class Cast extends FunctionNode
             }
         }
 
-        if ($platform instanceof SQLitePlatform && $this->second === 'TEXT') {
+        if ($platform instanceof AbstractMySQLPlatform && $this->second === 'TEXT') {
             $this->second = 'CHAR';
         }
 
--- a/src/Doctrine/Query/Cast.php
+++ b/src/Doctrine/Query/Cast.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Bolt\Doctrine\Query;
 
 use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
-use Doctrine\DBAL\Platforms\SQLitePlatform;
 use Doctrine\ORM\Query\AST\Functions\FunctionNode;
 use Doctrine\ORM\Query\AST\Node;
 use Doctrine\ORM\Query\Parser;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Doctrine/Query/Cast.php [hunk 1]
```diff
@@ -35,7 +34,7 @@ public function getSql(SqlWalker $sqlWalker): string
             }
         }
 
-        if ($platform instanceof SQLitePlatform && $this->second === 'TEXT') {
+        if ($platform instanceof AbstractMySQLPlatform && $this->second === 'TEXT') {
```
bobvandevijver:
```suggestion
        if (! $platform instanceof SQLitePlatform && $this->second === 'TEXT') {
```

This would keep it closer to what it was (https://github.com/bolt/core/commit/4aa9d133bb981e2ad473160842eadedc6d8a312b#diff-2b76b1288500f61bbdff2f047ee35a7fcfbbd2ac294636892bb16d25ddded623L25-L38), so I would prefer to do it like this.

bobvandevijver (changes requested):

src/Doctrine/Query/Cast.php
```diff
@@ -5,7 +5,6 @@
 namespace Bolt\Doctrine\Query;
 
 use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
```
bobvandevijver, line 7:
```suggestion
use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
use Doctrine\DBAL\Platforms\SQLitePlatform;
```

bobvandevijver (approved):

src/Doctrine/Query/Cast.php
bobvandevijver: Ah, you disabled the allow edits from maintainers, otherwise I could have merged this already 😢

PR #3713, opened 2026-04-26
Mysql - Fix field trait item types

kouz75:
Iterating over a select field on MySQL throws a fatal error

On SQLite, JSON is decoded as strings → current() returns string → no error.                                                                                                                                                                   
On MySQL, JSON integers are decoded as PHP int → current() returns int 
   → type error against the declared Field|string return type.                                                                                                           

### To reproduce    
**Content type**
````
homepage:
    name: Homepage
    singular_name: Homepage
    fields:
        related_entry:
            type: select
            values: entries/{title}
````    
**Twig**  
{{dump(record.related_entry)}}

**Error**
An exception has been thrown during the rendering of a template ("Bolt\Entity\Field\SelectField::current(): Return value must be of type Bolt\Entity\Field|string, int returned") in index.twig at line 10.

### DIFF
--- a/src/Doctrine/Query/Cast.php
+++ b/src/Doctrine/Query/Cast.php
@@ -35,7 +35,7 @@ class Cast extends FunctionNode
             }
         }
 
-        if ($platform instanceof SQLitePlatform && $this->second === 'TEXT') {
+        if ($platform instanceof AbstractMySQLPlatform && $this->second === 'TEXT') {
             $this->second = 'CHAR';
         }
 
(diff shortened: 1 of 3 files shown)


bobvandevijver:
Ah, can you rebase this on 6.1? 6.2 will be for the Vue upgrade, but I needed to make that the main branch to convince dependabot to use that branch for its updates...

bobvandevijver:
@kouz75 Something went wrong with the rebase, can you check that?

bobvandevijver (approved):

PR #3714, opened 2026-04-28
update thumbnail, allow user to choose output format

kouz75:
Add output format support to thumbnail URLs                                                                                                                                                                                                                
                                                                                                                                                                                                                                                             
  - Add optional format parameter to ThumbnailHelper, ImageExtension, and the thumbnail() Twig function                                                                                                                                                      
  - Encode the output format as a filename suffix in the URL (e.g. image.jpg.avif) instead of in the param string                                                                                                                                            
  - Parse the format suffix in ImageController to set Glide's fm parameter and resolve the correct source file                                                                                                                                               
  - Backward compatible: URLs without a format suffix continue to work unchanged        

Update thumbnail file path in Twig and filesystem to match image URL structure, allowing Apache/Nginx to serve them directly without any PHP call


Vondry:
As a general question - is there a command that would "warm up" thumbnail cache so that even first user can benefit from fast loading image?

kouz75:
There is no command for that yet. 
One way to fix this is to add a new command that parses all media files (or the content) and generates their thumbnails. The thumbnail settings can be defined in theme.yaml or passed as parameters to the command.      

This command may generate a lot of thumbnails that will never be displayed on the website.

src/Controller/ImageController.php
```diff
@@ -162,21 +179,38 @@ private function parseParameters(string $paramString): void
         $this->parameters = [
             'w' => (isset($raw[0]) && is_numeric($raw[0])) ? (int) $raw[0] : 400,
             'h' => (isset($raw[1]) && is_numeric($raw[1])) ? (int) $raw[1] : 300,
-            'fit' => $raw[2] ?? $this->config->get('general/thumbnails/default_cropping', 'default'),
```
bobvandevijver, line 165: Why has the usage of this setting been removed?

src/Controller/ImageController.php
```diff
@@ -45,17 +47,20 @@ public function thumbnail(Request $request, string $paramString, string $filenam
             return $this->sendErrorImage();
         }
 
+        $this->parseParameters($paramString);
+        $urlFilename = $filename;
+        $sourceFilename = $this->parseFormatFromFilename($filename);
```
bobvandevijver: Direct usage of the filename will cause a regression into a security incident (#3661) we had in the past, namely a path traversal. Paths can only be used after a `canonicalize` call.

bobvandevijver (changes requested):
I will need more time to definitively check this, but here's already my first findings. I also believe that are changes in this PR that are not necessarily needed to fix the main issue (which is fixing the webserver direct file serving).

src/Controller/ImageController.php
kouz75: I add the default value for fit/quality.

src/Controller/ImageController.php
bobvandevijver: The filename is still used directly without sanitation, this needs to be resolved first.

src/Controller/ImageController.php
```diff
@@ -82,7 +89,19 @@ private function getPath(Request $request, ?string $path = null, bool $absolute
...
+    private function parseFormatFromFilename(string $filename): string
+    {
+        $ext = mb_strtolower(pathinfo($filename, PATHINFO_EXTENSION));
+        if ($this->isSupportedFormat($ext) && pathinfo(pathinfo($filename, PATHINFO_FILENAME), PATHINFO_EXTENSION) !== '') {
+            $this->parameters['fm'] = $ext;
+
+            return mb_substr($filename, 0, -(mb_strlen($ext) + 1));
+        }
+
+        return $filename;
```
bobvandevijver: If I understand correctly, the goal is to retrieve the last part of the path here. Why not just explode on `.`, and validate if the match is one of the supported extensions?

bobvandevijver (changes requested):
Lets target this for 6.2 as it changes the paths. Also, can you enable edits from maintainers? If that was enabled I could have fixed the pipeline issues for you.

src/Controller/ImageController.php
kouz75: I'm not able to allow maintainers to edit the PR ???
I switch the PR to 6.2 and fix both issues.

bobvandevijver:
Okay, so I really wanted to take a real look at this PR and test it locally, but my time has run out working on a stupid failing a11y test (!3730), setting up a fully functional test environment with a complete nginx setup instead of the local symfony server and the recent responsible disclosures. So, rest assured, I haven't forgotten, I'm just lacking some time!

PR #3717, opened 2026-05-07
Prevent user to create/delete folder/files anywhere

kouz75:
An authenticated user with managefiles permission could supply a path parameter (e.g.    
  /../../src) to delete a directories anywhere on the filesystem that the web process can write to.
  
 - Replace Path::canonicalize() with the project's own PathCanonicalize::canonicalize() in the delete() and create() methods of FilemanagerController

### DIFF
--- a/src/Controller/Backend/FilemanagerController.php
+++ b/src/Controller/Backend/FilemanagerController.php
@@ -96,11 +96,15 @@ class FilemanagerController extends TwigAwareController implements BackendZoneIn
         $path = $this->getFromRequest($request, 'path');
         $location = $this->getFromRequest($request, 'location');
 
+        if (! is_string($path)) {
+            throw new \RuntimeException('Invalid filename');
+        }
+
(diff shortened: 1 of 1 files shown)


src/Controller/Backend/FilemanagerController.php
```diff
@@ -96,11 +97,15 @@ public function delete(Request $request): Response
         $path = $this->getFromRequest($request, 'path');
         $location = $this->getFromRequest($request, 'location');
 
+        if (! is_string($path)) {
+            throw new RuntimeException('Invalid filename');
```
bobvandevijver: I think a BadRequestHttpException would be better here.

bobvandevijver (changes requested):
@kouz75 What's the reason this is draft? By the way, no need to merge 6.1 in on every change, that will come when it is ready to merge 👍🏻

bobvandevijver (approved):

PR #3732, opened 2026-06-04
Fix: Resetting forgotten password

Vondry:
## Bug

 Requesting a password reset at `/bolt/reset-password` returned HTTP 500 with a `TypeError`: `buildResetEmail()` declared an  `array $config` parameter, but `config->get('general/reset_password_settings')` returns a `Bolt\Collection\DeepCollection`. Under `strict_types=1` PHP refuses the object→array coercion.

image

### 1. Primary fix
  Fixed by changing the parameter type hint to `DeepCollection`, which extends Laravel's `Collection` and implements `ArrayAccess`, so the existing `$config['mail_from']` lookups still work unchanged.
  
  ### 2. Test Environment Configuration

- Added `MAILER_DSN=null://null` to `phpunit.xml.dist` to prevent mailer configuration errors during tests.
- Hardened URL parsing in `src/Canonical.php` to safely handle malformed URLs and incomplete `parse_url()` results.

### 3. Test Coverage

Added `ResetPasswordControllerTest.php` to verify that:

- The password reset page loads successfully
- The reset form renders correctly
- Requests are handled without triggering `TypeError` exceptions
  
  ## Testing

https://github.com/user-attachments/assets/6ee1bcca-bb0a-442a-956c-2fdff00f5313

### DIFF
--- a/src/Canonical.php
+++ b/src/Canonical.php
@@ -58,13 +58,8 @@ class Canonical
 
         $requestUrl = parse_url($this->request->getSchemeAndHttpHost());
 
-        // Handle test environment or malformed URLs
+        // Nothing to do if the request URL is malformed.
         if ($requestUrl === false || ! isset($requestUrl['scheme'])) {
-            $this->setScheme('http');
-            $this->setHost('localhost');
-            $this->setPort(null);
-            $_SERVER['CANONICAL_HOST'] = 'localhost';
-            $_SERVER['CANONICAL_SCHEME'] = 'http';
             return;
         }
 
@@ -76,13 +71,8 @@ class Canonical
 
         $configUrl = parse_url($configCanonical);
 
-        // Handle malformed canonical URL
+        // Nothing to do if the canonical URL is malformed.
         if ($configUrl === false || ! isset($configUrl['scheme']) || ! isset($configUrl['host'])) {
-            $this->setScheme('http');
-            $this->setHost('localhost');
-            $this->setPort(null);
-            $_SERVER['CANONICAL_HOST'] = 'localhost';
-            $_SERVER['CANONICAL_SCHEME'] = 'http';
             return;
         }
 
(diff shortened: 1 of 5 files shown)


src/Canonical.php
```diff
@@ -58,6 +58,16 @@ public function setRequest(?Request $request = null): void
 
         $requestUrl = parse_url($this->request->getSchemeAndHttpHost());
 
+        // Handle test environment or malformed URLs
+        if ($requestUrl === false || ! isset($requestUrl['scheme'])) {
+            $this->setScheme('http');
+            $this->setHost('localhost');
+            $this->setPort(null);
+            $_SERVER['CANONICAL_HOST'] = 'localhost';
+            $_SERVER['CANONICAL_SCHEME'] = 'http';
```
bobvandevijver: I do not really like these fallbacks to localhost, as this would also run in production. So, just returning should probably be enough here.

bobvandevijver (changes requested):

src/Canonical.php
```diff
@@ -66,6 +76,16 @@ public function setRequest(?Request $request = null): void
 
         $configUrl = parse_url($configCanonical);
 
+        // Handle malformed canonical URL
+        if ($configUrl === false || ! isset($configUrl['scheme']) || ! isset($configUrl['host'])) {
+            $this->setScheme('http');
+            $this->setHost('localhost');
+            $this->setPort(null);
+            $_SERVER['CANONICAL_HOST'] = 'localhost';
+            $_SERVER['CANONICAL_SCHEME'] = 'http';
```
bobvandevijver: Same comment for this one.

src/Canonical.php
Vondry: [Yep, done](https://github.com/bolt/core/pull/3732/commits/5a28e0a1241d44d2ec2dc4dd3d403ce0a35d3d82)

src/Canonical.php
Vondry: [Yep, done](https://github.com/bolt/core/pull/3732/commits/5a28e0a1241d44d2ec2dc4dd3d403ce0a35d3d82)

PR #3746, opened 2026-06-30
Update `bolt.cm` links to `boltcms.io`

Hamham6:
Avoids browser warning page about expired `bolt.cm` certificate when following these links.


Hamham6:
As far as I can tell, the failing test is unrelated?

bobvandevijver:
Thank you for your PR! At this time we do not have access to the sites, which is why they are in the current state. While working on restoring access, I will keep this PR open to see what we will be doing with the domains in the future.

PR #3750, opened 2026-07-06
Fix TimedPublishSubscriber for doctrine/dbal 4.x

### LINKED ISSUE #3748: `TimedPublishSubscriber` silently fails to publish/depublish content on doctrine/dbal 4.x
**Environment:**
- `bolt/core`: 6.1.3
- `doctrine/dbal`: 4.4.3 (required by bolt/core as `^4.0`)
- PHP: 8.3.31

### Description

`Bolt\Event\Subscriber\TimedPublishSubscriber::onKernelRequest()` is supposed to run on every `kernel.request` and automatically:
- flip `status: timed` → `published` once `published_at` has passed, and
- flip `status: published` → `held` once `depublished_at` has passed.

On a project using `doctrine/dbal: ^4.0` (as required by `bolt/core`'s own `composer.json`), this **never works**, for three separate reasons, all hidden by the method's blanket `catch (Throwable) { /* Fail silently */ }` with no logging:

1. It calls `$conn->executeUpdate(...)`. `Connection::executeUpdate()` was removed in `doctrine/dbal` 4.0 (only `executeStatement()` remains), so every call throws:
   ```
   Error: Call to undefined method Doctrine\DBAL\Connection::executeUpdate()
   ```
2. Even after switching to `executeStatement()`, the parameter array uses colon-prefixed keys (`[':now' => $now]`) instead of plain keys (`['now' => $now]`), which throws:
   ```
   Doctrine\DBAL\ArrayParameters\Exception\MissingNamedParameter: Named parameter "now" does not have a bound value.
   ```
3. Once both of those are fixed, binding still fails because `$now` is a `Carbon`/`DateTime` object passed without an explicit DBAL type, throwing:
   ```
   Error: Object of class DateTime could not be converted to string
   ```
   This requires an explicit `['now' => Doctrine\DBAL\Types\Types::DATETIME_MUTABLE]` type array.

### Impact

Content with a "Depublish date" in the past keeps showing on the frontend indefinitely (status stays `published`), since nothing ever flips it to `held`. This affects every content type, not just one — it's a core scheduling mechanism that is completely inert on any project using doctrine/dbal 4.x.

### Steps to reproduce

1. Install `bolt/core` 6.1.3 with `doctrine/dbal ^4.0`.
2. Set any content record to `status = published` with a `depublished_at` in the past.
3. Visit any page (triggers `kernel.request`).
4. Observe the record's status is still `published` (expected: `held`).
5. Query/log the raw SQL — the `UPDATE` from `TimedPublishSubscriber` never even appears, because it errors out silently before executing.

### Suggested fix (verified working)

```php
// src/Event/Subscriber/TimedPublishSubscriber.php
use Doctrine\DBAL\Types\Types;

// ...

try {
    $conn->executeStatement($queryPublish, ['now' => $now], ['now' => Types::DATETIME_MUTABLE]);
    $conn->executeStatement($queryDepublish, ['now' => $now], ['now' => Types::DATETIME_MUTABLE]);
} catch (Throwable) {
    // Fail silently, output user-friendly exception elsewhere.
}
```

As a secondary concern, it may be worth logging the caught `Throwable` (even at debug level) instead of swallowing it completely, since that's what made this bug invisible for so long.

Senne:
- Replace removed Connection::executeUpdate() with executeStatement()
- Use plain (non colon-prefixed) named parameters as required by DBAL 4
- Pass an explicit Types::DATETIME_MUTABLE type for the bound Carbon/DateTime value
- Log caught exceptions at debug level instead of swallowing them silently

Fixes #3748

### DIFF
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -1,6 +1,4 @@
-<?php
-
-declare(strict_types=1);
+<?php declare(strict_types=1);
 
 $ignoreErrors = [];
 $ignoreErrors[] = [
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -1,4 +1,4 @@
-<?php declare(strict_types=1);
+<?php declare(strict_types = 1);
 
 $ignoreErrors = [];
 $ignoreErrors[] = [
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -2666,12 +2666,6 @@ $ignoreErrors[] = [
 	'path' => __DIR__ . '/src/Event/Subscriber/AuthSubscriber.php',
 ];
 $ignoreErrors[] = [
-	'message' => '#^Call to an undefined method object\\:\\:executeUpdate\\(\\)\\.$#',
-	'identifier' => 'method.notFound',
-	'count' => 2,
-	'path' => __DIR__ . '/src/Event/Subscriber/TimedPublishSubscriber.php',
-];
-$ignoreErrors[] = [
 	'message' => '#^Method Bolt\\\\Event\\\\Subscriber\\\\TimedPublishSubscriber\\:\\:__construct\\(\\) has parameter \\$tablePrefix with no type specified\\.$#',
 	'identifier' => 'missingType.parameter',
 	'count' => 1,
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -1,4 +1,6 @@
-<?php declare(strict_types = 1);
+<?php
+
+declare(strict_types=1);
 
 $ignoreErrors = [];
 $ignoreErrors[] = [
--- a/src/Event/Subscriber/TimedPublishSubscriber.php
+++ b/src/Event/Subscriber/TimedPublishSubscriber.php
@@ -24,8 +24,11 @@ class TimedPublishSubscriber implements EventSubscriberInterface
     private Connection $defaultConnection;
     private string $tablePrefix;
 
-    public function __construct($tablePrefix, ManagerRegistry $managerRegistry, private LoggerInterface $logger)
-    {
+    public function __construct(
+        $tablePrefix,
+        ManagerRegistry $managerRegistry,
+        private LoggerInterface $logger
+    ) {
         /** @var Connection $connection */
         $connection = $managerRegistry->getConnection('default');
         $this->defaultConnection = $connection;
--- a/src/Event/Subscriber/TimedPublishSubscriber.php
+++ b/src/Event/Subscriber/TimedPublishSubscriber.php
@@ -23,9 +23,8 @@ class TimedPublishSubscriber implements EventSubscriberInterface
 
     private Connection $defaultConnection;
     private string $tablePrefix;
-    private LoggerInterface $logger;
 
-    public function __construct($tablePrefix, ManagerRegistry $managerRegistry, LoggerInterface $logger)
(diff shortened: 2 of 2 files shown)


src/Event/Subscriber/TimedPublishSubscriber.php
```diff
@@ -18,12 +21,17 @@ class TimedPublishSubscriber implements EventSubscriberInterface
...
-    {
-        $this->defaultConnection = $managerRegistry->getConnection('default');
+    public function __construct(
+        $tablePrefix,
+        ManagerRegistry $managerRegistry,
+        private LoggerInterface $logger
+    ) {
+        /** @var Connection $connection */
+        $connection = $managerRegistry->getConnection('default');
+        $this->defaultConnection = $connection;
```
bobvandevijver:
```suggestion
        $this->defaultConnection = $managerRegistry->getConnection('default');
```

bobvandevijver (approved):

phpstan-baseline.php
```diff
@@ -1,4 +1,6 @@
-<?php declare(strict_types = 1);
+<?php
+
+declare(strict_types=1);
```
bobvandevijver:
```suggestion
<?php declare(strict_types=1);
```

phpstan-baseline.php
```diff
@@ -1,4 +1,4 @@
-<?php declare(strict_types = 1);
+<?php declare(strict_types=1);
```
bobvandevijver:
```suggestion
<?php declare(strict_types = 1);
```

src/Event/Subscriber/TimedPublishSubscriber.php
```diff
@@ -7,7 +7,10 @@
 use Bolt\Doctrine\TablePrefixTrait;
 use Bolt\Entity\Content;
 use Carbon\Carbon;
```
bobvandevijver, line 10:
```suggestion
use Bolt\Log\LoggerTrait;
use Carbon\Carbon;
```

src/Event/Subscriber/TimedPublishSubscriber.php
```diff
@@ -18,11 +21,14 @@ class TimedPublishSubscriber implements EventSubscriberInterface
...
+    private Connection $defaultConnection;
     private string $tablePrefix;
 
-    public function __construct($tablePrefix, ManagerRegistry $managerRegistry)
-    {
+    public function __construct(
+        $tablePrefix,
+        ManagerRegistry $managerRegistry,
+        private LoggerInterface $logger
+    ) {
```
bobvandevijver:
```suggestion
    public function __construct($tablePrefix, ManagerRegistry $managerRegistry)
    {
```

bobvandevijver (commented):
Actually, lets use the logger trait instead

src/Event/Subscriber/TimedPublishSubscriber.php
```diff
@@ -6,8 +6,12 @@
 
 use Bolt\Doctrine\TablePrefixTrait;
 use Bolt\Entity\Content;
+use Bolt\Log\LoggerTrait;
 use Carbon\Carbon;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Types\Types;
 use Doctrine\Persistence\ManagerRegistry;
+use Psr\Log\LoggerInterface;
```
bobvandevijver:
```suggestion
```

bobvandevijver:
Thank you @Senne!

PR #3757, opened 2026-07-23
Fix failing Cypress E2E suite: login redirect (cookie_secure)

Vondry:
### Problem

Every login-based test failed with:

> expected `http://127.0.0.1:8088/bolt/` but got `http://127.0.0.1:8088/`

Authentication succeeded, but users were redirected to the homepage instead of the Bolt dashboard, causing the `/bolt/` assertion to time out.

### Root cause

The regression resulted from the interaction of two independent commits:

| Commit | Change |
|--------|--------|
| `fbd0ed3a` — *Resolve framework configuration deprecations* | Set `cookie_secure: true` (previously omitted, using Symfony's default of `false`). |
| `db37ed5d` — *Block open redirect on login endpoint* | Replaced `new RedirectResponse(...)` with `HttpUtils::createRedirectResponse(...)` for post-login redirects. |

`HttpUtils::createRedirectResponse()` validates redirect targets against a regular expression generated from the session cookie configuration (`AddSessionDomainConstraintPass`). With `cookie_secure: true`, Symfony generates an **HTTPS-only** pattern (`{^https://<host>$}i`).

The failing flow was:

1. The Cypress `login` helper visits `/bolt`, so Symfony stores the target path as `http://127.0.0.1:8088/bolt`.
2. CI runs the application over plain HTTP (`symfony server:start --no-tls`).
3. After authentication, the saved `http://...` URL fails the HTTPS-only validation, so `createRedirectResponse()` falls back to `/`.
4. Authentication succeeds, but the user lands on the homepage, causing the `/bolt/` assertion to time out.

This was broader than a test-only issue. Any Bolt installation running over plain HTTP would also have broken login sessions because browsers discard `Secure` cookies on HTTP.

### The fix

`cookie_secure: 'auto'` is Symfony's recommended value for this deprecation. It applies the `Secure` flag only when the current request uses HTTPS.

As a result:

- On HTTP requests, redirect validation accepts both `http://` and `https://` same-host URLs, restoring the post-login redirect to `/bolt/`.
- On HTTPS requests, validation remains HTTPS-only, preserving protection against open redirects and scheme downgrades.

**Security impact:** None for correctly configured HTTPS deployments. On HTTPS, `auto` behaves identically to `true`. On HTTP, `true` never provided meaningful protection because browsers discard `Secure` cookies anyway.

### DIFF
--- a/config/packages/framework.yaml
+++ b/config/packages/framework.yaml
@@ -13,7 +13,11 @@ framework:
     handler_id: ~
     cookie_lifetime: 1209600 #expires in 14 days
     storage_factory_id: session.storage.factory.native
-    cookie_secure: true
+    # 'auto' sets the Secure flag only on HTTPS requests. Using `true` would force it on
+    # plain-HTTP installs too, where browsers then drop the cookie and, via the session
+    # domain constraint, restrict post-login redirects to HTTPS-only targets (which breaks
+    # the same-host redirect after login on HTTP). See LoginFormAuthenticator.
+    cookie_secure: 'auto'
     cookie_samesite: lax
   # When using the HTTP Cache, ESI allows to render page fragments separately
   # and with different cache configurations for each fragment
--- a/config/packages/framework.yaml
+++ b/config/packages/framework.yaml
@@ -13,10 +13,6 @@ framework:
     handler_id: ~
     cookie_lifetime: 1209600 #expires in 14 days
     storage_factory_id: session.storage.factory.native
-    # 'auto' sets the Secure flag only on HTTPS requests. Using `true` would force it on
-    # plain-HTTP installs too, where browsers then drop the cookie and, via the session
-    # domain constraint, restrict post-login redirects to HTTPS-only targets (which breaks
-    # the same-host redirect after login on HTTP). See LoginFormAuthenticator.
     cookie_secure: 'auto'
     cookie_samesite: lax
   # When using the HTTP Cache, ESI allows to render page fragments separately
--- a/yaml-migrations/m_2026-08-03-framework.yaml
+++ b/yaml-migrations/m_2026-08-03-framework.yaml
@@ -1,0 +1,7 @@
+file: packages/framework.yaml
+since: 6.1.7
+
+add:
+  framework:
+    session:
+      cookie_secure: 'auto'


config/packages/framework.yaml
```diff
@@ -13,7 +13,11 @@ framework:
     handler_id: ~
     cookie_lifetime: 1209600 #expires in 14 days
     storage_factory_id: session.storage.factory.native
-    cookie_secure: true
+    # 'auto' sets the Secure flag only on HTTPS requests. Using `true` would force it on
+    # plain-HTTP installs too, where browsers then drop the cookie and, via the session
+    # domain constraint, restrict post-login redirects to HTTPS-only targets (which breaks
+    # the same-host redirect after login on HTTP). See LoginFormAuthenticator.
+    cookie_secure: 'auto'
```
bobvandevijver, line 16: This was explicitly set to secure to follow the Symfony recipe (in https://github.com/bolt/core/commit/fbd0ed3a2a2fe80b0848039bf61070278652b4ff, which also added a migration).

As we only need this for the cypress run, I would prefer to override the configuration from the action instead, using (for example) sed.

.github/workflows/cypress_tests.yaml
```diff
@@ -12,12 +12,14 @@ jobs:
     name: Cypress Tests
     runs-on: ubuntu-latest
     env:
-      PHP_VERSION: 8.2
+      # composer.lock is ignored byt git, so CI resolves dependencies fresh on every run.
+      # The resulting api-platform / serializer stack needs PHP 8.4: on 8.2 the content
```
bobvandevijver: If this is indeed true, there is an issue with the dependencies not indicating that they require 8.4. So I very much doubt this change is needed at all.

From what I have seen, the 502 is just because the GitHub action runners are very, very slow, and that triggers proxy timeouts in the nginx process as started by the Symfony CLI. Unfortunately I haven't found a way yet to configure that proxy timeout. Retrying can also solve the issue, as it fully depends on the running the action is being run on (and yes, I do not like that at all).

bobvandevijver (changes requested):

config/packages/framework.yaml
Vondry: This isn't only CI-specific — it affects any install served over plain HTTP (local dev without SSL, LAN/intranet, Docker by container name). 

With `cookie_secure: true`, `AddSessionDomainConstraintPass` restricts `security.http_utils` to `https://` redirect targets, and `HttpUtils::createRedirectResponse()` rewrites anything else to `/` — that's the `/bolt` → `/` failure CI caught. With `auto`, both regexps are passed and the scheme is resolved per request: **identical restriction under HTTPS, working redirect under HTTP.**

<details>
<summary>Mechanism + behaviour table</summary>

[`AddSessionDomainConstraintPass` (6.4)](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Compiler/AddSessionDomainConstraintPass.php):

```php
if ('auto' === ($sessionOptions['cookie_secure'] ?? null)) {
    $secureDomainRegexp = sprintf('{^https://%s$}i', $domainRegexp);
    $domainRegexp = 'https?://'.$domainRegexp;
} else {
    $secureDomainRegexp = null;
    $domainRegexp = (empty($sessionOptions['cookie_secure']) ? 'https?://' : 'https://').$domainRegexp;
}
```

| `cookie_secure` | HTTPS request | HTTP request |
| --- | --- | --- |
| `true` | `https://` only | `https://` only → redirect rewritten to `/` |
| `auto` | `https://` only | `http://` and `https://` allowed |
| `false` | both | both |

On `localhost` the session still works (browsers treat it as a potentially-trustworthy origin and keep the `Secure` cookie), which is why the symptom is a wrong redirect rather than a failed login. On any other plain-HTTP host the cookie is dropped and login fails outright.
</details>

On following the recipe: our `framework.yaml` *is* the [6.4 framework-bundle recipe](https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/6.4/config/packages/framework.yaml) (`annotations: false`, `handle_all_throwables: true` verbatim), and that recipe ships `cookie_secure: auto`. Symfony is moving the same way.

<details>
<summary>Upstream direction</summary>

- 6.4 deprecates *not* setting it: "It will default to `auto` in 7.0" — [Configuration.php#L653](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L653)
- 7.0 makes it the default: `->defaultValue('auto')` — [Configuration.php#L664](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L664)

That's why the 7.x recipes collapse the session block to `session: true`.
</details>

Re: overriding from the workflow — sed leaves the redirect broken for anyone on HTTP, and CI would exercise a config we don't ship. If we want `true` committed, use a `when@dev:` / `when@test:` block (the file already has `when@test`); an env var won't work, since the compiler pass reads the value at compile time and gets the unresolved placeholder (symfony/symfony#28051).

No security impact on HTTPS. Behind a TLS-terminating proxy it relies on `trusted_proxies`, and 6.4's `trusted_headers` default already includes `x-forwarded-proto` — happy to add a `TRUSTED_PROXIES` note to the deployment docs here.

.github/workflows/cypress_tests.yaml
Vondry: I can remove the commit with the PHP version bump and just keep the `cookie_secure` for now

config/packages/framework.yaml
bobvandevijver: Actually, Symfony does not define a default at all (recipes are not considered defaults). If not set, it will fallback to the PHP-configuration.

I'm fine with making it 'auto' in Bolt, but it will need a yaml migration for 6.1.7. And please leave out the verbose AI comments.

bobvandevijver: You forget to add the yaml migration and to drop the comment

Vondry: Sorry, should be now complete.

bobvandevijver (approved):

bobvandevijver:
Note that Cypress is still failing, but that seems to be an issue with the API generator and the collection types.

```
"Illuminate\Support\TKey" is not a valid array key type.
```

But that is something for a different PR -> #3774

PR #3758, opened 2026-07-23
Reduce PHPStan baseline: fix generic object types and invalid array keys

Vondry:
## Summary

Reduces the PHPStan level-8 baseline by fixing the underlying issues instead of suppressing them. Removes **8 baselined errors** across two focused, self-contained commits, with no behavior changes or signature changes.

## Motivation

`phpstan-baseline.php` accumulates suppressed findings over time. This PR removes two cohesive, high-signal groups where the root causes were genuine type-safety issues (unverifiable method calls and invalid array key types) rather than annotation noise. Each change is small enough to review line by line.

### 1. Fix generic `object` types in the Field factory and discriminator

Instantiating classes via `new $classname()` / `new $class()` from a plain `string` produces a generic `object`, preventing PHPStan from verifying subsequent `Field` method calls.

- `FieldRepository::factory()` — narrows the instantiated object to `Field` using an inline `@var` annotation.
- `FieldDiscriminatorListener::extractFieldType()` — narrows the instantiated object to `FieldInterface` using an inline `@var` annotation.

Removes **7** baseline entries:

- 6 × `method.notFound`
- 1 × `return.type`

### 2. Fix invalid array key types (`offsetAccess.invalidOffset`)

Several array keys could previously be inferred as `null` or `array`, which PHPStan level 8 correctly rejects as potentially invalid offsets.

- `ContentValidator` — skips relations whose target `contentType` is `null`. Those keys were never consumed downstream because validation keys only on real content-type names with `allowExtraFields => true`.
- `TranslationsManager` and `FieldExtension` — cast `getId()` (`?int`) to `int`, matching the runtime guarantee that entities are already persisted in these code paths.
- `SelectQuery::getCheckboxFieldExpression()` — casts the result of `key(...)` (`int|string|null`) to `string`, preserving existing runtime behaviour.
- `SelectQuery` join maps — narrows `Filter::getKey()` (`string|array`) to `string` via an inline `@var`, matching how the value is already used (`sprintf('%s', ...)`).

Removes **1** additional baseline entry (`array|string`) and eliminates six non-baselined failures that appear on newer PHP versions.

## Verification

- `vendor/bin/phpstan analyse` (level 8) → **`[OK] No errors`**, with no analysis errors and no orphaned baseline entries.
- Ran the affected test suite: `FieldRepository::factory()` remains covered and passes; overall test results are identical to `master` (pre-existing unrelated failures remain unchanged).
- All modified methods are `private` except the two annotation-only changes, so there is no external API impact.

## Notes for reviewers

- The only intentional behavioural change is the `continue` in `ContentValidator::relationsToMap()`, which was verified to be functionally inert.
- Type narrowing is implemented exclusively with inline `/** @var ... */` annotations. No runtime constructs (such as `assert()`) were introduced.

### DIFF
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -2588,12 +2588,6 @@ $ignoreErrors[] = [
 	'path' => __DIR__ . '/src/Event/Listener/ContentFillListener.php',
 ];
 $ignoreErrors[] = [
-	'message' => '#^Call to an undefined method object\\:\\:getType\\(\\)\\.$#',
-	'identifier' => 'method.notFound',
-	'count' => 1,
-	'path' => __DIR__ . '/src/Event/Listener/FieldDiscriminatorListener.php',
-];
-$ignoreErrors[] = [
 	'message' => '#^Parameter \\#1 \\$objectOrClass of class ReflectionClass constructor expects class\\-string\\<T of object\\>\\|T of object, string given\\.$#',
 	'identifier' => 'argument.type',
 	'count' => 1,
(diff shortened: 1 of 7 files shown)


VentyCZ:
Sorry to bother on a merged PR, but just came across it and had to see how it was done :sweat_smile: 

Wouldn't `class-string<X>` instead of `@var`-ing the resulting instance be better? :angel:  these are just my 2 cents :)

bobvandevijver:
@VentyCZ Yes, it would. Feel free to submit a new PR. 👍🏻

PR #3773, opened 2026-08-03
 Remove four keys from the messages domain, moving remaining translations to validators/security

Vondry:
Third of the PRs splitting up https://github.com/bolt/core/pull/3753. Targets `6.2`, and touches the `messages`, `validators` and `security` catalogues.

Four keys exist in the `messages` domain but are only ever resolved from a different domain. This PR removes them from `messages` entirely, moving the translations that only existed there into the domain that actually serves them.

As discussed, this is classified as a **breaking change** for consumers who translate these keys from the `messages` domain themselves, so it targets `6.2` and adds an entry to `UPGRADING.md`.

## The four keys

| Key | Used at | Resolved in |
|---|---|---|
| `user.not_valid_display_name` | `src/Entity/User.php:35,36` (`Assert\NotBlank`, `Assert\Length`) | `validators` |
| `user.not_valid_email` | `src/Entity/User.php:49` (`Assert\Email`) | `validators` |
| `user.not_valid_password` | `src/Entity/User.php:57` (`Assert\Length`) | `validators` |
| `You have to login in order to access this page.` | `src/Security/AuthenticationEntryPointRedirector.php:26` (`trans(..., 'security')`) | `security` |

These five call sites are the **only** occurrences anywhere in the repository, including `vendor/`.

## Why the `messages` copies cannot matter

`validator.translation_domain` is configured as `validators` (verified against the live container; not overridden in `framework.yaml`), and `ExecutionContext:145` translates every validation message in that domain.

Symfony's `Translator` supports fallback **locales**, but not fallback **domains**. A lookup in `validators` can never fall back to `messages`. These entries are unreachable by construction, not merely unused.

## Commit 1 — remove duplicates (no behavioural impact)

28 units removed across 8 locales (`cs`, `el`, `en`, `fr`, `ru`, `sk`, `tr`, `uk`), in every case where the serving domain already held the same key.

Verified by resolving all four keys through the real translator for every locale, before and after: output is byte-identical. Worth noting that 10 of the 28 had drifted from the serving copy (e.g. `cs` `messages` said *Chybný email* while users have always seen *Neplatný e-mail* from `validators`), which confirms they were never rendered.

## Commit 2 — move the remainder (behaviour changes, all improvements)

The remaining 9 entries had no copy in the serving domain, so they are moved rather than deleted:

| Target | Moved in |
|---|---|
| `security.fr` | `You have to login…` |
| `security.nl` | `You have to login…` |
| `validators.hu` | `user.not_valid_email`, `user.not_valid_password` |
| `validators.zh_CN` | `user.not_valid_display_name`, `user.not_valid_email`, `user.not_valid_password` |

These four locales previously fell back to English for these keys and now show their own translation. Verified against the live translator across all 16 locales: exactly these 7 lookups change, nothing else.

The two remaining `messages.nl` entries are **dropped** rather than moved. Their targets were English (`Invalid email`, `Invalid password`) rather than Dutch, and the truncated password string would have shadowed the fuller `en` fallback (`Invalid password. The password should contain at least 6 characters.`). Dropping them leaves `nl` rendering unchanged.

The `messages` domain no longer contains any of the four keys, for any locale.

## Commit 3 — upgrade notes

Adds a *Removed translations* section to `UPGRADING.md` listing the removed keys, their serving domain, and what consumers should do.

`bin/console lint:xliff translations/` passes on all 44 files.


bobvandevijver (changes requested):
Strictly speaking, this is classified as a breaking change. We do not know what our consumers are using, and that could result in them no longer having a translation.

For this change, we will need to target 6.2, and add a note to the upgrade notes (which translations have been removed).

By targeting the 6.2 branch, the translations that were kept because they need to be removed can then also be done in the same (this) PR.

bobvandevijver:
By the way, I've just update 6.2 so that the latest translation updates for 6.1 are available there now as well.

Vondry:
All should be done :)

PR #3780, opened 2026-08-04
fix(translations): correct srcLang in 7 XLIFF files

Vondry:

## Changes
Seven translation files declared srcLang identical to `trgLang` (e.g. `srcLang="ru" trgLang="ru"`), describing them as monolingual same-language documents. The source side of every unit is English, so the correct value is `srcLang="en"`.

Affected: `messages.pl`, `messages.ru`, `security.nl`, `security.ru`, `validators.nl`, `validators.pl`, `validators.ru`

This aligns them with the other 34 files and with the output of Symfony's XLIFF 2.0 dumper. No runtime behaviour changes, as Symfony resolves the locale from the filename; the fix matters for translation tooling, which reads these attributes to determine direction.


bobvandevijver (approved):

bobvandevijver:
Thank you @Vondry!

PR #3784, opened 2026-08-04
fix(translations): restore dropped placeholders in bg/de/es/hu

Vondry:

## Summary

Fixes translations that dropped placeholders required by the calling code, preventing runtime values from being displayed.

## Changes

| File | Keys |
| --- | --- |
| `messages.bg` | `general.greeting`, `listing_details_box.listing_template` |
| `messages.de` | `contenttypes.generic.overview`, `contenttypes.generic.recent`, `reset_password.check_email_sent_text_2` |
| `messages.es` | `collection.add_item` |
| `messages.hu` | `collection.add_item`, `title.filtered_by` |

## Why

Symfony substitutes placeholders by exact name. When a translation omits a placeholder, the corresponding runtime value is lost.

Several of these were effectively hardcoded translations rather than direct equivalents of the source text. For example:

- The Bulgarian greeting always addressed the user as "admin" instead of using the provided `%name%`.
- The German ContentType headings were hardcoded to "Pages", making them incorrect for every other ContentType.

### DIFF
--- a/translations/messages.bg.xlf
+++ b/translations/messages.bg.xlf
@@ -154,7 +154,7 @@
     <unit id="b_RPpcB" name="general.greeting">
       <segment>
         <source>general.greeting</source>
-        <target>Привет, админ!</target>
+        <target>Привет, %name%!</target>
       </segment>
     </unit>
     <unit id="bmdn3u9" name="action.logout">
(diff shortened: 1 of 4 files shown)


bobvandevijver (approved):

bobvandevijver:
Thank you @Vondry!

PR #3785, opened 2026-08-04
fix(translations): replace unresolved %url% in built-with-bolt string

Vondry:

## Why

The footer calls the string with no parameters:

```twig
{# public/theme/skeleton/partials/_footer.twig:17 #}
{{ __('general.phrase.built-with-bolt') }}
```

So `%url%` was never filled in and rendered literally as `href='%url%'` — a relative link that 404'd on every page.

The URL is already hardcoded elsewhere

English and 8 of the 12 locales already write the URL directly: en, cs, de, fr, nl, ru, uk, zh_CN.

Only `el`, `hu`, `it`, `tr` used %url%. This brings them in line:

```
- <a href='%url%' target='_blank' ...>
+ <a href='https://boltcms.io' target='_blank' ...>
```

Wording untouched — only the href. Also applied to the unreachable duplicate copies of the same sentence so it can't come back.

`%url%` in `http_error*.suggestion` is left alone — English uses it there too.

### DIFF
--- a/translations/messages.cs.xlf
+++ b/translations/messages.cs.xlf
@@ -882,7 +882,7 @@
     <unit id="z0k8hRg" name="9fb3e6e">
       <segment state="translated">
         <source>9fb3e6e</source>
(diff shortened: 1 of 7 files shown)


bobvandevijver (approved):

bobvandevijver:
Thank you @Vondry!

PR #3791, opened 2026-08-13
Fix: keep the save button enabled after previewing (#3778)

### LINKED ISSUE #3778: Unable to save after previewing
Clicking “Preview” in an article opens the preview in a new tab, but it also disables the "Save changes” button, which makes it impossible to save the article after approving the preview.

Vondry:
Fixes #3778.

**Preview** submits the edit form with `formtarget="_blank"`, so the edit page never unloads. The submit handler added in #3704 disabled the save button on *every* submit, leaving it disabled for good after a preview.

`ajax-save.js` now skips disabling when the submitting button renders its response in another browsing context. Real saves, ctrl+s and Enter submits carry no `formtarget`, so they keep the double-submit protection.

Adds a Cypress spec covering all three cases. Verified it fails without the fix and passes with it.

## Testing

https://github.com/user-attachments/assets/84d81a8b-8a0c-4e21-acd4-f0361614a35e


bobvandevijver (approved):

bobvandevijver:
Thank you @Vondry!

PR #3794, opened 2026-08-14
Add TypeScript foundation and upgrade build tooling

Vondry:
## Add TypeScript foundation and upgrade build tooling

Second step of the [Vue 3 / TypeScript migration](https://github.com/bolt/core/pull/3751), which adds TypeScript to the existing **Vue 2.7** codebase without touching component logic

### Changes

- **Encore 1.8.2 → 4.7.0** — the last release supporting Vue 2 (5.0 dropped it). Brings `sass-loader` 14 and `postcss-loader` 8, and drops `vue-template-compiler`.
- **TypeScript** — `tsconfig.json` (strict, `.ts`-only include), `ts-loader` via `enableTypeScriptLoader()`, `npm run typecheck` + a CI step, and `typescript-eslint` scoped to `assets/**/*.ts`.
- **Filters converted to TS**, and Vue 2 global filters (`Vue.filter()` / `$options.filters`, both removed in Vue 3) replaced with direct imports at all 13 call sites.

### Review notes

- `enableVueLoader()`'s `version` option is **deliberately omitted**. It only accepts `2` or `3`, and `2` selects the target that still requires `vue-template-compiler`. Unset, Encore detects `vue@2.7` and uses the SFC compiler Vue 2.7 ships itself.
- SFCs are **not** type-checked yet — none use `lang="ts"`. That comes when they move to `<script setup lang="ts">` and the checker becomes `vue-tsc`.

### Behaviour

Unchanged, except `trim()`, `strip()` and `uppercase()` no longer throw a `TypeError` on `null`/non-string input. Values that previously rendered are unaffected — `trim()` still passes non-strings through untouched rather than coercing, so numbers aren't truncated and objects don't become `[object Object]`.

Verified by diffing the compiled render functions of all changed SFCs, and by a 348-case differential harness against the pre-migration implementations.

### Next

Vitest + tests for the converted filters, then components to Composition API.

## Testing

I manually clicked through the admin pages and verified that they work as previously

image


Vondry:
I created follow up https://github.com/bolt/core/pull/3797 with test setup.

PR #3799, opened 2026-08-30
Reduce PHPStan baseline: use class-string instead

VentyCZ:
Follow-up to #3758, continuing the effort to shrink `phpstan-baseline.php` by fixing the underlying type issues instead of suppressing them. (using class-string syntax instead of forcing with `@var`)


src/Repository/FieldRepository.php
```diff
@@ -121,17 +120,22 @@ public static function factory(Collection $definition, string $name = '', string
         return $field;
     }
 
+    /**
+     * @return class-string<Field>|null
```
bobvandevijver, line 124:
```suggestion
     * @return class-string<FieldInterface>|null
```

Wouldn't that make more sense here?

src/Repository/FieldRepository.php
```diff
@@ -121,17 +120,22 @@ public static function factory(Collection $definition, string $name = '', string
...
         $classname = ucwords($type) . 'Field';
 
+        /** @var array<class-string> $classes */
         $classes = array_map(
             fn (ClassMetadata $entity): string => $entity->getName(),
             self::$em->getMetadataFactory()->getAllMetadata()
         );
 
         // Classnames of all fields (classes that implement Bolt\Entity\FieldInterface)
+        /** @var Collection<int, class-string<Field>> $allFields */
```
bobvandevijver, line 138:
```suggestion
        /** @var Collection<int, class-string<FieldInterface>> $allFields */
```

Same here?

bobvandevijver (commented):
Failing checks seems to be flaky CI, but I do have these two suggestions. Would that make sense?

